PR #85, opened 2013-11-14
Expose proto

jonathanong:
we make `lib/request` and stuff prototypes again and create instances using `Object.create()`. this allows you to do stuff like:

``` js
app.context.__defineGetter__('session', function () {
  return new Session(this)
})
```

and stuff. this is how express currently works. however, i think we should keep these exposed prototypes undocumented. 

now the other modules like basic auth and body parsing will work by wrapping `require('basic-auth')(app, options)`. 

also, i cleaned up the prototype fixtures a little.

### DIFF
--- a/test/application.js
+++ b/test/application.js
@@ -295,3 +295,87 @@ describe('app.respond', function(){
     })
   })
 })
+
+describe('app.context', function(){
+  var app1 = koa();
+  app1.context.message = 'hello';
+  var app2 = koa();
+
+  it('should merge properties', function(){
+    app1.use(function *(next){
+      assert.equal(this.message, 'hello')
+      this.status = 204
+    });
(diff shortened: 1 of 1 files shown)


lib/application.js
```diff
@@ -89,13 +95,32 @@ app.callback = function(){
...
 };
 
 /**
+ * Initialize a new context.
+ *
+ * @api private
+ */
+
+app.createContext = function(req, res){
+  var context = Object.create(this.context);
```
robinpokorny: Why is the `Object.create` called again? The object with that prototype was created on L44.

I tested it and it seems to work without the second creating well.

The same applies to request and response.

(I was looking at the code, found this line's history and went to this PR.)

PR #388, opened 2015-01-06
add more error message when passed a non-generatorFunction to use

dead-horse:


tj:
hmm we might be better off using the inspect util for some cases, might be overkill since the lineno is going to be there anyway

dead-horse:
it is easy to exceed the limit in test cases. lol

jonathanong:
maybe do some wonderful https://github.com/tj/callsite stuff?

dead-horse:
I've tried to do some magic with `callsite`, but at last I found that the most important information to users is what object they passed to `app.use`.

jonathanong:
@dead-horse what do you mean?

tj:
I still think this impl is a little weird since we'd get `[object Object]` in some cases etc, util.format or whatever is probably better in that sense but then it starts to get super overkill IMO

jonathanong:
hmmm i'd like to just remove this assertion...

dead-horse:
close for now, no significant promotion.

PR #390, opened 2015-01-15
recommend iojs

jonathanong:
one thing i'd like to wait for is for nvm/travis to support iojs and start testing on that. i could also push  this now

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-
+iojs ?= node --harmony-generators
 SRC = lib/*.js
 
 include node_modules/make-lint/index.mk
@@ -9,15 +9,13 @@ TESTS = test/application \
 	test/response/*
 
 test:
-	@NODE_ENV=test ./node_modules/.bin/mocha \
+	@NODE_ENV=test $iojs ./node_modules/.bin/mocha \
 		--require should \
-		--harmony-generators \
 		$(TESTS) \
 		--bail
 
 test-cov:
-	@NODE_ENV=test node --harmony-generators \
-		node_modules/.bin/istanbul cover \
+	@NODE_ENV=test $iojs node_modules/.bin/istanbul cover \
 		./node_modules/.bin/_mocha \
 		-- -u exports \
 		--require should \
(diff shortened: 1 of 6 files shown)


jonathanong:
shit i forgot the bash command...

jonathanong:
nvm i don't know what i'm doing haha

mattma:
After I installed `iojs`, i got a command `iojs`. At the same time, it aliased my system `node` executable to `iojs`. Essentially, when you run `node xx`, it called `iojs` executable. 

```bash
/usr/local/bin/node -> iojs
```

Should we stick with `node`, by default, `iojs` will be called when user execute the `node`. 

Well, I have not tested in `nvm` cases, that may need to run `iojs` explicitly.

jonathanong:
but the commands don't run with `node --harmony-generators` if it's iojs. what i thought i was doing was something like `alias iojs ?= node --harmony-generators` but apparently i forgot how to bash

mattma:
```bash
alias iojs='node --harmony-generators '

 ➜ which iojs
iojs: aliased to node --harmony-generators

 ➜ iojs xxx.js
node: bad option: --harmony-generators
```

From `iojs` docs, 

The current behaviour of the `--harmony` flag on io.js is to enable staged features only. After all, it is now a synonym of `--es_staging`.

I think the correct Bash command should be 

```bash
# put it in .bashrc, or .zshrc
alias iojs='node --harmony '
```

Test the command above

```bash
 ➜ which iojs
iojs: aliased to node --harmony

 ➜ iojs xxx.js
# run and log stdout...
```

jonathanong:
how do you only alias `iojs` if it doesn't exist yet

mattma:
After some tests on this matter, I think this could be a solution.

```bash
if [ ! -x "$(command -v iojs)" ]
then
  alias iojs='node --harmony '
fi
```

read more about `command` [here](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/command.html)

tj:
typically you do `NODE ?= node` and run `make` with NODE=iojs set etc

jonathanong:
i'd like tests to run without iojs users having to do `NODE=iojs` :D

tj:
why? there's only a few people that work on the tests, or we can default it to iojs instead

jonathanong:
yeah we can default to iojs

PR #393, opened 2015-01-22
Fix url.format taking into account path on node@0.11.15

ruimarinho:
`node@0.11.15` was released with a patch that adds support for `path` to url.format (https://github.com/joyent/node/commit/d312b6d15c69cf4c438ed7d884e6396c481a57f6). This means that koa 0.15.0 is effectively incompatible with this node version - the next build on travis will fail for sure without a patch like this.

However, this completely broke `npm install` for git+ssh urls and as such has been reverted on io.js, making it work like on `node@0.11.14` again (https://github.com/iojs/io.js/commit/913addbff5481567262c387cef9594f809e4ef83).

I'm not sure if my solution is acceptable, but it is cross-compatible with `node@0.11.14`/`node@0.11.15`/`io.js@1.0.3`.

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -117,6 +117,8 @@ module.exports = {
   set path(path) {
     var url = parse(this.req);
     url.pathname = path;
+    url.path = null;
+
     this.url = stringify(url);
   },
 
@@ -167,6 +169,8 @@ module.exports = {
   set querystring(str) {
     var url = parse(this.req);
     url.search = str;
+    url.path = null;
+
     this.url = stringify(url);
   },
 


dougwilson:
I'm moving my comments from https://github.com/koajs/koa/pull/394 here

lib/request.js
```diff
@@ -117,6 +117,8 @@ module.exports = {
   set path(path) {
     var url = parse(this.req);
     url.pathname = path;
+    url.path = url.pathname + (url.search || '');
```
dougwilson: This is not the best solution; I had similar changes in my modules and the best solution is to do `url.path = null;` here.

lib/request.js
```diff
@@ -166,7 +168,9 @@ module.exports = {
 
   set querystring(str) {
     var url = parse(this.req);
-    url.search = str;
+    url.search = !!~str.indexOf('?') ? str : '?' + str;
+    url.path = url.path + url.search;
```
dougwilson: Just `url.path = null;`

zensh:
if I do something：

```js
this. querystring = querystring1
this.path = path
// ....
this. querystring = querystring2
```

jonathanong:
god damnit @dougwilson i'm just going to add your to the organization :P

jonathanong:
@dougwilson do you know what's up with 0.11.15? did someone file an issue?

dead-horse:
@dougwilson knows everything :D

dougwilson:
> i'm just going to add your to the organization :P 

lol

> do you know what's up with 0.11.15? did someone file an issue?

as far as i know the change was intentional, even though kind of crappy. the reason for the change is because `http.get` runs it's options through `url.format`, thus mixing concerns a lot :/ basically they changed `url.format` to prefer `options.path` if there, and of course `url.parse` will always have `.path`, so... :( imo we can just set path to `null` and be done with it; it also makes what we're passing into `url.format` a little less ambiguous anyway.

ruimarinho:
All fixed. Thanks @dougwilson 🙏 for the wise input. @jonathanong io.js has reverted this change, so I suspect node 0.11.16/0.12 will do the same.

dougwilson:
Really, who knows what node.js is going to do.

dead-horse:
can we merge this to fix the CI now?

ruimarinho:
@dead-horse, from my side I think so. @dougwilson's already reviewed the PR.

dougwilson:
i.m.o this change should be merged :) I've had this fix out in Express stuff for a while now in production without issue :) It seems more important for koa since people are actually going to use 0.11.x series, haha.

jonathanong:
lol this is so lame. does anyone need a new release ASAP?

hallas:
:helicopter:

ruimarinho:
We don't need it right now since we're moving directly to io.js :-)

tedeh:
Ugh, this completely broke my client's website. Of course none of the ubuntu ppa provide older versions than the latest of node 0.11 so I cant really downgrade *moving to io.js*

dougwilson:
Sounds like @tedeh is looking for a koa release with this fix :)

pepkin88:
@tedeh have you tried [n](https://www.npmjs.com/package/n)?

PR #394, opened 2015-01-22
Add io.js to .travis.yml

ruimarinho:
Needed to make it easier to test against different versions of node/io.js binaries locally and took the chance to add interim support for travis as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,14 @@
 language: node_js
 node_js:
   - "0.11"
+env:
+  global:
+    - IOJS_VERSION="v1.0.4"
+  matrix:
+    - BIN="iojs" FLAGS=""
+    - BIN="node" FLAGS="--harmony-generators"
+before_install:
+  - "[ $BIN != 'iojs' ] || (wget https://iojs.org/dist/${IOJS_VERSION}/iojs-${IOJS_VERSION}-linux-x64.tar.xz && sudo tar xJf iojs-${IOJS_VERSION}-linux-x64.tar.xz -C /opt && rm iojs-${IOJS_VERSION}-linux-x64.tar.xz)"
+  - "[ $BIN != 'iojs' ] || (sudo ln -s /opt/iojs-${IOJS_VERSION}-linux-x64/bin/iojs /usr/local/bin/iojs && sudo chmod +x /usr/local/bin/iojs)"
 script: "make test-travis"
 after_script: "npm install coveralls@2.10.0 && cat ./coverage/lcov.info | coveralls"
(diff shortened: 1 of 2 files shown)


dead-horse:
`nvm` already supported `iojs`,  travis-ci didn't upgrade there `nvm` yet?

popomore:
We can wait travis supporting it, and then config can be simple.

https://github.com/travis-ci/travis-ci/issues/3108

lib/request.js
```diff
@@ -117,6 +117,8 @@ module.exports = {
   set path(path) {
     var url = parse(this.req);
     url.pathname = path;
+    url.path = url.pathname + (url.search || '');
```
dougwilson: This is not the best idea; I had similar changes in my modules and the best solution is to do `url.path = null;` here.

lib/request.js
```diff
@@ -166,7 +168,9 @@ module.exports = {
 
   set querystring(str) {
     var url = parse(this.req);
-    url.search = str;
+    url.search = !!~str.indexOf('?') ? str : '?' + str;
+    url.path = url.path + url.search;
```
dougwilson: Just `url.path = null;`

jonathanong:
@ruimarinho can you take out the commits to the other PR from this PR?

jonathanong:
let's wait until travis updates nvm. not really worth adding all these hacks to the travis file. we know koa works on iojs right now...

dougwilson:
i'm constantly checking travis ci for support by rerunning https://travis-ci.org/expressjs/serve-static/jobs/47808983 over and over until it works.

ruimarinho:
@dead-horse no, it's taking them years to update nvm 😢

@jonathanong I can remove the travis config and keep the env settings on the makefile to make it easier to do cross-platform tests locally, defaulting to io.js perhaps? In any case, I just wanted to see how travis would behave with io.js. Personally, I'd be fine adding it even if it gets removed next week :)

ruimarinho:
@jonathanong rebased and pushed an update for io.js 1.0.4. Travis still hasn't updated nvm yet. Let me know if you still think it's better to close the PR.

jonathanong:
not really down with the travis.yml hack. i'd personally rather wait until travis actually supports iojs. @koajs/owners thoughts?

dougwilson:
I also agree on waiting, at least that is my plan AppVeyor already does io.js; I don't know what is taking Travis CI so long.

juliangruber:
waiting seems reasonable to me as well!

hallas:
I agree, wait :-)

ruimarinho:
Alright, closing :) thanks all!

jonathanong:
@ruimarinho wanna make a new PR for iojs? :D

ruimarinho:
@jonathanong done :)

PR #399, opened 2015-01-25
Add res.append(field, val) to append headers

dead-horse:
ported from express https://github.com/strongloop/express/commit/f6f78e5f02afb874e916af27e30f56e82e23878c


dead-horse:
@dougwilson you should port all the features from express. :D

lib/response.js
```diff
@@ -425,6 +425,33 @@ module.exports = {
...
+    var value = val;
+    if (prev) {
+      // concat the new and prev vals
+      value = Array.isArray(prev) ? prev.concat(val)
+        : Array.isArray(val) ? [prev].concat(val)
+        : [prev, val];
+    }
+
+    return this.set(field, value);
+  },
```
tj: not a big deal but I'd just do:

```js
var prev = this.get(field);

if (prev) {
  val = Array.isArray(prev)
    ? prev.concat(val)
    : [prev].concat(val);
}

return this.set(field, val);
```

unless I'm missing a case there

dougwilson: I think @dead-horse used an intermediate variable because there is a perf hit in v8 when reassigning the value of a function argument (though I think that may only be the case when also accessing `arguments` within the same function, I'm not sure).

dougwilson: Oh, it's a direct port :) I'd say that's why I choose to use an intermediate :) Just because I didn't want to be sure it was a per hit, haha.

jonathanong:
:+1: is the primary usecase links? if so, i think a `.link()` would make more sense. i don't think people should be setting cookie values themselves in koa :)

dougwilson:
Also the HTTP `Warning` header.

fengmk2:
:100:

dead-horse:
merge this and release a new version?
/cc @koajs/owners

PR #400, opened 2015-01-28
Add io.js support to .travis.yml

ruimarinho:
This will only take effect on the 3rd of February, which is when TravisCI plans to update their build environment.

http://docs.travis-ci.com/user/build-environment-updates/2015-02-03


jonathanong:
this won't work because iojs will throw when the `--harmony-generators` flag is passed

ruimarinho:
Ah, of course. I'll add back the makefile changes. We're defaulting to io.js for local dev, right?

jonathanong:
yup.

ruimarinho:
@jonathanong trying a different approach to see if we can make it work even if `nvm` is not used locally (e.g. `n` or homebrew is used). I'm using `which`, similar to the  `execPath` check.

dougwilson:
FYI, io.js support is coming to Travis CI on 2/3 (http://docs.travis-ci.com/user/build-environment-updates/2015-02-03/#Node.js-VM).

dougwilson:
lol, I completely did not read the top.

Makefile
```diff
@@ -1,23 +1,32 @@
...
 	test/context/* \
 	test/request/* \
 	test/response/*
 
 test:
-	@NODE_ENV=test ./node_modules/.bin/mocha \
+	@NODE_ENV=test $(BIN) $(FLAGS) \
+		./node_modules/.bin/_mocha \
 		--require should \
-		--harmony-generators \
```
fengmk2: why not simple way: always add `--harmony` flags?

ruimarinho: Since it is not required for io.js and so the majority of users won't run io.js with this flag, it ensures that the test suite is fully compliant.

dougwilson:
Either we jumped the gun, Travis CI is still rolling out the change, or they have reverted it, because the build fails: https://travis-ci.org/koajs/koa/jobs/49340154

jonathanong:
those fuckers

hallas:
haha

dougwilson:
Running a probe job on Travis I get:
```
$ nvm --version
0.20.0
```
which is the old, pre-io.js version :(

kriswill:
They have it in their docs: http://docs.travis-ci.com/user/languages/javascript-with-nodejs/

dougwilson:
@kriswill yes, but their instructions do not work, as evidenced by the failing Travis CI tests.

kriswill:
http://docs.travis-ci.com/user/build-environment-updates/2015-02-03/#Schedule

> This update was originally scheduled at 2015-02-03 15:00 UTC, but it has been rolled back. We will reschedule the update at a later date.

dougwilson:
Ah, cool. So my speculation that it could have been a rollback was right :) `iojs` did work for a small window today.

ruimarinho:
I hope it won't take them another 1.5 months for a new deploy...

dougwilson:
This is one problem with doing a giant monolithic deploy :)

kriswill:
yeah, one deploy every 2 months does seem like a long time.  A company that provides agile tools, and doesn't take their own medicine.

ruimarinho:
Yeah, pretty sure it wasn't nvm that caused any of the issues they must've detected :/ this could be rolled out the next day it was released without side effects.

PR #401, opened 2015-01-29
Change request's "ips" regex to match others.

Foxandxss:
So I am in this silly quest of cloning the repo, deleting all the code on it and rewriting it again step by step to pass the tests. It is a acceptable way to learn how koa works.

Anyway, I saw that the regex used in `ips` doesn't match the others on the file, so I thought it would be good to change it.


tj:
hahaha awesome, LGTM, thanks!

PR #403, opened 2015-01-29
Remove unused imports and exports

Foxandxss:
Well, I finished my rewrite of Koa (I know how it works now!) and I saw that there were some code that are not used. I could be wrong (too much hours rewriting this) but I don't think so... :P


jonathanong:
thanks!

PR #409, opened 2015-02-05
Make `npm test` work with io.js

adamme:
Hey guys,

I understand that supporting io.js in `.travis.yml ` is a WIP, which surely will make some changes to Makefile, but this commit still makes sense if you need to run `npm test` locally.

However,  the ``` `node -v` == v0* ``` part is ugly. Any better idea? :)

Thanks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,12 +11,12 @@ TESTS = test/application \
 test:
 	@NODE_ENV=test ./node_modules/.bin/mocha \
 		--require should \
-		--harmony-generators \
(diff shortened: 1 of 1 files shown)


targos:
You could also just use the `--harmony` flag, which works with io.js

adamme:
@targos Thanks for pointing that out! :) Damn, I was being silly. A new commit has been merged into this PR.

dead-horse:
@adamme can you squash the commits?

PR #412, opened 2015-02-08
Parse Content-Type with "content-type" instead of "media-typer"

### LINKED ISSUE #406: Replace media-typer with content-type

dougwilson:
The `media-typer` is going to eventually start parsing MIME types, rather than the HTTP Content-Type syntax. The `content-type` module is taking over this role :)

closes #406


dead-horse:
awesome

PR #413, opened 2015-02-08
onerror: Retain CORS headers on error

### LINKED ISSUE #411: Retain CORS headers in onerror
Right now, when you call `this.throw()` if the error is not handled, then koa returns a generic error to the user. Nothing wrong with that except if you included CORS headers in a middleware. Those are currently being stripped out, causing any result be unreadable in the browser (so to speak).

This is a side effect from an open issue in koa-cors here: https://github.com/evert0n/koa-cors/issues/17

If you're dealing with CORS, then being unable to trust in koa to handle the error can provide undesired effect. You'd always have to catch the error yourself and be extra careful not to accidentally throw errors that koa will catch.

I think that instead of removing all headers here: https://github.com/koajs/koa/blob/master/lib/context.js#L120 that we should instead reset it while copying all CORS related headers over. That way, the programmer doesn't have to be paranoid about errors.

I can provide a sample pull request if this is acceptable solution.

TheThing:
Fixes #411

* Any errors will now properly pass along CORS headers
* Add unit test for it

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -121,7 +121,7 @@ var proto = module.exports = {
     var corsHeaders = {};
 
     for (var key in this.res._headers) {
-      if (key.indexOf('Access-Control-') === 0) {
+      if (key.indexOf('access-control-') === 0) {
         corsHeaders[key] = this.res._headers[key];
       }
     }
--- a/lib/context.js
+++ b/lib/context.js
@@ -116,8 +116,19 @@ var proto = module.exports = {
       return;
     }
 
-    // unset all headers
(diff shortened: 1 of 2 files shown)


jonathanong:
wait a minute. how are you handling errors? `.onerror()` is a last-resort error handler.

TheThing:
That is a fair assessment that most project should implement their own error handling. I just assumed with the reply I got from #411 that also being able to rely on koa would be acceptable.

jonathanong:
yeah i didn't think about it much when i wrote that :P closing. thanks!

Issue #391, opened 2015-01-19
Redirect 302/303


chrisveness:
I know it's not profoundly important, but wouldn't it be better for the default redirect status to be 303 rather than 302?

Qv http://stackoverflow.com/questions/4764297, http://stackoverflow.com/questions/5129076.

jonathanong:
i wouldnt mind. is there some sort of consensus on this? anyone else care?

tj:
> Note: Many pre-HTTP/1.1 user agents do not understand the 303 status. When interoperability with such clients is a concern, the 302 status code may be used instead, since most user agents react to a 302 response as described here for 303.

Sounds about right to me, I've never seen the opposite behaviour but we're probably well beyond being safe to move to 303 as well, so I guess it doesn't really matter (unless someone can find a real case)

chrisveness:
That note is from [RFC 2616 (§10.3.4)](http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html), from 1999. See also [RFC2616 is Dead](https://www.mnot.net/blog/2014/06/07/rfc2616_is_dead), [RFC 7231](http://tools.ietf.org/html/rfc7231#section-6.4.4). I know, I’m just a pedant, but if there’s a time to do it right, it’s when Koa is young.

tj:
yikes ok that's old, SGTM!

dougwilson:
These days, don't refer to RFC 2616; RFC 7230 to 7235 are the main replacement RFCs :)

ruimarinho:
From RFC7231, it looks like 307 should be used instead of 302, not 303. I think that the latter should be reserved for situations where one wants to explicitly reference a new location for a new resource:

> If the result of processing a POST would be equivalent to a
   representation of an existing resource, an origin server MAY redirect
   the user agent to that resource by sending a 303 (See Other) response
   with the existing resource's identifier in the Location field.

chrisveness:
@rulmarinho - not as I read the spec:

> [303 (See Other)] is primarily used to allow the output of a POST action 
> to redirect the user agent to a selected resource, since doing so provides 
> the information corresponding to the POST response in a form that can 
> be separately identified, bookmarked, and cached, independent of the 
> original request.

whereas

> The 307 (Temporary Redirect) status code indicates that the target
> resource resides temporarily under a different URI and the user agent
> MUST NOT change the request method if it performs an automatic
> redirection to that URI.

[...]

PR #418, opened 2015-02-27
make default redirect status 303 in place of 302

chrisveness:


jonathanong:
:+1: needs votes from the community!

gabeio:
Reading on http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#10.3.4 this will break any redirects for older browsers that use http/1.0.

Personally I would rather this be a setting. If you know users are going to be using older devices or you want to support older devices it might be nice to not break this backward compatibility for no reason. As both act nearly identical.

dougwilson:
It also breaks HTTP/1.1 browsers that didn't (yet) understand 303.

dougwilson:
But I believe those browsers are so old, that people who need to support them can always change the setting back. I think the main thing is does the 302 -> 303 change itself make sense.

gabeio:
Well currently looking at the code; it's not a setting is my point I'd rather see this become a setting that can be switchable and then I would agree that changing to 303 will be an okay change as they are basically the same thing.

chrisveness:
As @dougwilson says, it's a default which can be easily overridden. The most recent references I can find relate to [Windows XP](http://support.microsoft.com/kb/947513) and [IE 3.0 / Netscape 2.0](http://stackoverflow.com/questions/535053).

chrisveness:
Historical aside: [Alan Flavell](http://www.alanflavell.org.uk/www/post-redirect.html) (writing 10/15 years ago) sums things up rather well, to my mind: *Historically, in HTTP/1.0, status 302 (temporary relocation) was documented to call for the response that is now, in HTTP/1.1, documented for status 307. However, the popular browsers did not implement that specification: instead, they implemented the response that is now documented for status 303.*

fengmk2:
303 is fine, but 302 is `de facto standards`, most people only know 302 and 301.
Agree with @gabeio , make redirect status to optional, defaut is 302.
Then people know about 303 they can set default status to 303.

jansauer:
:+1: 303 sounds like a good compromise between 302 and 307. But i agree that 307 would be the right solution based on the rfc. #391

tunnckoCore:
303 redirects don’t appear to work properly to search engines :-1: 
http://bloggingmax.net/303-redirect-and-its-effect-on-seo/ - article from February 2015

I cant understand why you need 303? And why actually is 302 not a 301.

It should be configurable with options or somthing

gabeio:
@tunnckoCore 301 is a redirect that the browser will learn and attempt not to make the request again.
a 302 the browser will continue to keep asking the server the same request and will not learn the redirect.
303 is just a new 302...

tunnckoCore:
> 303 is just a new 302

just, must, bust, hust.. no matter.
I cant understand whats the problem is to be setting, not by default, when SEOs stated that its not good thing. :)

fengmk2:
I like keep `302` status for tmp redirect by default. 

:-1:

hallas:
Let's keep it at `302` for now.

PR #426, opened 2015-03-29
test: improve test coverage for application and response

fengmk2:
from 98% to 100%

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -204,6 +204,7 @@ function *respond(next) {
   if (null == body) {
     this.type = 'text';
     body = this.message || String(code);
+    /* istanbul ignore else */
     if (body) this.length = Buffer.byteLength(body);
     return res.end(body);
   }
--- a/lib/application.js
+++ b/lib/application.js
@@ -204,8 +204,7 @@ function *respond(next) {
   if (null == body) {
     this.type = 'text';
     body = this.message || String(code);
-    /* istanbul ignore else */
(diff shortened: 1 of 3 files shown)


fengmk2:
app.respond on `null == body` can't be coverage, because `body` never be null or empty string.

```js
  // status body
  if (null == body) {
    this.type = 'text';
    body = this.message || String(code);
    if (body) this.length = Buffer.byteLength(body);
    return res.end(body);
  }
```

dougwilson:
If it can truely never be, then why is the check necessary at all?

fengmk2:
@dead-horse maybe we need to delete `if (body)` detect.

lib/application.js
```diff
@@ -204,6 +204,7 @@ function *respond(next) {
   if (null == body) {
     this.type = 'text';
     body = this.message || String(code);
+    /* istanbul ignore else */
     if (body) this.length = Buffer.byteLength(body);
```
fengmk2: code always be Number, so body never be empty string.

jonathanong: cheater!!! :D

jonathanong: yeah let's delete the `if ()` :+1:

fengmk2:
all done @dougwilson @jonathanong

dougwilson:
LGTM

jonathanong:
LGTM

fengmk2:
Landed 8b14b91bae7fe9302eaefaecbacb67a1f51e494c

PR #427, opened 2015-03-30
make sure req.host and req.hostname return strict string

fengmk2:
It will more convenient to use `this.host` as a strict string, like `this.host.split('.')`.

Otherwise we need to detect `this.host` is string or undefined first.

```js
if (typeof this.host === 'string') {
  // dosomething ...
} else {

}
```

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -212,7 +212,7 @@ module.exports = {
     var proxy = this.app.proxy;
     var host = proxy && this.get('X-Forwarded-Host');
     host = host || this.get('Host');
-    if (!host) return;
+    if (!host) return '';
     return host.split(/\s*,\s*/)[0];
   },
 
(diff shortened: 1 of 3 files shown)


dead-horse:
LGTM

hallas:
I like it. LGTM

targos:
+1

fundon:
:+1:

juliangruber:
you can still do `!this.host` since `"" == false` so +1 from me

jonathanong:
:+1:

juliangruber:
although, hm this is a breaking change. if someone did `this.host === null` or something this will break. so, i'm for waiting for the next major release or dismissing this change.

fengmk2:
@juliangruber we can release 0.19.0 before 1.0.0 publish.

tj:
sounds fine to me, although in that case you have you'll just get `['']`which is probably less desirable

jonathanong:
should i release a 0.19?

fengmk2:
please!

jonathanong:
released, then added you and @dead-horse to npm so you guys can make releases too!

PR #431, opened 2015-04-06
use of eqeqeq instead of eqeq

fundon:


tj:
thanks but no need! other values won't coerce to 'HEAD' etc magically so it's not a big deal, typeof is always a string and so on

fundon:
okay, agreed!

PR #434, opened 2015-04-14
non-error thrown, fixed #432

### LINKED ISSUE #432: Assert in ctx.onerror not caught
I had a promise that rejected with a string as an error instead of a proper Error object. There is an [assert in context.js](https://github.com/koajs/koa/blob/c307c81c4313e13ed8464aafbccd1544e8ecf3a8/lib/context.js#L106), but it's never caught in [application.js](https://github.com/koajs/koa/blob/c307c81c4313e13ed8464aafbccd1544e8ecf3a8/lib/application.js#L128) so both the error string and the assert are silently swallowed and no response is sent to the client.

dead-horse:
don't block the response

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -7,7 +7,6 @@ var createError = require('http-errors');
 var httpAssert = require('http-assert');
 var delegate = require('delegates');
 var statuses = require('statuses');
-var assert = require('assert');
 
 /**
  * Context prototype.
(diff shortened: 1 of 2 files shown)


jonathanong:
:+1:

jonathanong:
you wanna make a release? or i can

dead-horse:
i'll release 0.19.1

fengmk2:
+1

PR #436, opened 2015-04-16
Update doc example to use ctx.state

soygul:
Avoids confusion..


dead-horse:
LGTM, it matches `this.state.user` in the previous chapter.

PR #438, opened 2015-04-28
make sure helpers return strict string

dead-horse:
not sure if this is look good to everyone, but it reduce lots of code like:

```
(this.charset || '').toLowerCase();
(this.get('ua') || '').indexOf('chrome');
(this.type || '').match(/^text/);
```

and this won't break your app if your are using `!this.get('referer')` etc.

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -301,9 +301,9 @@ module.exports = {
 
   get charset() {
     var type = this.get('Content-Type');
-    if (!type) return;
+    if (!type) return '';
 
-    return contentType.parse(type).parameters.charset;
+    return contentType.parse(type).parameters.charset || '';
   },
 
   /**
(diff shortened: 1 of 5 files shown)


jonathanong:
LGTM. can't see any possible breaking changes. anyone else?

targos:
LGTM but I think there are missing `return '';` in request.get(field)

juliangruber:
LGTM

juliangruber:
Strictly it's still a breaking change because we didn't tell people to use lose equality checks, if someone is strictly comparing `typeof this.get('referer') === 'undefined'` it breaks. i'm not sure though if someone does that even.

fengmk2:
LGTM

tj:
:+1:

dead-horse:
0.20.0 ?

jonathanong:
go for it!

PR #443, opened 2015-05-21
return same object from request.query

aheckmann:
Before this change, calling request.query when there was no querystring
resulted in a new object created and returned each time. If the resulting
object was ever changed, accessing request.query would not reflect it and
cause weird bugs.

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -131,8 +131,6 @@ module.exports = {
 
   get query() {
     var str = this.querystring;
-    if (!str) return {};
-
     var c = this._querycache = this._querycache || {};
     return c[str] || (c[str] = qs.parse(str));
   },
--- a/test/request/query.js
+++ b/test/request/query.js
@@ -7,6 +7,13 @@ describe('ctx.query', function(){
       var ctx = context({ url: '/' });
       ctx.query.should.eql({});
     })
+
+    it('should return the same object each time it\'s accessed', function(done) {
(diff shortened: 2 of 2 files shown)


tj:
SGTM, I think we have `this.querystring` returning `''` by default now so we could probably just remove the condition

test/request/query.js
```diff
@@ -7,6 +7,12 @@ describe('ctx.query', function(){
       var ctx = context({ url: '/' });
       ctx.query.should.eql({});
     })
+    it('should return the same object each time is\' accessed', function(done) {
```
reyawn: I might be missing something, but why escape the single quote?

tj: engrish! "it is"? haha, space above the `it` would satisfy my OCD as well :D

aheckmann: hahaha fixing

lib/request.js
```diff
@@ -131,7 +131,7 @@ module.exports = {
 
   get query() {
     var str = this.querystring;
-    if (!str) return {};
+    if (!str) str = '';
```
reyawn: It might be clearer to just do:
```
var str = this.querystring || '';
```

aheckmann:
ok done. thanks for the feedback

tj:
@koajs/owners anyone else have feels?

jonathanong:
:+1: I have good feels

Issue #444, opened 2015-05-24
update supertest to ^1.0.1


jonathanong:
there are some breaking changes with supertest. if someone wants to become a contributor, you could fix the tests :)

same with `should`, but i care much less about that

PR #445, opened 2015-05-24
upgrade supertest to ^1.0.1

yorkie:
Just help to close the #444 in my way, upgrade the `supertest` version and make the test cases be green.

/cc @jonathanong


yorkie:
Thanks for merge @dead-horse :)

PR #448, opened 2015-06-02
allow content type specification on this.throw

pkoretic:
This is useful when someone wants to return application/json error messages to clients using this.throw

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -115,11 +115,13 @@ var proto = module.exports = {
       return;
     }
 
+    var contentType = this.type
+
     // unset all headers
     this.res._headers = {};
 
-    // force text/plain
-    this.type = 'text';
+    // force text/plain if not specified
+    if (!contentType) this.type = 'text';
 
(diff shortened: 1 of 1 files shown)


juliangruber:
That is the default error handler, you can see it will abort if you did your own error handling here already: https://github.com/pkoretic/koa/blob/patch-1/lib/context.js#L113.

So I guess you can do this:

```js
app.on('error', function(err){
  this.body = 'your body';
  // etc...
});
```

and `app.onerror` won't respond any more.

pkoretic:
@juliangruber  that doesn't send anything to client, using koa 0.21
both throw and this.throw get catched in app.on('error') but this.body or this.status is ignored?

jonathanong:
@pkoretic you're going to have to fix the tests before i look at it more closely!

pkoretic:
we solved this as follows https://github.com/koajs/koa/wiki/Error-Handling
I don't want to pursue this anymore

PR #451, opened 2015-06-17
add this.req check at querystring() method

AlexeyKhristov:
If connection drop by client-side and use post-query logging (for example winston wrapper), then cause a error "Cannot read property 'url' of undefined" at parse() method

This pull request fix a problem

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -154,6 +154,7 @@ module.exports = {
    */
 
   get querystring() {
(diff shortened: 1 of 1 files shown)


jonathanong:
can you add a test and an explanation in the test case?

jonathanong:
8804b7ba6f7d85626970d1ea636ac002488d9095

PR #462, opened 2015-07-18
add slack badge

Pana:
FYI


hallas:
We already have Gitter. We should have either or -- not both?

jonathanong:
@Pana do you own the slack group? would prefer slack over gitter, etc.

PR #468, opened 2015-08-20
Do not log on expected http errors

sterlinghw:
When using `this.assert` and `this.throw`, I don't think it is appropriate to log the http error since it is a completely expected error. For example, 400 client errors (missing username and such) are things I would like the client to know about, but I definitely don't want it to be included in my logs. Checking the `expose` property allows filtering of expected errors, but still maintains error logs for errors that one might be interested in like actual code errors or other 500 errors.

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -163,7 +163,7 @@ app.createContext = function(req, res){
 app.onerror = function(err){
   assert(err instanceof Error, 'non-error thrown: ' + err);
 
-  if (404 == err.status) return;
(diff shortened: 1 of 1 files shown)


dead-horse:
LGTM

jonathanong:
391650518f877e7a2631f38b4dc8659a1d1fbb8c

dougwilson:
FWIW Express 5.0 is destined to make the same change.

PR #472, opened 2015-08-23
remove anti pattern

gyson:
remove `yield* next` in lib/application, which caused annoy `A promise
was converted into a generator …` message.

benchmark result:
* when bench with native Promise, it has no impact for both stable and
experimental ones.
* when bench with Bluebird, it’s about 5-10% faster than original for
both stable and experimental ones.


jonathanong:
nice!

Issue #473, opened 2015-08-25
API for extending context


cesarandreu:
As an application developer, I want to extend the context with app globals, such as database instance, model instances, and other client instances (e.g. s3, redis, etc). I prefer this approach over actual globals or singletons because those options are more difficult to test.

I made [koa-add-to-context](https://github.com/cesarandreu/koa-add-to-context), a middleware that mixes in keys into the request's context. However, since these values are all static, instead I could just do something like:

```js
Object.assign(app.context, {
  models: models,
  db: db,
  s3: s3
})
```

This would be more efficient, since it lets me avoid a middleware on every request. The problem is that I'm not sure if `app.context` is considered a public API or not. If it's not, I'd propose we add an API for enhancing the application context. Thoughts?

magicdawn:
+1

I think there are plugins & middleware, plugins just do extend request/reponse/context stuff. And they 
> avoid a middleware on every request

https://cnodejs.org/topic/553f8c242bd4939b1e905587

jonathanong:
when we were building this, @tj didn't want for it to be public. i use it all the time, but i don't want people to abuse it. @koajs/owners ?

yorkie:
-1 on extending the shared object request/response/context without any sandbox. I used to share something by using `context.{my-namespace}.{custom-key}`.

fengmk2:
Extend `app`, `context`, `request` and `response` before your app instance start to listen a port.

```js
const app = koa();
// ... add yours middlewares whatever

// extend whatever you want
app.db = createMySuperDBClient();

app.context.render = myReader;

// or pass the app to your koa modules
someFeatureModule(app);
```

juliangruber:
it seems like a standardized approach will cause more conflicts than everyone doing it their own way

tj:
yea extending the context is nicer than having middleware plop stuff on the context, so I'm +1 for having the object doc'd

fundon:
+1, I like this way that outside the middleware.

aheckmann:
and here i thought it was already public haha +1

On Thu, Aug 27, 2015 at 11:01 AM, Fangdun Cai <notifications@github.com>
wrote:

> +1, I like this way that outside the middleware.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/koajs/koa/issues/473#issuecomment-135508059>.
>



-- 
Aaron
@aaronheckmann <https://twitter.com/#!/aaronheckmann>
soundcloud.com/ajhecky
[...]

PR #475, opened 2015-08-31
Add docs for app.context

travisjeffery:


travisjeffery:
Per https://github.com/koajs/koa/issues/473, we already have tests for `app.context` here: https://github.com/koajs/koa/blob/master/test/application.js#L1003-L1029

fengmk2:
+1

fundon:
+1

PR #478, opened 2015-09-09
Update travis Node.js version & remove io.js

demohi:


dead-horse:
i think we'd better keep iojs in travis

demohi:
@dead-horse  I added it back

.travis.yml
```diff
@@ -3,7 +3,8 @@ node_js:
   - "0.12"
   - "iojs-v1"
   - "iojs-v2"
-  - "iojs-v3"
+  - "iojs-v3"
```
dougwilson: Looks like there is some accidental trailing whitespace added to the v3 line.

demohi: @dougwilson  yes..   thanks

dead-horse:
can you squash(`git rebase master`) these commits?

targos:
Wouldn't it be better to put `4` so that it automatically installs the latest 4.x?

fengmk2:
landed 520163fe57e8ea6888e6116f2ea841213608eff3

Issue #476, opened 2015-09-03
add support for ctx.origin


matthewmueller:
Shorthand for: `this.protocol + '//' + this.host`

jonathanong:
@fengmk2 @dead-horse didn't we have another issue similar to this?

hallas:
I remember something with host + port but not this in particular

dead-horse:
https://github.com/koajs/koa/blob/master/lib/request.js#L69

matthewmueller:
yah, I think origin is probably more useful for servers talking with servers

jonathanong:
PR anyone?

chentsulin:
I submit one PR for this.

dead-horse:
@matthewmueller https://github.com/koajs/koa/pull/480

do you need to publish a new version to npm?

jonathanong:
merged https://github.com/koajs/koa/pull/480

PR #480, opened 2015-09-20
implement ctx.origin

chentsulin:
implement `ctx.origin` for #476


lib/context.js
```diff
@@ -176,6 +176,7 @@ delegate(proto, 'request')
   .access('query')
   .access('path')
   .access('url')
+  .access('origin')
```
dead-horse: `.getter('origin')`

chentsulin: Thanks, already fixup.

dead-horse:
LGTM

Issue #447, opened 2015-05-25
Add a silent option


cesarandreu:
I do a lot of my development in `test` environment, i.e. run mocha with `-w` flag and write code with tests. But when I get errors JS errors it's annoying, because koa doesn't log errors when `'test' == this.env`.

I'd propose adding an `app.silent` option, and using that instead of `'test' == this.env` to determine if errors should be printed to stderr by default. The initial value for `app.silent` can be set to the result of `'test' == this.env` so it doesn't break backwards compatibility.

This is not only useful for errors, though. `app.silent` can be used from your logging middleware, to determine if messages should be sent to stdout / stderr or not.

Adding this option shifts the concern to the application developer. If I want, I can have whatever set of conditionals I want, e.g. `app.silent = process.env.SILENT || process.env.CI`.

Thoughts?

jonathanong:
talking about https://github.com/koajs/koa/blob/0ad06c9810079174660fa4948ca183d9b90efd3c/lib/application.js#L167? you should be using your own error handler instead during testing, which is what i do.

tj:
in general I'm +1 for anything that isn't env-specific, envs should define config not the other way around so I'm down

jonathanong:
reading it more, i'm cool with an `app.silent` option. @cesarandreu wanna make a PR?

tejasmanohar:
@jonathanong This also benefits me! If this hasn't been PR'd or worked on yet by @cesarandreu, I'd love to do it as an intro to contributing to KoaJS.

Just to clarify, the `app.silent` option would replace the env-specific checks in https://github.com/koajs/koa/blob/0ad06c9810079174660fa4948ca183d9b90efd3c/lib/application.js#L167, right? Would/should there be any defaults based on your env? I was thinking if you're in test, the silent mode would be on by default but dev could override w/ the option (i.e. use this.silent if passed but if not, just default value of `silent` based on if this is running in 'test' env)

tejasmanohar:
I implemented it without any env-based defaults in PR #486 since @tj said

>in general I'm +1 for anything that isn't env-specific, envs should define config not the other way around so I'm down

That said, I can easily add env-based defaults if that's preferred.

tejasmanohar:
This issue can now be closed since #486 has been merged // cc @cesarandreu

jonathanong:
https://github.com/koajs/koa/pull/486 thanks!!!!

Issue #367, opened 2014-11-19
Middleware logging... how to?


soggie:
Hi guys, I'm looking at `koa-i18n` and it uses `debug` as a means to log information. I think this is sub-optimal and makes it hard for koa users to control what kind of information gets printed.

My suggestion is to provide an `info` and `debug` method alongside `throw` in `ctx` so that there is a standard way for middlewares to bubble up messages and as an end-user of koa.js you can choose your own logging library to handle the logs.

What do you guys think?

hallas:
I'm leaning towards providing a suite of log methods like warn, info, log, debug and having them emit a log event on the application which comes with the log record and the context. That way people can listen on the event and log it the way they need to.

cmwelsh:
+1 with a default log configuration to write logs to stderr. Koa should be the central dispach for middleware logs and even console output if we want to foster an ecosystem.

fengmk2:
Summary above, there're two ways for logging:

- Event

```js
app.on('log', function (log) {
   // log.level = string, log level
   // log.message = string
   // log.ctx = ctx
});
```

- logger instance
 - methods: `debug()`, `info()`, `warn()`, `error()`
 - `app.logger`
 - `ctx.logger`, alias to `app.logger`

wenshiqi0:
I prefer event,not the instance.Make a chance to write logger on my own,and feel much more fun.Actually I still think to make a middle ware for package the request ,response and ctx. = = Why not

tj:
I'd vote for structured if this is something that people want to see in the core. printf-style is too hard to work with in practice, basically useless for querying so I wouldn't want to promote that out of the box

yanickrochon:
I agree with @tj. The middleware would allow, just like `koa-view`, to use logging adapters more easily (such as winston, etc.), and would not tie this functionality to the core.

What I don't like about events is the unnecessary steps :

1. user calls `app.log` (or `app.emit('log', ...)`)
2. koa forwards to the even listener**s** (thus make koa responsible for such listeners)
3. listener receive events and user-custom call their logger
4. logger dispatch message

However, with a middleware, we get rid of steps 2 and 3 altogether :

1. user calls `app.logger.log` or `this.logger.log`, however the middleware is implemented.
2. logger dispatch message

Thus, completely separate logger from core and even allow users to implement their own middlewares.

aheckmann:
[...]

Issue #219, opened 2014-02-20
profile middleware  


tj:
just a reminder to myself: use compose hook to create a profiler layer that will report durations spend in each layer

jonathanong:
let's just solve this once in for all in `koa-compose`. how about something like this:

```js
function compose(middleware, profiler){
  return function *(next){
    if (!next) next = noop();

    var i = middleware.length;

    while (i--) {
      var fn = middleware[i]
      var profile = profiler(middleware[i])
      next = function *(next) {
        profile.start()
        yield* next
        profile.end()
      }.call(this)
    }

    return yield *next;
  }
}
```

jeromew:
(Replying to https://github.com/koajs/koa/issues/196#issuecomment-192083703)

What I would like to do with profiling is to have a way to profile whole koa application, even if I am using middlewares that have already been composed in other libraries. In production, you never know where a bottleneck is going to come from so a view on other modules would be a +.

the profiling system should also have close to no impact in production if profiling is not activated.

The way I see it (I have no implementation and am not sure if it makes sense) would be something like the `debug` module.

a module exporting middlewares could have a profiling name a-la debug.

Then at runtime via 2 environment variables, I could 

 * choose by name which middlewares I want to profile
 * define which profiler I want to use

I don't know if this makes sense to you.

Regarding the modification of koa-compose, I once proposed that we should be able to define our own composer (use `koa-compose` if you want no profiling, use `koa-compose-myprofiler' to have the behavior of myprofiler.

Adding the `profiler` variable directly to `koa-compose` adds some of this behavior but :

 * there is a systematic perf hit on the profiler mechanism even if I want no profiling
 * I don't see how I could profile third-party koa middlewares modules

Regarding the API, I proposed something in https://github.com/koajs/koa/issues/196#issuecomment-61021512 (maybe the code is not very clear). Basically, the idea to compose `middleware` was to

   * first map the `middleware`array to myprofiler(middleware[i])

````js
    var myprofiler = function *myprofiler(next){
       // things to do before
       yield *gn.call(this, next);
       // things to do after
    }
````

  * and then call `compose` on this new array

The difference I see is that you version is synchronous (there is no way to wait on `profile.start()` and `profile.end()`)

The API should probably give the user the choice of sync vs async. 
 * I could see a usage for a "step by step" profiler/debugger where I need to click "next" at every stage. 
 * I could also see scenarios where I could want to make sure something is logged before going to the next step even if it at a perf cost

[...]

PR #486, opened 2015-10-05
app.silent option to turn off err logging

tejasmanohar:
Solves #447 

// cc participants- @jonathanong @tj @cesarandreu

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -166,6 +166,8 @@ app.onerror = function(err){
 
   if (404 == err.status || err.expose) return;
   if (this.silent) return;
+  // DEPRECATE env-specific logging in v2
+  if ('test' == this.env) return;
 
   var msg = err.stack || err.toString();
   console.error();
--- a/lib/application.js
+++ b/lib/application.js
@@ -165,7 +165,7 @@ app.onerror = function(err){
   assert(err instanceof Error, 'non-error thrown: ' + err);
 
   if (404 == err.status || err.expose) return;
-  if ('test' == this.env) return;
+  if (this.silent) return;
 
   var msg = err.stack || err.toString();
   console.error();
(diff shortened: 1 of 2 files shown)


tj:
lgtm

tejasmanohar:
@tj hm, somewhere else as in #447 or elsewhere? (sorry, pretty new to koa)

tj:
I think it's scattered a bit, some in https://github.com/koajs/koa/issues/367, and https://github.com/koajs/koa/issues/219 is somewhat related. I'll leave this open for a little if anyone else wants to give some feedback but it works for me! thanks man

tejasmanohar:
alrighty.

are we going to consider this a breaking change? i guess it does change what logs when and based on what... so maybe breaks the API?

travisjeffery:
Works for me too. I'm okay wither way w.r.t. to versioning, but based on expierence with other projects people tend to complain a lot even if you change logging and don't follow semver strictly.

tejasmanohar:
@travisjeffery yeah, i think a major version would be a good idea... like 1.1

tj:
Since it's just test related I think it would be fine, unless we leave that 'test' == check in until 2.x, doesn't really hurt

tejasmanohar:
@tj Ah ok, you're right- would you like me to add the 'test'' env check back in AFTER the `this.silent` check? that would make it a non-breaking change.

tejasmanohar:
i.e.
```javascript
if (this.silent) return;
if ('test' == this.env) return;
```
so codebases relying on the 'test' env to not log won't be affected by this change. in v2, we can remove the 'test' env check and make people use the silent option? @tj

tj:
sounds great! maybe throwing in a comment so we remember to deprecate it later `// DEPRECATE (v2)` or something

tejasmanohar:
done- 6c19c41. yay for semver!

fengmk2:
LGTM.

PR #489, opened 2015-10-05
ensure all JS files have blank newline at top/bot

tejasmanohar:
Just a few files don't have a blank newline at the top... though that's really nice when editing this in `vim` to skip up to the top and that's why I assume it's there in the rest of the files? (may be wrong).


tejasmanohar:
One file was missing a blank newline at the bottom so added it in 5e21238594340e9ea1c85aaaedeecf06830b33db

PR #492, opened 2015-10-09
Template strings

tejasmanohar:
PR to replace ugly string concatenation w/ nice ES6 template strings (interpolation)... I confirmed that tests pass on commit 2ae5757 (before modifying tests/*) as well as `HEAD`.

### DIFF
--- a/benchmarks/experimental/async.js
+++ b/benchmarks/experimental/async.js
@@ -8,7 +8,7 @@ app.experimental = true;
 // number of middleware
 
 var n = parseInt(process.env.MW || '1', 10);
-console.log('  %s async middleware', n);
+console.log(`  ${n} async middleware`);
 
 while (n--) {
   app.use(async function (next){
--- a/benchmarks/middleware.js
+++ b/benchmarks/middleware.js
@@ -6,7 +6,7 @@ var app = koa();
 // number of middleware
 
 var n = parseInt(process.env.MW || '1', 10);
-console.log('  %s middleware', n);
+console.log(`  ${n} middleware`);
 
 while (n--) {
   app.use(function *(next){
--- a/lib/application.js
+++ b/lib/application.js
@@ -163,7 +163,7 @@ app.createContext = function(req, res){
  */
 
 app.onerror = function(err){
-  assert(err instanceof Error, 'non-error thrown: ' + err);
+  assert(err instanceof Error, `non-error thrown: ${err}`);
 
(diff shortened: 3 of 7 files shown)


tejasmanohar:
Agh snap! Looks like Node.js v0.12 doesn't support ES6 template strings with or without `--harmony`... however, io.js did at the time v0.12 was latest :cry: 

Maybe we should keep this open and possibly use for v2.0 of Koa? I imagine by that time we'd be thinking Node 4.0+? Thoughts? @tj

thelinuxlich:
Template strings are much slower

tejasmanohar:
@thelinuxlich if you look at the changes in context, not sure how much it'd matter... benchmarks?

thelinuxlich:
http://kpdecker.github.io/six-speed/

tejasmanohar:
@thelinuxlich Hm, fair. In context (please check the minor changes made), do you think the performance loss outweighs readability benefits? This kinda pushes me saying use Babel, use Babel, etc...

targos:
There are benchmarks in the project. You can start by running them before and after your changes.

tj:
lgtm, I'm assuming ops/s for simple string stuff is well in the millions so even if it's a tad slower that probably doesn't matter

jonathanong:
https://github.com/koajs/koa/commit/65ed40f40f0ec3bb15f9d132a5effff1cfca96a1!

felixfbecker:
I also don't think template strings need to be used _everywhere_. If you just concat something like `'Invalid error code' + code` there is no need to invoke a template function. It is only beneficial if you got a huge string with lots of values or when you use tagged template strings.

targos:
if a template string isn't tagged, it is desugarized by V8 to string concatenation. There is no function invocation. I personally think it improves readability to use them, even for small concatenations.

tejasmanohar:
Template strings > string formatting > concatenation from a readability perspective. Tagged template strings have little to no use to us in koa itself anyways :)

felixfbecker:
From MDN:
 > The default function just concatenates the parts into a single string

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/template_strings

 > The static String.raw() method is a tag function of template strings, similar to the r prefix in Python or the @ prefix in C# for string literals. It's used to get the raw string form of template strings (that is, the original, uninterpreted text).

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/raw

PR #493, opened 2015-10-09
Use arrow functions where there's benefit

tejasmanohar:
So... arrow functions _can_ be slower (though not considerably, I don't think), and except cleanness, there's not much reason to use them when there's no .bind, assigning `this` above the fn, or another workaround in place.

As a result, I'm just PR'ing a change to use arrow functions in 1 location where there is benefit because `var self = this;` is meh.

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -117,13 +117,12 @@ app.callback = function(){
   var fn = this.experimental
     ? compose_es7(this.middleware)
     : co.wrap(compose(this.middleware));
-  var self = this;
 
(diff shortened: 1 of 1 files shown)


tejasmanohar:
Meh, I guess the same goes for this! Though arrow functions aren't in v0.12 w/ `-- harmony` heh

jonathanong:
https://github.com/koajs/koa/commit/dc2c197d903e5a7a43a049174f0a5e018a2299d2

tejasmanohar:
thx

PR #498, opened 2015-10-11
ES 2015

jonathanong:


targos:
It's ES2015 :wink:

jonathanong:
oh... i was going to add the es2016 stuff here too. i guess maybe we should do an es2015 and es2016

tejasmanohar:
@jonathanong es2016 stuff is far and long from coming to node natively though, no?

jonathanong:
if people care about performance, we can push it all and transpile with babel. don't think the performance difference will be much.

otherwise, we can split it up into es2015 and es2016. i'll change this one to es2015

tejasmanohar:
>if people care about performance, we can push it all and transpile with babel. don't think the performance difference will be much.
yeah, performance difference isn't that big in this case, but here are stats for reference- http://kpdecker.github.io/six-speed/.

that said, regardless of performance, my point was es2016 features coming natively in node stable without flags are kinda far off. not all of es2015 is good-to-go in node yet w/o `--harmony`... so we kinda *have to use babel* if we want es2016, which requires more discussion.

tejasmanohar:
@jonathanong maybe we should also make new branch (off es2016) called `es2015`, delete `es2016` from origin, and open new PR? current name may be a bit misleading

jonathanong:
https://github.com/koajs/koa/pull/499

PR #499, opened 2015-10-11
ES 2015

jonathanong:


dead-horse:
LGTM

tejasmanohar:
thoughts on es6 classes? requires a major API change- i.e. you _must_ use the `new` keyword afaik

```
var app = koa();
/// becomes
var app = new koa();
```

(i think, that's one of the ES2015 features I haven't used much)

classes don't really offer any benefits - 100% sugar - questioning whether it's a good refactor or not.

felixfbecker:
Not related directly but any comments on why koa exports prototypes in source code? I find that very confusing.

lib/request.js
```diff
@@ -3,12 +3,12 @@
...
  */
 
-var contentType = require('content-type');
-var stringify = require('url').format;
-var parse = require('parseurl');
-var qs = require('querystring');
-var typeis = require('type-is');
-var fresh = require('fresh');
+const contentType = require('content-type');
+const stringify = require('url').format;
```
ruimarinho: Can't this be deprecated in favor with es6 template strings?

tejasmanohar: @ruimarinho can't exactly tell what you're pointing to here. which line/statement/dependency?

tejasmanohar:
@felixfbecker hm. in which context?

tejasmanohar:
another thing to note- `class Application extends Emitter` would (roughly) translate to `Object.setPrototpyeOf(Application, Emitter)` rather than `Object.setPrototypeOf(Application.prototype, Emitter.prototype);` -> https://github.com/koajs/koa/blob/master/lib/application.js#L56

felixfbecker:
in context.js for example

tj:
@felixfbecker nothing special, just separates the protos into different files, they're used here https://github.com/koajs/koa/blob/master/lib/application.js#L47

felixfbecker:
it just confused me when I wanted to unit test middleware and tried to somehow create a context manually to call the middleware with. Could not get it to work.

tejasmanohar:
@felixfbecker ah, hmm... not sure what the problem would be there. maybe open another issue w/ isolated, runnable example?

lib/request.js
jonathanong: `stringify = require('url').format`? not sure how that works

jonathanong:
LOL so many merge conflicts now

tejasmanohar:
aghh!

tejasmanohar:
someone needs to write a regex / pattern to solve these.... lol

juliangruber:
ETOOMANYCONFLICTS

tejasmanohar:
@jonathanong how'd you solve the conflicts so fast? :tada:

tejasmanohar:
@jonathanong Is this basically our v2 branch? i.e. Should all changes to be reviewed for v2 be sent here (else, where? master isn't ideal.)?

jonathanong:
@tejasmanohar yeah for now. i think branching off `v1.x` from `master` and making this master would be a good idea too.

tejasmanohar:
@jonathanong aight.

>i think branching off v1.x from master and making this master would be a good idea too.
i agree

jonathanong:
merging into master!

tejasmanohar:
alright, then classes will come in a separate PR to master when i get some free time :)

tejasmanohar:
@jonathanong is there a reason we have a separate branch for v1 rather than just a git tag? https://github.com/koajs/koa/tree/v1.x.

jonathanong:
@tejasmanohar in case we need to maintain it more

tejasmanohar:
@jonathanong aha makes sense

PR #501, opened 2015-10-11
use strict mode to unlock more es2015 features

tejasmanohar:
classes + block-scoping is only available in strict mode. also just enforces some more good code rules... that we're passing since e717733aa86efa8639745de5d179a611b5431d79

### DIFF
--- a/benchmarks/experimental/async.js
+++ b/benchmarks/experimental/async.js
@@ -1,4 +1,6 @@
 
+'use strict';
+
 const http = require('http');
 const koa = require('../..');
 const app = koa();
--- a/benchmarks/experimental/index.js
+++ b/benchmarks/experimental/index.js
@@ -1,4 +1,6 @@
 
+'use strict';
+
 // support async await by babel
 require('babel/register')({
   optional: ['asyncToGenerator']
--- a/benchmarks/middleware.js
+++ b/benchmarks/middleware.js
@@ -1,4 +1,6 @@
 
+'use strict';
+
 const http = require('http');
 const koa = require('..');
 const app = koa();
--- a/benchmarks/run
+++ b/benchmarks/run
@@ -1,3 +1,6 @@
+
+'use strict';
+
 #!/usr/bin/env bash
 
 echo
--- a/lib/application.js
+++ b/lib/application.js
@@ -1,4 +1,6 @@
 
+'use strict';
+
 /**
  * Module dependencies.
  */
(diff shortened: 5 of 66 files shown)


jonathanong:
damn you're on a roll!!!

tejasmanohar:
@jonathanong hehe, thx! will be giving a talk on Koa on Wednesday :sunglasses:

jonathanong:
ooooo where at?

tejasmanohar:
@jonathanong local meetup- http://www.meetup.com/nashjs/events/224842082/. ~1h long... will build a mini dynamic blog like [koablog-mongo](https://github.com/marcusoftnet/koablog-mongo). your [campjs talk](https://www.youtube.com/watch?v=bNVcVx8TfPc) helped me understands lots of koa :100:

tejasmanohar:
anyhow, i'll get someone to take a video and put it up on YT

tejasmanohar:
what i find really strange is that `const` is working fine without strict mode though the error of using `let` in strict mode is
`SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode`
and, the docs also say `const` is only available in strict mode https://nodejs.org/en/docs/es6/ :question: 

i assume we'll probably end up using strict mode anyways... so don't think it matters but still strange

targos:
The spec says that both `const` and `let` should work in sloppy mode. However in V8 only `const` is currently supported. It's true that the error message is a bit misleading.
There is an issue tracking this: https://code.google.com/p/v8/issues/detail?id=3305
PR to update node doc: https://github.com/nodejs/new.nodejs.org/pull/230

On the upcoming Node.js 5.0.0, which will include V8 4.6, `let` will be available in sloppy mode behind the flag `--harmony-sloppy-let`.

tejasmanohar:
@targos aha, makes sense! new something was up with the node docs there... clear mismatch, but nice job validating w/ spec and PR - i was too slow on that haha

to me, strict mode sounds like a good idea for koa regardless of the es2015 features

targos:
> to me, strict mode sounds like a good idea for koa regardless of the es2015 features support in/out of it

I agree. To me, strict mode should be enabled everywhere.

jonathanong:
ugh. let's do this against master! was gonna try to cherry-pick into master but there were too many conflicts. @tejasmanohar do you want to do this? otherwise, i can

tejasmanohar:
@jonathanong i'll do it

PR #504, opened 2015-10-11
switch `const` for `var` in `test/application.js`

tejasmanohar:
Looks like I missed this when merging


tejasmanohar:
bleh wrong branch

PR #505, opened 2015-10-11
switch `const` for `var` in `test/application.js`

tejasmanohar:
Looks like I missed this when merging earlier


jonathanong:
https://github.com/koajs/koa/commit/0f241572d69fb8aa7585842319ffb6462b50915a

PR #506, opened 2015-10-11
update minimum node version in README for ES2015

tejasmanohar:
- dropped iojs since there was the whole node foundation merge
- upped node v to 4.0.0


jonathanong:
https://github.com/koajs/koa/commit/e60c563ba71b8aabd8746b8c26ca23b83da91236

PR #512, opened 2015-10-11
ES2015 Classes - Tracking PR

tejasmanohar:
After a load of merge conflicts, here's the updated tracking PR (not ready for merge) for classes. Note that only Application has been refactored so far- Context, Response, and Request are pending. Also, this is a *big* breaking change since all Koa 1.x.x apps won't work due to `koa()` rather than `new koa()`.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -44,8 +44,8 @@ $ npm install koa
 ## Example
 
 ```js
-const koa = require('koa');
-const app = koa();
+const Koa = require('koa');
+const app = new Koa();
 
 // logger
 
--- a/benchmarks/experimental/async.js
+++ b/benchmarks/experimental/async.js
@@ -2,8 +2,8 @@
 'use strict';
 
 const http = require('http');
-const koa = require('../..');
-const app = koa();
+const Koa = require('../..');
+const app = new Koa();
 
 app.experimental = true;
 
--- a/benchmarks/middleware.js
+++ b/benchmarks/middleware.js
@@ -2,8 +2,8 @@
 'use strict';
 
 const http = require('http');
-const koa = require('..');
-const app = koa();
+const Koa = require('..');
+const app = new Koa();
 
 // number of middleware
 
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -30,7 +30,7 @@ $ node my-koa-app.js
 
 ```js
 var koa = require('koa');
-var app = koa();
+var app = new Koa();
 
 app.use(function *(){
   this.body = 'Hello World';
@@ -56,7 +56,7 @@ app.listen(3000);
 
 ```js
 var koa = require('koa');
-var app = koa();
+var app = new Koa();
 
 // x-response-time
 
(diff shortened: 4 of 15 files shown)


jonathanong:
> Also, this is a big breaking change since all Koa 1.x.x apps won't work due to koa() rather than new koa().

is there a way to avoid this? i remember there's `new.target` added to the spec, but that might be ES2016. we could also just wrap the export in a function.

tejasmanohar:
@jonathanong no good way that i know of. maybe a hacky way- here's what i heard from #javascript in IRC- https://gist.github.com/tejasmanohar/fc4315b1c7474933c232
> no
> if you need that, you can't use "class".
> you can but that creates other problems
> mainly that `instanceof` won't work and the prototype won't match
> basically it means you don't expose the "class", you just expose a factory function. *but* i wouldn't do that.

... which makes sense. doesn't sound optimal to me... but easy to implement

tj:
I'm down for branching off 1.x and master == 2.x to turf the koa() fn

tejasmanohar:
I'm ± 0 on ES2015 classes tbh. main points-
- -1 -> covers up the underlying proto structure (especially w/ inheritance) which is meh 
- +1 -> _minor_ readability benefits

I don't recommend exposing a factory function... other unintended consequences. If there was something _by deisgn_ in ES2015 that allowed us to pass on the `new` keyword, then I'd :+1: it

jonathanong:
it would just suck writing stuff like:

```js
const app = new (require('koa'))();
```

but i guess that's irrelevant with ES6 modules come:

```js
import Koa from 'koa'
const app = new Koa()
```

tejasmanohar:
>I'm down for branching off 1.x and master == 2.x to turf the koa() fn
As @tj said, we're bumping up to 2.x.x so I don't think this is a problem.

If we're going to use classes, it doesn't make sense to hide it w/ a factory function as @ljharb says in IRC. IMO, the question is more whether or not we should really refactor this to use classes.

tejasmanohar:
@jonathanong Yeah, indeed. ES6 imports aren't in Node 4 yet... guessing they'll come in Node 5.x.x but still be 2nd-class to `require()` since they don't offer any real benefits (as far as I can tell) until static analysis is actually used to help performance.

tejasmanohar:
@jonathanong 
`const app = new (require('koa'))();`
but really, should one be doing this anyways/regardless of the `new` keyword? even `const app = require('koa')()` isn't great imo.
```
const Koa = require('koa')
const app = new Koa()
```
like in the example is ideal formatting-wise...

jonathanong:
problem is i write the one-line solution everywhere lol. but eventually we're all going to have to migrate to the two line solution anyways because of es6 modules so i don't mind forcing `new`

tejasmanohar:
@jonathanong eh... "_eventually_" there may be a long long time (if ever- same w/ classes). remember, es6 modules aren't first-class. you can only dynamically import things with `require()`

personally, i'm interested in hearing +/-/± 0/1 on ES6 classes from more ppl in the community.

tj:
never hurts to do the simplest thing, ditching the `if (!(this instanceof ...)` hack we have now regardless of `class` is fine with me, not a big deal either way it doesn't change any functionality

tejasmanohar:
fair enough

PR #515, opened 2015-10-12
modularize tests

tejasmanohar:
test/application.js was a monster

### DIFF
--- a/test/context/toJSON.js
+++ b/test/context/toJSON.js
@@ -1,7 +1,7 @@
 
 'use strict';
 
-const context = require('../context');
+const context = require('.');
 
 describe('ctx.toJSON()', function(){
   it('should return a json representation', function(){
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ include node_modules/make-lint/index.mk
 
 REQUIRED = --require should --require should-http
 
-TESTS = test/application \
+TESTS = test/application/* \
 	test/context/* \
 	test/request/* \
 	test/response/* \
--- a/test/application/index.js
+++ b/test/application/index.js
@@ -1,0 +1,56 @@
+
+'use strict';
+
+const request = require('supertest');
+const assert = require('assert');
+const koa = require('../..');
+
+describe('app', function(){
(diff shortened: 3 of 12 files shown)


tejasmanohar:
hmm. any ideas on why tests and coverage checks are passing? all tests pass locally w/ `npm test` but not w/ `make test-travis`, and they were just re-arranged so not quite sure. i think it's something to do w/ the request testing and how i removed some of the require's when decoupling the "application" tests but not really sure.

jonathanong:
yeah i've seen that before locally. not sure.

jonathanong:
screen shot 2015-10-11 at 9 24 41 pm

this failed for me though

test/context/toJSON.js
```diff
@@ -1,7 +1,7 @@
 
 'use strict';
 
-const context = require('../context');
+const context = require('.');
```
jonathanong: `require('.')` might be a special cast for `process.cwd()`... not sure if that's still the case

tejasmanohar: hmm, as far as i can tell `'.'` is just shorthand for `'./'` in node's directory system... but may be wrong.

tejasmanohar:
@jonathanong hmm, tests pass locally on my end w/ `npm test` - is that what you're running them with, too? also, verify `git rev-parse HEAD` to see we're both on the same commit 4d2f88cbe263413c89a0c3097465cf4a141598d4?

on your `process.cwd()` note... does replacing `require('.')` w/ `require('./')` or `require('./index.js')` help with your tests? i would imagine if `'.'` was being treated as `process.cwd()` then, this would fail saying it can't find a module `index.js` in the root dir or resort to `lib/application.js` since that's the `main` in `package.json` which would make all the context tests fail though...

test/context/toJSON.js
targos: > hmm, as far as i can tell '.' is just shorthand for './' in node's directory system... 

That is correct.

tejasmanohar:
will retry and commit more frequently to see where/when travis fials

PR #517, opened 2015-10-12
Modularize Tests (complete)

tejasmanohar:


tejasmanohar:
@jonathanong Hm so, modifying the things w/ context's tests seems to mess things up w/ Travis (in a way that's really hard to debug- see build error)... but not w/ `npm test` on my local.

Your tests failed earlier when mine didn't so if you don't mind, could you run it again to see what the result is on this commit 7968d3c? I'll try to make some assumptions from there to figure out what's wrong.

tejasmanohar:
Hm so even with `const context = require('.')` swapped out for `const context = require('./index.js');` across `test/context/*.js` and `require('../context')` swapped w/ `const context = require('../context/index.js')` throughout `test/`, Travis seems to fail on first request (via `make test-travis`) so I don't think it's any issue of mismatching `process.cwd()` w/ shorthand `'.'` path in Node.

jonathanong:
yeah super weird... not sure what's going on. after a few tries, it starting working locally for me too, but not test-travis. might be some coverage thing that's going on.

tejasmanohar:
@jonathanong lol yeah, all up to the last commit, it didn't work on first try for me either w/ `npm test`. then, i changed some things and tried again... and boom, it started working. then, i started questioning how those changes made it work but couldn't get it back to failing with `npm test` and w/ `make test-travis` the error is vague... failing on first req.

tejasmanohar:
@jonathanong passing! @ljharb helped me nail it in IRC. we were making tests run `test/context/index.js` since that was previously not in the `context/` dir https://github.com/tejasmanohar/koa/blob/modularize_tests/Makefile#L8

after realizing that test/context.js is not actually a test... i decided to retract the commit and push back to my branch. travis now passes everything fine. i put the context file in helpers/ dir instead. sorry for force push there... shouldn't have done that.

anyways- thoughts on new layout?

jonathanong:
:+1: i will try to look at this tomorrow!

tejasmanohar:
nocie

PR #518, opened 2015-10-12
add myself (tejasmanohar) to the authors list

tejasmanohar:
didn't know this was here so updated it. is there a reason behind the AUTHORS file? seems like a hassle to manually update when https://github.com/koajs/koa/graphs/contributors already exists.


jonathanong:
there's a way to do this automatically... i don't know what it is!

slaskis:
@jonathanong if you use tjs [git-extras](https://github.com/tj/git-extras) you can just `git authors`...

tejasmanohar:
@jonathanong `git log --format='%aN <%aE>' | sort -f | uniq > AUTHORS`

jonathanong:
sweet

PR #519, opened 2015-10-12
Use object methods

slaskis:
I noticed you're doing all kinds of fun refactoring right now so I just wanted to suggest switching to shorthand functions (or "object methods" as they seem to be called in ES2015). 

For no other reason than that I think they align nicely with the `get x()`/`set x()` object shorthands.

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -24,7 +24,7 @@ const proto = module.exports = {
    * @api public
    */
 
-  inspect: function(){
+  inspect() {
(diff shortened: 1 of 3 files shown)


jonathanong:
:+1:

tejasmanohar:
nice! this already gives us some of the readability benefits of classes... all that's really left is a clear `super()`, `extend`, and `constructor()` i guess haha

Issue #495, opened 2015-10-11
Linting


jonathanong:
TODO:

- [ ] run linting in CI 
- [ ] add `eslint` and `babel-eslint` into this package so it lints ES6+
- [ ] add `.eslintrc` to this repo so editors lint
- [ ] add `.editorconfig`

jonathanong:
oh maybe there is linting... we need to add it to the CI

juliangruber:
no semicolons :P

dead-horse:
+1 for standard

tj:
that whitespace though :( `function hello (name) {`, looks so disjointed, 1run(function () { ... })`, ಥ_ಥ. I don't care about semi's but if anything the widely adopted implicit standard is to use them so I'd -1 that

tejasmanohar:
+1 on linter

-1 on standard.
- in @tj's examples, the spaces b/w `hello` and `(` and `function` and `(` kinda annoy me.
- semicolons are more popular/conventional... i guess they also help avoid edge cases though that shouldn't be much of a problem here.
- `For the ternary operator in a multi-line setting, place ? and : on their own lines.` also depends on context, imo so wouldn't want that in the linter

i think we should just use our own eslint configuration, there's babel-eslint too

stojanovic:
+1 for standard, not perfect but good enough.
For semicolons, it does require them in front of ( and [ so there should be no edge cases.
eslint config also works, it's probably easier in this case, because standard will require some changes and eslint can be adjusted to a current code style.

jonathanong:
we already have linting... i just don't know where the rules are coming from. you can run `make lint`. we just need to:

- update the rules or package so it works with the current code as well as ES6+. i get a lot of `Unexpected token ILLEGAL`s
- add rules so that linting is strict
- add the command to the CI

juliangruber:
it's https://www.npmjs.com/package/make-lint which depends on https://www.npmjs.com/package/eslint and has the defaults

tj:
hrmm repo is gone

tejasmanohar:
> For semicolons, it does require them in front of ( and [ so there should be no edge cases

@stojanovic i don't like inconsistencies. i know standard forces semicolons where needed (or, i actually thought it just warns against doing things where they are needed), but i'd rather not mix & match when things just work and are conventional w/ semicolons.... yet i don't have a strong preference either which way

stojanovic:
[...]

PR #520, opened 2015-10-12
eslint + editorconfig

jonathanong:
closes https://github.com/koajs/koa/issues/495

- [ ] update rules so that they are more inline with the current styles

we should be making minimal code changes. we're only extending from standard's rules since there's just a lot of good stuff there.


targos:
I can help with that

tejasmanohar:
do we want to just use standard instead? i refactored most of that in #524 and actually seems like a pretty good tool (though i don't mind either which way).

juliangruber:
i think just anything sane that doesn't require a massive diff will work. removing all the semis seems a bit too much for what it's worth

tejasmanohar:
@juliangruber `standard --format` removed them all for me

jonathanong:
whoa whoa whoa that diff is ridiculous. ideally we just adjust the rules so it makes our styles consistent. not change the whole thing to standard :P

jonathanong:
we should just make a "tj" config haha

tejasmanohar:
lol alright, fair enough

tj:
I just find it ironic that the "standard" is almost the complete opposite of what 90% of js devs do haha

tejasmanohar:
@tj lol true.
- semicolons are most popular
- function name () { } , pretty sure function name() { } is most popular
etc

targos:
@tj :+1: that's so true

kriswill:
you could use something like https://github.com/Flet/eslint-config-semistandard/blob/master/eslintrc.json, and relax rules to taste.

targos:
https://github.com/koajs/koa/pull/525

PR #526, opened 2015-10-13
refactor Application into an es6 class

tejasmanohar:
basically #512 without merge conflicts.

i only refactored application into es6 classes. request, response, and context don't really make sense as es6 classes to me since they don't have a constructor, i don't know why you'd `new` them... makes more sense as obj literal / proto to me. making a `new Koa()` does make sense semantically though (well, arguably since you probably wouldn't make multiple of them except in unit tests.. or pass args- sounds like a singleton). 

thoughts?

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -44,8 +44,8 @@ $ npm install koa
 ## Example
 
 ```js
-const koa = require('koa');
-const app = koa();
+const Koa = require('koa');
+const app = new Koa();
 
 // logger
 
--- a/benchmarks/experimental/async.js
+++ b/benchmarks/experimental/async.js
@@ -2,8 +2,8 @@
 'use strict';
 
 const http = require('http');
-const koa = require('../..');
-const app = koa();
+const Koa = require('../..');
+const app = new Koa();
 
 app.experimental = true;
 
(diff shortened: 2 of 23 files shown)


tj:
whoop!

tejasmanohar:
woohoo

targos:
I wonder if we shouldn't still export a factory function ? This is a big breaking change for users.
I personally don't care as I'm willing to update my current applications to use `new`.

```js
module.exports = function() {
  return new Application();
};
```

tejasmanohar:
@targos we could indeed. that was discussed here- https://github.com/koajs/koa/pull/512#issuecomment-147255796 ... look thru there and drop points here if you think the factory function is still ideal.

tejasmanohar:
@targos 
i don't like to make huge breaking changes either, but in a major version, i don't think it'd be unreasonable to ask users to use the `new` keyword especially when koa is cutting-edge and moving so fast (old users were probably mostly on harmony... i'm sure they're cool w/ change lol)

takeaway: imo, if we're going to return a factory, we probably should skip on classes because this
- breaks `instanceof` and proto
- is unconventional / negates the feature

targos:
Yeah I agree with you.
Please don't revert, I'm glad that there is at least one popular library that doesn't fear to go forward!

tejasmanohar:
:)

though whether classes are _forward_ or not is an interesting question... but enough there lol ;)

Issue #532, opened 2015-10-14
'Content-Type' hard not to send


rektide:
I'm working to implement a spec that calls for a [201 Created response with no body](https://tools.ietf.org/html/draft-ietf-webpush-protocol-00#section-5). Currently Koa is giving me a application/octet-stream with a content-length of 0, which seems a little silly. I'd like some way to not send a Content-Type.

Related is #277.

jonathanong:
that's weird. if there's a `content-length` of `0`, then there should not be a `content-type` because there is no body.

a PR for a fix or even just a test-case would be great (against v1.x)!

tejasmanohar:
I can take on a PR for this issue after replicating it. Will look tomorrow... that said, if you have an isolated runnable example / repo / gist, that would be great and speed up the whole process, @rektide!

danneu:
Reproduce with:

``` javascript
var koa = require('koa');
var route = require('koa-route');

var app = koa();

app.use(route.get('/test', function*() {
  this.status = 201;
  this.body = '';  // else koa turns it into 204 No Content
  this.type = null;
}));

app.listen(3000, () => console.log('Listening on 3000'));
```

```
$ http localhost:3000/test
HTTP/1.1 201 Created
Connection: keep-alive
Content-Length: 0
Content-Type: application/octet-stream
Date: Sun, 18 Oct 2015 00:30:37 GMT
```

Maybe the fix is as simple as changing this:

``` javascript
// lib/response.js
set type(type) {
  this.set('Content-Type', getType(type) || 'application/octet-stream');
},
```

to this:

``` javascript
// lib/response.js
set type(type) {
  if (null == type) { // handle this.type = null or undefined
    this.res.removeHeader('Content-Type');
  } else {
    this.set('Content-Type', getType(type) || 'application/octet-stream');
  }
},
```

tejasmanohar:
Yes, can replicate. On it :smile:

tejasmanohar:
#534 covered it

PR #534, opened 2015-10-19
Fix for #532. 'Content-Type' hard not to send.

ssotomayor:
Since it's impossible to know the order in which the user will set the response options, i'm removing the Content-Type header on *lib/application.js*. However the response will differ if the body is null or empty string. Is there any reason why *'application/octet-stream'* is the default type?

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -212,9 +212,6 @@ function respond() {
     return res.end(body);
   }
 
-  // remove content-type header if there is no content
-  if (0 == this.length) this.remove('Content-Type');
-
   // responses
   if (Buffer.isBuffer(body)) return res.end(body);
   if ('string' == typeof body) return res.end(body);
--- a/lib/application.js
+++ b/lib/application.js
@@ -212,6 +212,9 @@ function respond() {
     return res.end(body);
   }
 
+  // remove content-type header if there is no content
+  if (0 == this.length) this.remove('Content-Type');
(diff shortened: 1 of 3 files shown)


jonathanong:
can you add a test for this? otherwise, :+1:

dougwilson:
I am :-1: on this proposed fix, because it's perfectly valid to have a zero length response for types like `text/plain`, `text/csv` and many other types with valid zero length documents. i.m.o. the fix the the original issues should be achieved in a different way such that if a user explicitly sets a content type, it does not get removed just because the response is zero length.

fundon:
:-1: 
 https://github.com/koajs/koa/blob/master/lib/response.js#L282-L285

jonathanong:
well if @dougwilson says so... it depends on the content type then, right? JSON shouldn't have length 0. others may

jonathanong:
you should add this test case: https://github.com/koajs/koa/issues/532#issuecomment-148963938

but why not just remove the type if it's unknown? specifically when he sets `.type=null`. we can just never set binary as the type ourselves.

ssotomayor:
You prefer to remove the content-type on null? That could also be a solution, i'l update the PR later i'm a bit swamped with stuff right now.

ssotomayor:
Continued here: https://github.com/koajs/koa/pull/536

PR #536, opened 2015-10-19
Ssotomayor feature/532 content type hard not send

### LINKED ISSUE #532: 'Content-Type' hard not to send

jonathanong:
closes #532
closes #534

@dougwilson @fundon @ssotomayor how about this?

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -301,7 +301,12 @@ module.exports = {
    */
 
   set type(type) {
-    this.set('Content-Type', getType(type) || 'application/octet-stream');
+    type = getType(type) || false;
+    if (type) {
+      this.set('Content-Type', type);
+    } else {
+      this.remove('Content-Type');
+    }
   },
 
   /**
--- a/test/application/respond.js
+++ b/test/application/respond.js
@@ -34,6 +34,28 @@ describe('app.respond', function(){
     });
   });
 
+  describe('when this.type === null', function(){
+    it('should not send Content-Type header', function(done){
+      var app = new Koa();
+
+      app.use(function *(){
+        this.body = '';
+        this.type = null;
+      });
+
+      var server = app.listen();
+
+      request(server)
+        .get('/')
+        .expect(200)
+        .end(function(err, res){
+          if (err) return done(err);
+          res.should.not.have.header('content-type');
+          done();
+        });
+    });
+  });
+
   describe('when HEAD is used', function(){
     it('should not respond with the body', function(done){
       const app = new Koa();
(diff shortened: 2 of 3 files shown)


ssotomayor:
Seems in the same direction as [this](https://github.com/koajs/koa/commit/cd97fccdfda606b7a59992da94df39d07270bfb6#diff-d86a59ede7d999db4b7bc43cb25a1c11R305), however as i pointed out before, i'm not sure why is `application/octet-stream` as the default option, there are even tests checking that. If there is no logical reason i support the option of setting `null` as any header option to NOT send it, then the user decides.

fundon:
LGTM

jonathanong:
anyone else?

tejasmanohar:
LGTM. agree w/ @dougwilson's comment on the previous PR for this fix

juliangruber:
+1

fengmk2:
+1

dougwilson:
LGTM

tj:
@ssotomayor octet stream is a reasonable default I think, everything is an octet stream haha, but :+1:  for this

ssotomayor:
Just closed the other PR, i agree with the fixes. :+1:

evilive3000:
Hi. Is there a way to respond with an _undefined_ body and **DO NOT** set a _content-type_.

I'm trying to make an internal redirect by nginx depend on the headers that returned from koa's backend. All I need is only one field `X-Accel-Redirect`; but Koa appends `Content-Type: text/plain` header which will be transparently proxied through nginx. But if content-type were absent in koa's respond Nginx automatically decides what type it is and sets it correctly.

I found the code where the magic happened [application.js#L218](https://github.com/koajs/koa/blob/master/lib/application.js#L218):

```javascript
  // status body
  if (null == body) {
    body = ctx.message || String(code);
    if (!res.headersSent) {
      ctx.type = 'text';
      ctx.length = Buffer.byteLength(body);
    }
    return res.end(body);
  }
```
according to this I will always get non empty body?

evilive3000:
Right after posting prev comment I found that if I explicitly set body to null `ctx.body = null;` I receive 204 status code as expected. Wonder why it is not default behavior. When I don't set the body shouldn't it be treated as if I set it to `null`

Issue #281, opened 2014-05-07
integrate proxy-addr


tj:
re: https://github.com/visionmedia/expressjs.com/issues/152#issuecomment-42467655
or whatever we decide is the way to go in those discussions

https://github.com/expressjs/proxy-addr/

jonathanong:
@dougwilson what do you think about integrating this now?

jonathanong:
@dougwilson or are you still planning to release a `trust` module?

dougwilson:
`proxy-addr` will be split apart some more at some point, but AFAIK there isn't a reason `proxy-addr` cannot be used as-is currently in `koa`.

jonathanong:
:+1:

amit777:
Hi, we noticed an issue when there are potentially multiple proxies and app.proxy = true.  We occasionally end up with private IP addresses in the logs. I think the situation can happen in this example:

```
User (10.1.1.1) -> Corporate Web Proxy(1.1.1.1) ->  Our HAProxy(2.2.2.2) -> KoaApp (10.2.2.2)
```

in this scenereo, 10.1.1.1 is returned in this.ip instead of 1.1.1.1

Maybe we also could use another option called "app.filterPrivateIps = true"  which will actually filter out any RFC1918  ips from this.ips.  indutny's "ip" module has an .isPrivate() method that may be helpful.

amit777:
Oh cool, it looks like proxy-addr actually may solve the exact issue I described above.

ruimarinho:
@amit777 this may or may not be an option depending on your setup, but one alternative thing you can do is to filter those ips before they hit the koa app. nginx allows this via the [realip](http://nginx.org/en/docs/http/ngx_http_realip_module.html), i.e. to trust the `X-Forwarded-For` header by pre-setting a list of trusted ips (including private ones).

jonathanong:
o i c. proxy-addr doesn't do as much as i thoguht it would in its current form. right now, the benefit is minimal.

- https://github.com/jshttp/forwarded/pull/1
- parse all `X-Forwarded-*` headers with trust

amit777:
Just as a note, I found this in one of my logs even though I don't have ipv6 enabled on the server.  Seems like some strange proxy setting headers in a funny way.  The IP address at the end is my own public ip (though i've changed it to something random and invalid for anonymity).

::ffff:267.211.128.66

amit777:
Altenatively, maybe could have something like an array of trusted headers by priority...

[...]

PR #541, opened 2015-10-23
support x-real-ip

atian25:


slaskis:
I would move that code into `get ips()` instead.

atian25:
@slaskis  if `x-forward-for[0] != x-real-ip` then ?

slaskis:
well in your PR you prioritise `x-real-ip` anyway. So if you would add that first as the first option in `get ips()` it should have the same effect.

atian25:
ok, so `get ips()` should add `x-real-ip` when `x-forward-for[0] != x-real-ip`.
x-real-ip = 100.100.100.100
x-forward-for = 200.200.200.200,300.300.300.300
so `get ips()` = [100.100.100.100, 200.200.200.200,300.300.300.300] ?

tj:
sounds redundant, why not just use x-forwarded-for? from the cases I'm seeing x-real-ip == the first in x-forwarded-for

jonathanong:
i'd also prefer if this logic was in a separate module. see:

- https://github.com/koajs/koa/issues/281
- https://github.com/jshttp/proxy-addr
- https://github.com/jshttp/forwarded

atian25:
copy~

Issue #538, opened 2015-10-22
Update docs for new async (ctx, next) => middleware signature


jonathanong:

menems:
Maybe add usage of koa-convert for old signature?

jonathanong:
@menems sure!

PR #542, opened 2015-10-23
Add 2.0.0 Examples #538

menems:


tejasmanohar:
overall, :+1:

Readme.md
```diff
@@ -42,6 +42,78 @@ $ npm install koa
...
+// response
+
+app.use(function *(ctx){
+  ctx.body = 'Hello World';
+});
+
+app.listen(3000);
+```
+
+## Example with old version, Koa < 2
```
juliangruber: do we really want to keep koa@2 around in the main readme? I'd vote for linking to a wiki page for this purpose.

yanickrochon:
I'd advocate to promote using promises (or Babel) over Generators, since 2.0 wraps the middleware with `co.wrap` anyway, which adds some extra function calls. Also, the new API does not require middlewares to be async, so always using an async function is useless in some case.

yanickrochon:
Also, and if I'm not mistaken, the promise-based middleware needs to return something otherwise how can the previous middleware know when the promise is resolved or rejected? Wouldn't it's usage be

```
app.use(function (ctx, next){
  const start = new Date;
  return next(ctx).then(function() {
    const ms = new Date - start;
    console.log(`${ctx.method} ${ctx.url} - ${ms}`);
  });
});
```

instead? Or is the middle ware wrapped and handle everything automagically?

gyson:
yes, `return` is needed.

Also, based on latest spec, you can use `return next()` instead of `return next(ctx)`.

menems:
Ok i change the that , i haven't seen in compose 
```
return Promise.resolve(fn(context, function next() {
```

menems:
@yanickrochon it 's very dangerous to forget the return token, because the previous middleware is call even without the return , but not as expected (in async mode). As you say, the new api not require async functions so : no return -> sync function

yanickrochon:
@menems well, yes, obvously, since the `next` middleware is `yield`ed or `await`ed. I got that. I was just pointing out the promise example :)

Readme.md
jonathanong: :+1:

Readme.md
```diff
@@ -42,6 +42,78 @@ $ npm install koa
...
+// response
+
+app.use(function (ctx){
+  ctx.body = 'Hello World';
+});
+
+app.listen(3000);
+```
+
+## Example with generator
```
jonathanong: let's note here that this is a stop-gap solution - it will eventually be removed when `async` functions are native.

Readme.md
menems: So , if I understand, I remove the generator function example for new koa version or keep it with a mention saying : it will eventually be removed when async functions are native.?

jonathanong: keep it with a mention :)

menems: done

yanickrochon: I'd personally go with discouraging generator function from Koa 2.0. And start a migration process to eventually drop it altogether and remove `co` as a dependency. With Babel and promises (and sync'ed middlewares, there's really no need to support yet another async solution... which is hacky and was not meant for this situation. Generator functions not only lack debugging standard, but are totally misused (i.e. they confuse everyone as they were not meant as a control flow solution).

Readme.md
```diff
@@ -42,6 +42,80 @@ $ npm install koa
  - [Introduction Screencast](http://knowthen.com/episode-3-koajs-quickstart-guide/) - An introduction to installing and getting started with Koa
 
 ## Example
+```js
+const Koa = require('koa');
+const app = new Koa();
+
+// logger
+
+app.use(function (ctx, next){
```
felixfbecker: Shouldn't this use arrow functions if you're using them in line 76 aswell?

Readme.md
felixfbecker: Didn't we agree to remove co dependency? In that case it should be `app.use(co.wrap(function *(ctx, next) { ... }))`

Readme.md
tejasmanohar: yea

menems:
so according to #558, I removed the generator example
And use arrow function when it 's possible, (everywere :) )
It's ok?

yanickrochon:
Arrow functions are available out-of-the-box (w/o dev or es flags), so I'd say : yes. And the consensus seem to agree on the removal of `co`, so the example should show how to use the old (legacy) way by manually wrapping with `co` (not supplied with koa).

menems:
like that @yanickrochon ?

jonathanong:
i'll merge this now since overall its good. if you guys have any more requests, more PRs!

menems:
@felixfbecker fixed

felixfbecker:
I think it's fine like this.

yanickrochon:
@menems lol yes :) it's fine

tejasmanohar:
Looked over this again- should the Koa 1.x.x example really go in the README? I feel like that should just stay on 1.x.x branch and put a note that master / 2.x.x is under active dev / currently alpha.

felixfbecker:
@tejasmanohar thought about that too. You would have to select the 1.x.x tag in GitHub. Imo this is cleaner.

yanickrochon:
The problem, right now, is that it is not clear that the documentation is about Koa@2. IMO

felixfbecker:
Bluebird has this aswell. In there readme they have a note at the bottom that readme and docs for 2.x (current is 3.x) is maintained under that branch.

PR #547, opened 2015-10-24
Refactor application.js - use arrow function in callback and remove EventEmitter from 'event' module

stojanovic:


fundon:
Good catch!

juliangruber:
+1

tj:
:+1:, wonder why `(=> expr)` isn't legal :-1:

stojanovic:
@tj I guess because of the reserved words, ie. (async => expr) wouldn't work as you expect :)

tj:
hmm should I think that'd still be unambiguous since it's a keyword grammar-wise but I guess it does look confusing haha

stojanovic:
Ah, but `async` is not reserved keyword yet as far as I know.
You can do `let async = 'whatever'`.
Maybe it'll be in future, but even babel let you do that at the moment.

tejasmanohar:
that's actualyl not an issue w/ babel. it's actually intended. `async` is not going to be a reserved word / keyword due to backwards compatibility. `var async = require('async')` for example would break...

so i think it does make sense to do `() => ...` for argument-less functions (rather than `=> ...`)

stojanovic:
Yes, you are right, I read something about that and totally forgot.

tejasmanohar:
(that said, do note that everything isn't implemented correctly in babel... like let/const are still hoisted and binding of `this` in arrow functions isn't as straight-forward as appears)

PR #549, opened 2015-10-25
Refactor tests - add arrow functions

stojanovic:
Again arrow functions.
However, I changed only one line functions, and I left something like this as it was:
```
app.on('error', function(err){
  err.message.should.equal('boom');
  done();
});
```
Not sure if we are getting anything with this:
```
app.on('error', err => {
  err.message.should.equal('boom');
  done();
});
```
except maybe consistency (using arrow functions for all callbacks) and 4-6 characters less.

### DIFF
--- a/test/application/onerror.js
+++ b/test/application/onerror.js
@@ -9,9 +9,8 @@ describe('app.onerror(err)', function(){
   it('should throw an error if a non-error is given', function(done){
     const app = new Koa();
 
-    (function(){
-      app.onerror('foo');
-    }).should.throw(AssertionError, {message: 'non-error thrown: foo'});
+    (() => app.onerror('foo'))
+    .should.throw(AssertionError, {message: 'non-error thrown: foo'});
 
     done();
   });
@@ -22,9 +21,7 @@ describe('app.onerror(err)', function(){
 
     err.status = 404;
 
-    const output = stderr.inspectSync(function(){
-      app.onerror(err);
-    });
+    const output = stderr.inspectSync(() => app.onerror(err));
 
     output.should.eql([]);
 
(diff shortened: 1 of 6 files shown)


tejasmanohar:
I usually just use all arrow functions for arguments and such to be consistent, but correct- there's no benefit as you mentioned so maybe it's not worth the PR diff.

stojanovic:
There's one more "special case", this doesn't work:
```
ctx.res.end = () => throw new Error('response sent');
```
It needs to be written like this:
```
ctx.res.end = () => {
  throw new Error('response sent');
};
```
So I [left it unchanged](https://github.com/koajs/koa/blob/master/test/application/index.js#L36).

stojanovic:
@tejasmanohar I agree, but that should be a part of a separate PR since it wasn't like that earlier. I can do it but not sure if everyone thinks the same.

tejasmanohar:
@stojanovic Well, previously the prop `.should.be.ok` was directly chained to `)`, now it's on a new line so indentation is needed for it to be readable imo.

stojanovic:
@tejasmanohar moved to the same line. +1 for indentation everywhere for that situation.

test/application/onerror.js
```diff
@@ -9,9 +9,7 @@ describe('app.onerror(err)', function(){
   it('should throw an error if a non-error is given', function(done){
     const app = new Koa();
 
-    (function(){
-      app.onerror('foo');
-    }).should.throw(AssertionError, {message: 'non-error thrown: foo'});
+    (() => app.onerror('foo')).should.throw(AssertionError, {message: 'non-error thrown: foo'});
```
juliangruber: do we care about the 80 chars limit still?

stojanovic: I was trying to keep it under 80 chars, but there's a few longer lines so I just moved this to the same line after a comment.

jonathanong:
meh. the only reason i'd merge this is because future PRs for ES6 refactoring in inevitable.

if we make these sort of changes, can we update the eslint file with relevant rules?

stojanovic:
@jonathanong done.
There's one more rule, but it needs to be added in #551:
```
  prefer-arrow-callback: 2
```
cc. @broucz

jonathanong:
great :+1:

PR #553, opened 2015-10-25
Refactor test: use arrow function

broucz:


broucz:
Also, I didn't refactor so far function like:

```js
function escape(html){
  return String(html)
    .replace(/&/g, '&amp;')
    .replace(/"/g, '&quot;')
    .replace(/</g, '&lt;')
    .replace(/>/g, '&gt;');
}
```

to 

```js
const escape = html => {
  return String(html)
    .replace(/&/g, '&amp;')
    .replace(/"/g, '&quot;')
    .replace(/</g, '&lt;')
    .replace(/>/g, '&gt;');
};
```

@juliangruber can you confirm the need to refactor this also ?

stojanovic:
That could be:
```
const escape = html =>
  return String(html)
    .replace(/&/g, '&amp;')
    .replace(/"/g, '&quot;')
    .replace(/</g, '&lt;')
    .replace(/>/g, '&gt;');
```
But I don't think we should refactor them.

broucz:
Even:

```js
const escape = html =>
  String(html)
    .replace(/&/g, '&amp;')
    .replace(/"/g, '&quot;')
    .replace(/</g, '&lt;')
    .replace(/>/g, '&gt;');
```

and you're probably right, it's not strictly refactor.

tj:
lgtm, though I almost prefer the originally haha

broucz:
@tj it's just more or less trendy as I guess nobody will re-define `escape` in this file :)

test/application/onerror.js
```diff
@@ -5,24 +5,24 @@ const stderr = require('test-console').stderr;
...
 const AssertionError = require('assert').AssertionError;
 
-describe('app.onerror(err)', function(){
-  it('should throw an error if a non-error is given', function(done){
+describe('app.onerror(err)', () => {
+  it('should throw an error if a non-error is given', done => {
     const app = new Koa();
 
-    (function(){
+    (() => {
```
stojanovic: You don't need `{}` here, check comments in #549.

test/application/onerror.js
```diff
@@ -5,24 +5,24 @@ const stderr = require('test-console').stderr;
...
 
-  it('should do nothing if status is 404', function(done){
+  it('should do nothing if status is 404', done => {
     const app = new Koa();
     const err = new Error();
 
     err.status = 404;
 
-    const output = stderr.inspectSync(function(){
+    const output = stderr.inspectSync(() => {
```
stojanovic: Oneliner too, `{}` are not needed.

broucz:
@stojanovic yep sure and thanks for the review.
I wasn't sure if I had to remove them due to easier update & visibility concern. But according to #549 I'll start to update them and wait a bit for other feedback before push, just to be sure :).

But are we all agree that for `description(...)` level we can keep `{}`, as for example:

```js
describe('the request method is not GET and HEAD', () => {
  it('should return false', () => {
    const ctx = context();
    ctx.req.method = 'POST';
    ctx.fresh.should.be.false;
  });
});
```

can be also refactor in:

```js
describe('the request method is not GET and HEAD', () =>
  it('should return false', () => {
    const ctx = context();
    ctx.req.method = 'POST';
    ctx.fresh.should.be.false;
  })
);
```

Adding an other test will need to re-use `{}`, and so updates lines that are not related to PR for example.

cc/ @juliangruber @tj @tejasmanohar

stojanovic:
Hm, they are not needed but maybe we should keep them, it's sometimes large blocks so it's more readable with `{}`.
I can close #549 since you'll cover all those cases in your refactor.

test/application/onerror.js
```diff
@@ -5,72 +5,62 @@ const stderr = require('test-console').stderr;
...
-describe('app.onerror(err)', function(){
-  it('should throw an error if a non-error is given', function(done){
+describe('app.onerror(err)', () => {
+  it('should throw an error if a non-error is given', done => {
     const app = new Koa();
 
-    (function(){
-      app.onerror('foo');
-    }).should.throw(AssertionError, {message: 'non-error thrown: foo'});
+    (() => app.onerror('foo')).should.throw(AssertionError, {message: 'non-error thrown: foo'});
```
broucz: Here > 80 characters
Should be better to format as: ?
```js
(() => app.onerror('foo'))
.should.throw(AssertionError, {message: 'non-error thrown: foo'});
```

test/response/status.js
```diff
@@ -7,59 +7,53 @@ const statuses = require('statuses');
...
   });
 
-  describe('when a status string', function(){
-    it('should throw', function(){
-      assert.throws(function(){
-        response().status = 'forbidden';
-      }, 'status code must be a number');
+  describe('when a status string', () => {
+    it('should throw', () => {
+      assert.throws(() => response().status = 'forbidden', 'status code must be a number');
```
broucz: Here > 80 characters
Should be better to keep: 
```js
it('should throw', () => {
  assert.throws(() => {
    response().status = 'forbidden';
  }, 'status code must be a number');
});
```

jonathanong:
needs a rebase. is there an eslint rule for this?

test/response/status.js
jonathanong: do we even have a 80 character limit? if it's not an eslint rule, it doesn't matter!

broucz: not so far, I kept one line this time as it seems easier to read.

broucz:
eslint rules were added later during #555 merge

stojanovic:
I think we can add this one too here:
```
prefer-arrow-callback: 2
```

broucz:
looks good, I'll use it also

broucz:
@stojanovic `prefer-arrow-callback: 2` is perfect !
I didn't refactor using 80 char limit as you partially (or will) cover it in #555 , but other lint rules are ok.

jonathanong:
ugh needs another rebase :(

@broucz make sure you're running `git rebase`, not `git merge`. we shouldn't see these "merge" commits

felixfbecker:
When I refactored all the generator tests I just removed all the *. After rebase those would have to be changed to arrow functions too.

jonathanong:
@felixfbecker oh nice!

felixfbecker:
See #588 for reference :)

felixfbecker:
And imo if the function is named it is not suited for an arrow function. Arrow functions are for callbacks.

ericmdantas:
Even though `done => 'foo'` is correct, I still like to have the parenthesis, as in `(done) => 'foo'`. 

But that could be just me.

tejasmanohar:
>And imo if the function is named it is not suited for an arrow function. Arrow functions are for callbacks.
:+1: - unless you're dealing with some of the cases where it's cleaner to arrow it in an obj literal

>Even though done => 'foo' is correct, I still like to have the parenthesis, as in (done) => 'foo'.
:-1:, see no reason to personally- the first is cleaner, reads more like an expression to me

kriswill:
When I first started using arrow functions, I put parens on them all.  But after using them for a while, I've started removing them for single argument arrow functions, since that style is easier to scan and type and feels more ergonomic.

felixfbecker:
+1 for [`"arrow-parens": [2, "never"]`](http://eslint.org/docs/rules/arrow-parens)

broucz:
Sorry for the delay guys... I was moving these last days and so kind of far from internet.

@jonathanong rebase is done and I hope fine this time. If you still have some issues with just let me know. I made some mistake on this branch, but if needed I can PR using a cleaner one.

Issue #552, opened 2015-10-25
Indentation for multiple props in new lines


stojanovic:
I think this
```
request(server)
  .get('/')
  .expect(404)
  .end(function(err){
    if (err) return done(err);
    calls.should.eql([1, 2, 3, 4, 5, 6]);
    done();
  });
```
is a bit more readable than this
```
request(server)
.get('/')
.expect(404)
.end(function(err){
  if (err) return done(err);
  calls.should.eql([1, 2, 3, 4, 5, 6]);
  done();
});
```

Again, it's just a readability, there's no real gain here.
Should we update this or leave it as it is?

Also, do we need to follow 80 chars limit? I saw some lines a bit longer, but it's an easy fix.

cc. @tejasmanohar

tj:
sounds fine to me, but I don't want every commit for Koa to be bikeshedding, maybe 1% of the recent activity is "real" progress

stojanovic:
Cool, I'll do it later today.

felixfbecker:
I think the opposite.

broucz:
More or less the same for me even if I'm more used to @stojanovic indent.
We should add a lint rule for this (both version can be find -at least- in test folder), also to avoid some [blank line](https://github.com/koajs/koa/blob/master/test/application/respond.js#L148) (it could be intentional here, better comprehension).

tejasmanohar:
+1, this indentation is more common too... I think Koa is one of the few places I've seen it differently

stojanovic:
Actually it was mixed in Koa.
[Some parts](https://github.com/koajs/koa/blob/master/lib/context.js#L166) of the current master already has that kind of indentation.

tj:
:+1: for linting, then we have nothing to bikeshed

stojanovic:
Done in #555.

PR #550, opened 2015-10-25
Update test: application -> use() should throw if not a function

broucz:
Missing for 100% coverage.

PR #555, opened 2015-10-25
Refactor indentation

stojanovic:
Refactor mentioned in #552:

- Adds indentation for props in new lines
- Preserves 80 chars limit

Only exception is [this line](https://github.com/koajs/koa/blob/master/test/application/respond.js#L597) because it'll be exactly 80 chars after #553 is merged to master.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -12,4 +12,3 @@ rules:
   arrow-parens: [2, "as-needed"]
   arrow-spacing: 2
   dot-location: [2, "property"]
-  max-len: [2, 80, 4, {"ignoreComments": true, "ignoreUrls": true}]
--- a/.eslintrc
+++ b/.eslintrc
@@ -11,3 +11,5 @@ rules:
   yoda: 0
   arrow-parens: [2, "as-needed"]
   arrow-spacing: 2
+  dot-location: [2, "property"]
+  max-len: [2, 80, 4, {"ignoreComments": true, "ignoreUrls": true}]
--- a/test/request/accepts.js
+++ b/test/request/accepts.js
@@ -8,8 +8,14 @@ describe('ctx.accepts(types)', function(){
     describe('when Accept is populated', function(){
       it('should return all accepted types', function(){
         const ctx = context();
-        ctx.req.headers.accept = 'application/*;q=0.2, image/jpeg;q=0.8, text/html, text/plain';
-        ctx.accepts().should.eql(['text/html', 'text/plain', 'image/jpeg', 'application/*']);
+        ctx.req.headers.accept =
+          'application/*;q=0.2, image/jpeg;q=0.8, text/html, text/plain';
+        ctx.accepts().should.eql([
+          'text/html',
+          'text/plain',
+          'image/jpeg',
+          'application/*'
+        ]);
       });
     });
   });
@@ -18,7 +24,8 @@ describe('ctx.accepts(types)', function(){
     describe('when Accept is populated', function(){
       it('should return false', function(){
         const ctx = context();
-        ctx.req.headers.accept = 'application/*;q=0.2, image/jpeg;q=0.8, text/html, text/plain';
+        ctx.req.headers.accept =
+          'application/*;q=0.2, image/jpeg;q=0.8, text/html, text/plain';
         ctx.accepts('image/png', 'image/tiff').should.be.false;
       });
     });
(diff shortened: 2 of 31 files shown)


broucz:
Not successful coverage will be fix with #550 merge.

stojanovic:
Perfect, I'll wait for that then :)

jonathanong:
similarly to #550, are there relevant eslint rules?

stojanovic:
@jonathanong rules added. I just don't know if the rule for indented properties exist at all.
But we now have rules for dot location and 80 chars line length.

.eslintrc
```diff
@@ -11,3 +11,6 @@ rules:
   yoda: 0
   arrow-parens: [2, "as-needed"]
   arrow-spacing: 2
+  dot-location: [2, "property"]
+  "max-len": [2, 80, 4, {"ignoreComments": true, "ignoreUrls": true, "ignorePattern": "^\\s*var\\s.+=\\s*require\\s*\\("}]
```
broucz: May you forget to remove parentheses around `max-len` no ?

stojanovic: @broucz ha, you are right :)
Fixed, I also simplified that rule a bit, so it has less than 80 chars long :)

broucz: if even the rule is < 80: nice ;)

jonathanong:
needs a rebase =/

do you also mind squashing your commits so you don't have any like "fix merge conflicts"? otherwise, i'll just squash them all into one

stojanovic:
Fixed and I squashed all into one since it makes sense to have just one commit here.
Next time I'll do this immediately.

test/request/accepts.js
```diff
@@ -8,8 +8,14 @@ describe('ctx.accepts(types)', function(){
     describe('when Accept is populated', function(){
       it('should return all accepted types', function(){
         const ctx = context();
-        ctx.req.headers.accept = 'application/*;q=0.2, image/jpeg;q=0.8, text/html, text/plain';
-        ctx.accepts().should.eql(['text/html', 'text/plain', 'image/jpeg', 'application/*']);
+        ctx.req.headers.accept =
```
jonathanong: not sure i like this... weird IMO lol

stojanovic: It is indeed weird, but I don't know hot to compress it in 80 chars better :D

tejasmanohar: @stojanovic you may have misunderstood my point earlier

`ctx.accepts('text/html', 'text/plain', 'image/jpeg', 'application/*').should.equal('text/html');` is fine all in 1 line. it's just that if it's getting _too long_... then you can do something like

```javascript
ctx
  .accepts('text/html', 'text/plain', 'image/jpeg', 'application/*')
  .should.equal('text/html')
```

indenting -> chaining one function but not the others is odd.. so yea, what it currently is seems weird to me too lol @jonathanong

jonathanong: anyone actually want to limit to 80 characters? i personally don't mind long lines as long as it's just a really long constant/literal

stojanovic: I can remove that limit and rebase if that works better, it'll definitely be a bit less weird.
@tejasmanohar ok, that's easy fix too.

felixfbecker:
The refactor on indentation is great, but the forced 80 column limit actually makes stuff _harder_ to read. We all use 16:9 monitors and our editors support soft wrap. 80 columns was a convention for 4:3 cathode ray tubes running `vi`.

tj:
haha agreed, same argument as tabs being better, your editor can display whatever it likes

tejasmanohar:
if it gets way way too long or a big chain of functions, then i'd prefer the
```javascript
obj
  .x(...)
  .y(...)
  .z(...)
  .a(...)
```
format just for readability, but i don't care about arbitrary hard limits like 80 chars. just word wrap in your editor if you'd like :)

test/request/accepts.js
tejasmanohar: yep, i don't care about hard limits on chars / line

felixfbecker:
Agree with @tejasmanohar!

stojanovic:
It should be a bit better now, I removed 80 chars limit and weird line breaks

jonathanong:
@stojanovic you gotta rebase master, not merge master!

PR #558, opened 2015-10-27
Remove co dependency

### LINKED ISSUE #557: Drop co dependency
This has already been discussed in #533 and #415 . Quoting @yanickrochon:
 > Since  co  has been Promise-based for a while now, I think Koa should drop it as dependency and move to a Promise solution. The advantage of using promises is quite evident and should not have a negative impact in most cases. Also, old middlewares can use  co.wrap  if they need to be re-used, until they migrate to be 2.0 compatible. To me, since there is a solution that not only works but is clean means that there's no problem in dropping  co  in favour of Promise or async/await (using a transpiler for now).
> 
 > Furthermore, Generators have always been "hacky". It does not promote good usage of the design pattern :)

This would also allow us to use Bluebird.coroutine instead of co if we want.
Would submit PR

felixfbecker:
Closes #557

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -21,7 +21,6 @@ const assert = require('assert');
 const Stream = require('stream');
 const http = require('http');
 const only = require('only');
-const co = require('co');
 
 /**
  * Expose `Application` class.
@@ -93,7 +92,9 @@ module.exports = class Application extends Emitter {
   use(fn) {
     debug('use %s', fn._name || fn.name || '-');
     if (typeof fn !== 'function') throw new TypeError('middleware must be a function!');
-    if (isGeneratorFunction(fn)) fn = co.wrap(fn);
+    if (isGeneratorFunction(fn)) {
+      throw new Error('Support for generators has been removed. Use Promises or wrap your generator with co.wrap');
+    }
     this.middleware.push(fn);
     return this;
   }
--- a/test/application/use.js
+++ b/test/application/use.js
@@ -62,4 +62,10 @@ describe('app.use(fn)', function(){
     (() => app.use('not a function')).should.throw('middleware must be a function!');
     done();
   });
+
+  it('should throw error for generator', function(){
+    const app = new Koa();
+
+    (() => app.use(function *(){})).should.throw('Support for generators has been removed. Use Promises or wrap your generator with co.wrap');
+  });
 });
--- a/test/application/use.js
+++ b/test/application/use.js
@@ -11,21 +11,21 @@ describe('app.use(fn)', function(){
 
     app.use(function(ctx, next){
       calls.push(1);
-      next().then(function(){
+      return next().then(function(){
         calls.push(6);
       });
     });
 
     app.use(function(ctx, next){
       calls.push(2);
-      next().then(function(){
+      return next().then(function(){
         calls.push(5);
       });
     });
 
     app.use(function(ctx, next){
       calls.push(3);
-      next().then(function(){
+      return next().then(function(){
         calls.push(4);
       });
     });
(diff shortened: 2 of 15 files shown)


felixfbecker:
Unfortunately it is hard for me to test because this module relies on makefiles and `make` isn't available on Windows. It would be nice if the tests just worked without `make`.

stojanovic:
@felixfbecker I can send PR tomorrow to add everything from Makefile to npm scripts, so you'll be able to run tests and lint without `make`.

felixfbecker:
Did it by running mocha and eslint manually.

stojanovic:
We just need that in npm scripts :)

yanickrochon:
Before this PR is merged, all documentations should be updated to reflect the change :)

jonathanong:
anyone object to this?

tejasmanohar:
:+1: from me. we should put up a migrating from koa 1.x.x -> 2.x.x / major changes doc, right? @jonathanong - If so, I can PR it later this week

tj:
:+1:

test/application/use.js
```diff
@@ -9,22 +9,25 @@ describe('app.use(fn)', function(){
     const app = new Koa();
     const calls = [];
 
-    app.use(function *(ctx, next){
+    app.use(function(ctx, next){
       calls.push(1);
-      yield next();
-      calls.push(6);
+      next().then(function(){
```
gyson: `return next()`

test/application/use.js
```diff
@@ -9,22 +9,25 @@ describe('app.use(fn)', function(){
...
+        calls.push(6);
+      });
     });
 
-    app.use(function *(ctx, next){
+    app.use(function(ctx, next){
       calls.push(2);
-      yield next();
-      calls.push(5);
+      next().then(function(){
```
gyson: `return next()`

test/application/use.js
```diff
@@ -9,22 +9,25 @@ describe('app.use(fn)', function(){
...
+        calls.push(5);
+      });
     });
 
-    app.use(function *(ctx, next){
+    app.use(function(ctx, next){
       calls.push(3);
-      yield next();
-      calls.push(4);
+      next().then(function(){
```
gyson: `return next()`

felixfbecker:
@gyson good find. Kind of confusing though that all tests pass even though I didn't do this?!

gyson:
@felixfbecker because this test case is very simple and it just accidentally passed even w/o `return`.

Tests would fail if we test `.catch` or delay some `next()` call.

For example

```js
const Koa = require('koa')

const app = new Koa();
const calls = [];

app.use(function(ctx, next){
  calls.push(1);
  return next().catch(function(){
    calls.push(4); // w/o `return`, this wont be called
  });
});

app.use(function(ctx, next){
  calls.push(2);
  return next().then(function(){
    calls.push(3);
    throw new Error()
  });
});
```

lib/application.js
```diff
@@ -93,7 +92,9 @@ module.exports = class Application extends Emitter {
   use(fn) {
     debug('use %s', fn._name || fn.name || '-');
     if (typeof fn !== 'function') throw new TypeError('middleware must be a function!');
-    if (isGeneratorFunction(fn)) fn = co.wrap(fn);
+    if (isGeneratorFunction(fn)) {
+      throw new Error('Support for generators has been removed. Use Promises or wrap your generator with co.wrap');
```
jonathanong: should be a `TypeError` :)

jonathanong: i'll fix, squash, and merge

felixfbecker: You're right of course.

Issue #559, opened 2015-10-27
Remove Makefile, put everything in npm scripts


felixfbecker:
`make` is not available on Windows. NPM scripts however can run anywhere. Makefiles are also unnecessarily complicated for something simple as running `mocha` or `eslint` with a few options. Also, in npm scripts, you don't have to specify the path of the binary.
@stojanovic looking forward to your PR

targos:
I'm working on it ;)

stojanovic:
Me too :)
Go for it, since I just stared :)

targos:
https://github.com/koajs/koa/pull/560

felixfbecker:
I don't get why `make` is faster... Aren't they both just ways of calling js modules? Just that for the one method, I need `build-essential` on my machine.
Btw as far as I know in Windows you set env variables with `set`

jonathanong:
how many of koajs committers/contributors use windows?

PlasmaPower:
Also, remember that cygwin/mingw, while somewhat of a pain, are still options, so it's not totally unworkable for Windows users.

jonathanong:
who's for and against this? my personal opinion: we should use npm scripts for wider compat, but not at the expense of having an unmanageable `package.scripts` object.

alsotang:
I support Makefile because npm script is a fool choose for project. It only
support raw string, only.

2016-05-14 23:40 GMT+08:00 jongleberry <notifications@github.com>:

fl0w:
I'm with jonathan, npm scripts unless they become unbearable.

detailyang:
Makefile is enough and  npm scripts is not bad. But for universality, maybe npm scripts is more better:)

jeffijoe:
I vote npm scripts as well - they can get hairy, sure, but thats when you start breaking them out and chaining them instead.

PlasmaPower:
@jeffijoe Have a good way to do `mocha test/**/*.js` that's compatible

zanaca:
I have compared a machine running Node that was started with NPM  and  one that was started in pure node command. The one started by ```npm start``` was using more memory and cpu just because NPM was "processing"  the Node process itself. Since that day my Makefile calls node directly.

So, maybe, Makefile should be preserved and you guys should run a benchmark.

tj:
That's one reason I don't use npm scripts, the extra ~300ms bugs me, and it adds extra output. Not a huge deal, but if Windows was less of a mess it wouldn't be a big deal :D

[...]

PR #560, opened 2015-10-27
Remove Makefile, put everything in npm scripts

targos:
Fixes: #559


targos:
I just don't know what to do with the benchmarks

package.json
```diff
@@ -4,7 +4,10 @@
   "description": "Koa web app framework",
   "main": "lib/application.js",
   "scripts": {
-    "test": "make test"
+    "lint": "eslint lib test",
+    "test": "NODE_ENV=test mocha --require should --require should-http test/application/* test/context/* test/request/* test/response/*",
+    "test-cov": "NODE_ENV=test istanbul cover ./node_modules/.bin/_mocha -- -u exports --require should --require should-http test/application/* test/context/* test/request/* test/response/* --bail",
```
stojanovic: `_mocha` is already in the path in scripts, no need for `.node_modules/.bin/_mocha`, this should work fine:
```
"test-cov": "NODE_ENV=test istanbul cover _mocha -- -u exports --require should --require should-http test/application/* test/context/* test/request/* test/response/* --bail",
```

juliangruber:
-1, make is a lot faster than pm

package.json
targos: you're right, I changed it.

targos:
Timings on my side:
```
make test > /dev/null  0.65s user 0.07s system 106% cpu 0.675 total
npm test > /dev/null  1.12s user 0.13s system 100% cpu 1.240 total
```
I personally don't care, 1-2 seconds to run tests is fast enough.
Seems that it doesn't change anything on travis (always around 50s)

stojanovic:
@juliangruber you are right, it is faster, also we still have Makefile for benchmarks.
But people on windows needs to run tests and coverage manually. Maybe we can keep both, and use Make on travis, but who ever wants can use one or another.
It's a bit redundant of course.

felixfbecker:
This is what I get when I run with make in an Ubuntu VM:
```
/media/sf_git/koa/node_modules/.bin/_mocha:4
case `uname` in
^^^^

SyntaxError: Unexpected token case
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:414:25)
    at Object.Module._extensions..js (module.js:442:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:311:12)
    at Function.Module.runMain (module.js:467:10)
    at startup (node.js:134:18)
    at node.js:961:3
```
But why do I have to go into an Ubuntu VM just so I can run a make file that then calls js files that could be easily put into npm scripts?  
I don't quite get why `make` should be significantly faster - in the end those two are only two ways of launching js files.

Btw in Windows I think you can only set env variables with `set` command.

felixfbecker:
`env NODE_ENV=test mocha` should work cross-platform if we really need the NODE_ENV, assuming you have git installed and in path on Windows.

felixfbecker:
But imo using stuff like NODE_ENV is bad practice anyways (see http://12factor.net/config). I looked at the test code and actually never uses it, so why does need to be there?
The NODE_ENV-dependent error handling is annotated as deprecated.
There is one test in `test/application/toJSON.js` that assumes the env variable of `test`, when it really should do it like `test/application/index.js:50` and set the env beforehand and reset after.

felixfbecker:
For benchmarks just keep the Makefile in the benchmarks subdir. Oh and btw instead of 
`mocha --require should --require should-http test/application/* test/context/* test/request/* test/response/*`
we can just write
`mocha --require should --require should-http --recursive` (mocha looks in test directory by default)
or
`mocha --require should --require should-http test/**/*.js`
But now enough spam for this thread :D

yanickrochon:
I agree with @felixfbecker about `NODE_ENV` being misused. We can change `process.env` when testing (self contained tests) and relying on external settings is bad maintenance practice.

jonathanong:
don't have an opinion here...

but if we move it to npm, we should move the config to an `mocha.opts` file or something so the commands aren't 2000 characters long

felixfbecker:
Like I said, you don't need to specify the paths in mocha. It could be as short as
```sh
mocha -r should -r should-http --recursive
```
will look inside test folder and all subfolders.
Also, I think we agreed on removing NODE_ENV setting (would require one change in one of the test as mentioned).

targos:
@felixfbecker I updated according to your comments. PTAL

felixfbecker:
looks great :+1:

package.json
```diff
@@ -4,7 +4,10 @@
   "description": "Koa web app framework",
   "main": "lib/application.js",
   "scripts": {
-    "test": "make test"
```
jonathanong: we need a `npm run bench` task too then

jonathanong:
@koajs/owners thoughts?

tj:
I find it really weird that running scripts is in npm's domain, but no major complaints here

yanickrochon:
@tj I presume this is because npm *can* execute pre-install and post-install scripts... so the feature is already there. At some point, I guess it just became standard to run tests and stuff with npm. Personally, I think it's somewhat within it's domain, as npm provides tools (i.e. running scripts) to ensure some package quality. Anyhow, it's much easier to write script instructions in `package.json` than creating a Makefile... at least to people who never wrote such files before :)

travisjeffery:
-1 here too, make is faster and more flexible and is currently the one place you need to look to find the testing/benchmarking stuff which makes it more approachable.

felixfbecker:
I don't think it is more approachable. Every Node developer knows npm scripts. They look at a package.json and immidietly know what it does. They expect to do `npm test` to run the tests after they forked because that works across all modules from npm. A makefile is a complicated piece of file that only works on specific platforms and does nothing what npm scripts cannot do. If you need more complicated build scripts, write a js file and call it in your npm script.

juliangruber:
The package.json is still using the bench script, can someone add that?

tj:
It's just another thing that npm does poorly, but seeing as windows people seemed to have never come up with a reasonable Make implementation (no reason they couldn't) I'm cool with it

yanickrochon:
Using `make` shaves a mere second or two when running tests over `npm`. As for benchmark, they are totally relative and being faster with a Makefile means nothing at all since they cannot be compared between computers anyhow. So, really, it all comes down to a preference.

That being said, as @felixfbecker said, every Node programmer expect packages to provide `npm test` to run the test suites. And since either solutions are standards and really no big deal in terms of maintenance, why don't we have both of them? Even as a Linux user, I don't personally use `make` (force of habit, I guess) and Windows support is inevitable. I cannot conceive writing apps in Windows that will be running on Linux... but I presume some people cannot live without a full blown IDE like VS, but having an implemented standard (i.e. `npm test`) be available is irrelevant to the OS. It feels like the old war between using spaces or tabs... except that, now, we can have both :)

dead-horse:

felixfbecker:
Why? This is just dev stuff.

tejasmanohar:
@targos @felixfbecker I think it's fine to send a PR for this to `master` as long as it doesn't bring all the `v2.x` commits over (since the branches switched around). If that's merged, should probably send a PR to `v2.x` as well.

targos:
I'm fine if someone else wants to take care of this. I am on holiday for the next two weeks.

jonathanong:
@targos i'm going to close this because i don't think we've gotten a consensus and the PR is out of date. sorry :( thanks for PR though

markelog:
Why an [issue](https://github.com/koajs/koa/issues/559) for this is still open?

PlasmaPower:
@markelog because it's still a pain for windows users to run tests

markelog:
So this was closed only because pr had some conflicts?

PlasmaPower:
@markelog It was left to rot because of a number of reasons, mainly that it resulted in slower performance and that there aren't any Koa contributers on Windows that we know of.

felixfbecker:
@PlasmaPower It's a chicken-or-egg problem ;)

markelog:
@PlasmaPower your comments contradict to each other :/, if pr for this is not acceptable, issue should be closed

PlasmaPower:
@markelog There might be another solution, IDK I haven't looked into it. Maybe we could not use the Makefile only for Windows users?

@felixfbecker Yep, somewhat, but you can run the tests with cygwin/mingw anyway.

jeffijoe:
@PlasmaPower we probably won't have any Windows code contribs if we can't support it. ;)

PlasmaPower:
@jeffijoe As I've commented above, most Windows devs are running Cygwin/Mysys anyway and therefore have access to Make. However, if we have enough of them it would be helpful to switch to native NPM scripts.

jeffijoe:
I'm running MINGW which ships with Git and I don't have make.

PlasmaPower:
@jeffijoe Install MYSYS.

PR #562, opened 2015-10-28
babel-node required -> Babel

tejasmanohar:
was looking over the docs and stumbled across this. there are a few different ways to use babel in node apps like precompile, babel-node to compile and execute on the fly, require hooks, and whatnot so i'd prefer to be less specific


felixfbecker:
:+1:

Issue #564, opened 2015-10-28
Broken middleware since 2.0.0


rbaprado:
Hello, I just upgraded to Koa 2.0.0 and due to the generator support dropping, all my middleware simply broke (i.e. koa-session). I tried co.wrap them but got other errors (I will provide more information shortly). Could someone please present an example of using middleware not yet ported to 2.0.0?

Original code:
`app.use((session(app));`

My attempt:
`app.use(co.wrap(session(app)));`

madbence:
Try [`koa-convert`](https://www.npmjs.com/package/koa-convert), it can transform a generator-based middleware into a promise based one.

rbaprado:
So it would become:
`app.use(convert(session(app))`?

menems:
yep

rbaprado:
That did the job. Could someone explain why that is different than co.wrap?

menems:
the difference is the signature, in new Koa, the context is pass as argument not in legacy

rbaprado:
Don't you think the error message koa throws recommending to use co.wrap a bit misleading then?

yanickrochon:
Yes. `co.wrap` is only to use a gnerator function with Koa 2.0. Old middlewares need to be converted, until they are updated to follow the current next major release's middleware signature.

madbence:
in `koa@2` the context is an arg, so you write your middlewares like `async (ctx, next) => ...`, `co.wrap` just wraps your `function* (next) {...}` middleware into an `async (next) => ...`, so it won't work.

you *can* use generators, if the have proper signature (`function* (ctx, next) { ... }`)

rbaprado:
Ok, I got it now. Thank you all.

PR #565, opened 2015-10-28
Add usage of koa-convert for legacy middleware

menems:
#564, #538 

I change the old version by the used of koa-convert,

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -92,7 +92,7 @@ module.exports = class Application extends Emitter {
   use(fn) {
     debug('use %s', fn._name || fn.name || '-');
     if (typeof fn !== 'function') throw new TypeError('middleware must be a function!');
-    if (isGeneratorFunction(fn)) throw new TypeError('Support for generators has been removed. Use Promises or wrap your generator with co.wrap');
+    if (isGeneratorFunction(fn)) throw new TypeError('Support for generators has been removed. See the documentation for examples of how to convert old middleware https://github.com/koajs/koa#example-with-old-signature');
     this.middleware.push(fn);
     return this;
   }
--- a/test/application/use.js
+++ b/test/application/use.js
@@ -66,6 +66,6 @@ describe('app.use(fn)', function(){
   it('should throw error for generator', function(){
     const app = new Koa();
 
-    (() => app.use(function *(){})).should.throw('Support for generators has been removed. Use Promises or wrap your generator with co.wrap');
+    (() => app.use(function *(){})).should.throw(/.+/);
   });
 });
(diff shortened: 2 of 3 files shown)


yanickrochon:
Perhaps modify the error message thrown for the dropped generator function support and specify that old middleware should use koa-convert, so issues like #564 are better handled in the future (because we won't see the last of this otherwise).

menems:
I'm agree, but how to test if the function have a old signature? check if the first arg is a context object?

felixfbecker:
You cannot check that but you could give a hint in the error message that the user should use koa-convert if he is using a legacy third-party middleware.

menems:
"Support for generators has been removed. Use Promises or wrap your generator with co.wrap. legacy middleware should be convert using koa-convert" ??

felixfbecker:
Maybe change "Use Promises or wrap your generator with co.wrap" to "See the documentation for examples of how to convert old middleware."

menems:
agree, it's shorter

menems:
oups

lib/application.js
```diff
@@ -93,7 +93,7 @@ module.exports = class Application extends Emitter {
     debug('use %s', fn._name || fn.name || '-');
     if (typeof fn !== 'function') throw new TypeError('middleware must be a function!');
     if (isGeneratorFunction(fn)) {
-      throw new TypeError('Support for generators has been removed. Use Promises or wrap your generator with co.wrap');
+      throw new TypeError('Support for generators has been removed. See the documentation for examples of how to convert old middleware');
```
jonathanong: we should provide a link here

menems:
link added

jonathanong:
@menems oh no, we should be linking to the exact documentation for converting. not just the repo. :)

test/application/use.js
```diff
@@ -66,6 +66,6 @@ describe('app.use(fn)', function(){
   it('should throw error for generator', function(){
     const app = new Koa();
 
-    (() => app.use(function *(){})).should.throw('Support for generators has been removed. Use Promises or wrap your generator with co.wrap');
+    (() => app.use(function *(){})).should.throw('Support for generators has been removed. See the documentation for examples of how to convert old middleware https://github.com/koajs/koa');
```
jonathanong: for the test, you could also just test that a substring exists so you don't have to make these two strings EXACTLY the same

menems:
linking on the readme anchor?

jonathanong:
yeah

menems:
not complet , next time i will create a branch :)

menems:
now I think is good

PR #578, opened 2015-10-31
fix Application.inspect() – missing .proxy value (v1.x port)

tj:


tj:
oops, opened against master

Issue #581, opened 2015-11-01
Best practice for async with Koa 2.0


derek-duncan:
First off, Koa 2.0 –> great work! I've been playing with it today and have really appreciated the focus on modularity and the future of js.

In light of this new release, what's the most preferred/best practice for handling *asynchronous operations*? In previous iterations of Koa, generators have been exclusively used to provide handle async and avoid callback hell. With this new release, generator support has been completely dropped in preference of async functions, which rely heavily on Promises.

Anyways, I'm often posed with the challenge of choosing one of the above options to handle asynchronous operations. However, I want to generally keep my choice consistent across a codebase since generators and async functions seem to not work fluently together. Also, as a personal preference, I'd like to avoid wrapping so many functions/modules in a helper function (like `koa-convert`) in lieu of forcing these various options to work together. Thus, I just wanted to hear the Koa community's thoughts on the matter before I go dive into an unconventional direction.

felixfbecker:
- If you don't mind transpiling, use async/await with Babel
 - If you want the beauty of async/await without transpiling, use co.wrap or Bluebird's coroutine with yield instead of await (this is what I do, but if you have coworkers it's a bit harder to explain).
 - If you don't want to use generators or transpile, just return a promise in your middleware and chain it for your async operations.

For modules that don't return promises, promisify with Bluebird - it means you don't depend on wrapper modules that might not be always up to date.

koa-convert is mainly to convert from the old _signature_ of `this` as context to `(ctx, next) => ...`.

I hope this helped :)

tejasmanohar:
:+1: - @felixfbecker's comment (more or less) should be in one of our docs

felixfbecker:
Maybe we should simply add some async database operations to the existing examples in the docs for asyncawait/promise/generator so it's clearer how to combine them?

felixfbecker:
Oh and of course you can construct your own promise if you're working with a very EventEmitter-heavy library which are tough to promisify (ldapjs as an example):
```js
app.use((ctx, next) => new Promise((resolve, reject) => {
  ldap.search('dn=users', (err, result) => {
    if (err) return reject(err)
    result.on('searchEntry', entry => {
      ctx.body = entry.object
      resolve()
    })
    result.on('error', reject)
  })
})

yanickrochon:
[...]

PR #582, opened 2015-11-02
Add DB examples to Readme

felixfbecker:
Adds some async operations to the examples for #581.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -59,7 +59,9 @@ app.use((ctx, next) => {
 // response
 
 app.use(ctx => {
-  ctx.body = 'Hello World';
+  return someDbOperation().then(result => {
+    ctx.body = 'Hello World, here is your result: ' + result;    
+  })
 });
 
 app.listen(3000);
@@ -82,8 +84,9 @@ app.use(async (ctx, next) => {
 
 // response
 
-app.use(ctx => {
-  ctx.body = 'Hello World';
+app.use(async ctx => {
+  let result = await someDbOperation();
+  ctx.body = 'Hello World, here is your result: ' + result;
 });
 
 app.listen(3000);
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -59,7 +59,9 @@ app.use((ctx, next) => {
 // response
 
 app.use(ctx => {
-  ctx.body = 'Hello World';
+  return someDbOperation().then(result => {
```
stojanovic: You don't need `{}` and `return`.

Readme.md
```diff
@@ -59,7 +59,9 @@ app.use((ctx, next) => {
...
+    ctx.body = 'Hello World, here is your result: ' + result;
```
stojanovic: Use template strings here:
```js
ctx.body = `Hello World, here is your result: ${result}`;
```

tejasmanohar:
Personally, -1 / :-1:. would leave this out of the README since it's getting application-specific. That said, if we do include it, I recommend we change `someDbOperation` to `someAsyncOperation` (or `someAsyncOp`) to make it as generic as possible and avoid questions about what DB operation? :P

stojanovic:
:+1:

Readme.md
tejasmanohar:
```javascript
res => ctx.body = `Hello World, here is your result: ${result}`
```

stojanovic: Yes, I really like arrow functions because of that :)

juliangruber:
-1 from me as well, it's out of scope for koa

yanickrochon:
Why not putting all these examples and extra documentation in the wiki? Keep the README light and simple.

MarkHerhold:
:-1: for the README.
:+1: for putting it in the Wiki under a common use-cases section.

Maybe include a link to the wiki page in the README?

felixfbecker:
I agree with the comments here.

ericmdantas:
Pretty much what @MarkHerhold said

dead-horse:
please put this example to https://github.com/koajs/examples :)

PR #583, opened 2015-11-02
Update Readme.md

scmart:
Babel will not currently transpile an async arrow function into a generator function, it will leave it as a regular arrow function, and thus not work.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -73,7 +73,7 @@ const app = new Koa();
 
 // logger
 
-app.use(async (ctx, next) => {
+app.use(async function(ctx, next) {
   const start = new Date;
   await next();
   const ms = new Date - start;


felixfbecker:
But... why? It could easily be transpiled into
```js
function *(ctx, next) {
 ...
}.bind(this)

juliangruber:
upgrading your node should fix this

scmart:
> But... why? It could easily be transpiled into
> 
> function *(ctx, next) {
>  ...
> }.bind(this)

I agree, but it doesn't, and I don't think having an example that doesn't work with the node 4.2, and Babel 6 is good.

menems:
+1,  I tried babel 6 last night with transform-async-generators plugins, and asyncArrowFunction didn't work. And I don't think when async will be integrated in v8 asyncArrowFunction will too.
#585

targos:
Someone should file a bug report about it in babel

jonathanong:
oh i see

jonathanong:
going to see if babel is going to fix this before merging!

jonathanong:
they fixed this :)

PR #586, opened 2015-11-04
ensure parseurl always working as expected

dead-horse:
`url.path = null` is introduced in https://github.com/koajs/koa/commit/d108926f4634fabe4b18c49ab0ab25a169dbd665

The code rely on `parseurl` will unpredictable when people write some code like `this.path = this.path`.

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -127,6 +127,8 @@ module.exports = {
 
   set path(path) {
     const url = parse(this.req);
+    if (url.pathname === path) return;
+
     url.pathname = path;
     url.path = null;
 
@@ -178,6 +180,8 @@ module.exports = {
 
   set querystring(str) {
     const url = parse(this.req);
+    if (url.search === `?${str}`) return;
+
     url.search = str;
     url.path = null;
 
--- a/test/request/path.js
+++ b/test/request/path.js
@@ -2,6 +2,7 @@
 'use strict';
 
 const context = require('../helpers/context');
+const parseurl = require('parseurl');
 
 describe('ctx.path', () => {
   it('should return the pathname', () => {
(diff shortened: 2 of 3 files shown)


lib/request.js
```diff
@@ -128,7 +128,7 @@ module.exports = {
   set path(path) {
     const url = parse(this.req);
     url.pathname = path;
-    url.path = null;
+    url.path = url.pathname + url.search;
```
dead-horse: maybe we should add `if (url.pathname === path) return` to avoid `stringify(url)`

tejasmanohar: yes, i think so, too. @dead-horse

dougwilson:
If tgis is a bug, can we nit add a test case?

dead-horse:
@dougwilson added

test/request/path.js
```diff
@@ -28,4 +29,11 @@ describe('ctx.path=', function(){
     ctx.originalUrl.should.equal('/login');
     ctx.request.originalUrl.should.equal('/login');
   });
+
+  it('should not affect parseurl', function(){
```
stojanovic: Use arrow functions in callbacks, otherwise lint is throwing an error now.

dead-horse:
ping @dougwilson @jonathanong

dougwilson:
Makes sense to me.

jonathanong:
LGTM. should be added to `v1.x` too!

fengmk2:
LGTM

dead-horse:
ported to v1.x https://github.com/koajs/koa/commit/c243baa4d1e795a1c2a7767711185469f4e0ccb9 and released `2.0.0-alpha.3` & `1.1.2`.

PR #590, opened 2015-11-06
Added more use case when middleware is not a function

yanickrochon:
I noticed, this morning that if `app.use()` is called, or `null` or `undefined` is passed, then `debug` will throw and will not return a descriptive error to the user; it should throw `"middleware must be a function!"` whatever how `app.use` is called.


lib/application.js
```diff
@@ -99,7 +99,7 @@ module.exports = class Application extends Emitter {
    */
 
   use(fn) {
-    debug('use %s', fn._name || fn.name || '-');
+    debug('use %s', fn && (fn._name || fn.name) || '-');
```
dead-horse: just move this line below `if (isGeneratorFunction(fn)) ...` ?

jonathanong: yeah i prefer this solution

yanickrochon: I had an initial `if (fn === null || fn === undefined)`, but decided that it wasn't necessary... I also debated moving `debug` below the `if` (for type and generator check), but figured that having a debug output before throwing could be useful :)

dead-horse:

yanickrochon:
Holy crap! This wouldn't be needed had it been done before :P

PR #592, opened 2015-11-07
update readme

nswbmw:
just change

```
console.log(`${ctx.method} ${ctx.url} - ${ms}`)
```

to

```
console.log(`${ctx.method} ${ctx.url} - ${ms}ms`)
```


dead-horse:
rebase and update your commit message

PR #597, opened 2015-11-18
Use ES6 classes

simplyianm:
This converts `Context`, `Request`, and `Response` to ES6 classes to make things more semantic.


tejasmanohar:
-0 for classes. Doesn't make a difference nor change the public API. Don't agree that it's any more semantic either.
+1 for semicolons for consistency 

PS: Same feeling about #598

yanickrochon:
I disagree with @tejasmanohar. While it does not change the public API, this seems be "cleaner" than creating instances from plain objects. Also, this is now the preferred way to create instances. And the [application.js](https://github.com/koajs/koa/blob/master/lib/application.js) file *is* implemented using ES6 classes, so this does make for consistency in the project.

simplyianm:
@tejasmanohar -- Application is inconsistent as it uses a class whilst everything else doesn't. I think classes make things a lot more straightforward, as we're basically using the prototypes as classes anyway.

juliangruber:
+1

tejasmanohar:
Hmm, as per Application, I don't really like it as a class anymore since it's a Singleton that you'd most likely only `new` once... but still +/-0, don't really mind.

One question- Why is [this](https://github.com/simplyianm/koa/blob/4e2d084ff06b71e84b98c74b4140e2c87fcbe234/lib/application.js#L137-L139) using `Object#create`?

simplyianm:
@tejasmanohar I'll merge all my changes into one PR to make things simpler. Closing https://github.com/koajs/koa/pull/598

tejasmanohar:
Looks like Travis bugged out [here](https://travis-ci.org/koajs/koa/jobs/91965797). Maybe someone with access can re-run that build?

That said, I can attest that the tests pass locally (kinda beats the point of CI but whatever)

lib/context.js
```diff
@@ -50,22 +74,24 @@ const proto = module.exports = {
...
-   * @param {Mixed} test
+   * @param {Mixed} value
    * @param {Number} status
    * @param {String} message
+   * @param {Mixed} opts
    * @api public
    */
-
-  assert: httpAssert,
+  assert(value, status, msg, opts) {
```
alsotang: use `...args` is more es6-style

tejasmanohar: +1

simplyianm: I think this is only supported in Node 5, according to the REPL.

simplyianm: We could in theory use `assert() { httpAssert.apply(this, arguments); }` but I think the current is more clean. Thoughts?

dead-horse: `...args` in node@4 need `--harmony` flag. :(

tejasmanohar: ooh true, forgot destructuring wasn't stable in node- too used to babel haha

tejasmanohar: on that note, +1 for what's currently there @simplyianm

yanickrochon: Why the attempt to save bytes? Using `.apply` or any other method takes away argument clarity and is, in fact, slower anyway.

tejasmanohar: no point, current way is super readable.

jonathanong: problem with this version is that you can no longer do:

```js
this.assert.equal(1234, 1234, 'Were not equal to 1234')
```

this wasn't documented though... and i never actually used it. it would be a breaking change though. https://github.com/jshttp/http-assert/blob/master/index.js

jonathanong: also, the arguments to `http-errors` via `http-assert` are not strictly `status, msg, opts`, so that might be a little confusing

jonathanong: ^ PR for docs for that would be greatly appreciated

jonathanong:
other than the `http-assert` stuff, i'm +0. anyone else?

lib/context.js
simplyianm: Alternatively you could do `Context.prototype = httpAssert`, but I thought
that would be ugly/hard to follow. What do you think?

jonathanong: o i c... maybe we can just do in the constructor `this.assert = httpAssert`

jonathanong: i don't know ES6 classes well enough. i thought you could do something like `assert = httpAssert` but i can't find that

simplyianm: Yeah, I'll add it in the constructor

yanickrochon:
Yes, I like `this.assert = httpAssert;` better. +1 for me

tejasmanohar:
+/-0 for me as well

dead-horse:

simplyianm:
@dead-horse  -- https://github.com/koajs/koa/pull/605

PR #598, opened 2015-11-18
Make Context an ES6 class

simplyianm:
Semantic

// cc @tejasmanohar @jonathanong

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -14,7 +14,7 @@ const statuses = require('statuses');
  * Context prototype.
  */
 
-const proto = module.exports = {
+const Context = module.exports = class Context {
 
   /**
    * util.inspect() implementation, which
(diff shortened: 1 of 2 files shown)


juliangruber:
+1

lib/context.js
```diff
@@ -50,22 +50,24 @@ const proto = module.exports = {
...
-   * @param {Mixed} test
+   * @param {Mixed} value
    * @param {Number} status
    * @param {String} message
+   * @param {Mixed} opts
    * @api public
    */
-
-  assert: httpAssert,
+  assert(value, status, msg, opts) {
```
alsotang: why not use `...args`?

simplyianm: More readable to anyone viewing the code. However, I see that argument as well (no pun intended)

Use https://github.com/koajs/koa/pull/597 if you want to discuss this more.

PR #605, opened 2015-11-24
Use ES6 classes - v2.x

macalinao:
This converts `Context`, `Request`, and `Response` to ES6 classes to make things more semantic.

See https://github.com/koajs/koa/pull/597 for the old PR.


lib/application.js
```diff
@@ -127,28 +131,8 @@ module.exports = class Application extends Emitter {
     };
   }
 
-  /**
-   * Initialize a new context.
-   *
-   * @api private
-   */
-
   createContext(req, res) {
```
lourd: Instead of keeping `createContext` around you could replace where it's used [in the http callback](https://github.com/simplyianm/koa/blob/33467bf49d00a145ba5172d99d3ed9718ff4e108/lib/application.js#L128) with the call to the constructor.

lib/application.js
```diff
@@ -42,9 +41,14 @@ module.exports = class Application extends Emitter {
...
-    this.response = Object.create(response);
+
+    this.Context = class AppContext extends Context {};
+    this.Request = class AppRequest extends Request {};
+    this.Response = class AppResponse extends Response {};
+
+    this.context = this.Context.prototype;
+    this.request = this.Request.prototype;
+    this.response = this.Response.prototype;
   }
```
lourd: What's your thinking behind subclassing `Context,` `Request`, and `Response` here? And then saving the original class reference anyway? Why not `this.Context = Context` etc, and drop lines 49-51?

simplyianm: To preserve original behavior.

tj: most of it looks good to me, but I do think we should either ditch the prototypes or ditch the classes there, both is a little confusing. In general the use-case was to extend those for your app as you need, so the protos make more sense here to me at least

simplyianm: i think the prototypes should be ditched. doesn't really make too much sense to extend these per app. (usually there is only one app)

lib/application.js
simplyianm: true. again, did it for reverse compatibility, but it doesn't make sense to keep it, you're right

lib/application.js
tj: I suppose with the new sugar this would be the recommended way to go anyway:

```js
app.Context = class Context extends app.Context {
  whatever(){}
}
```

looks a little weird but hey haha. sucks you can't have `class extends app.Context`.

jonathanong: >  I do think we should either ditch the prototypes or ditch the classes there, both is a little confusing.

:+1: using classes and prototypes in the same place is weird to me. should be one or the other

jonathanong:
@simplyianm the code as-is won't be merge. do you want to update it based on our comments?

- Ditch all the prototypes
- Make extending the context/request/response objects use class inheritance (which is mostly removing features and adding docs):

```js
app.Context = class Context extends app.Context {
  whatever(){}
}
```

Let us know! thanks!

jonathanong:
FWIW i'm -1/2 on this PR because I don't really see any benefits and making it classes just makes it more confusing to me

simplyianm:
I'll update it

lib/application.js
tejasmanohar: :+1: same, mixing is messy

popomore:
+1

it's easy to extend Application/Context/Request/Response rather than override

tejasmanohar:
ping @simplyianm

simplyianm:
Fixed @tejasmanohar @jonathanong

tejasmanohar:
nice! probably best to squash commits so commit history isn't polluted :P

tejasmanohar:
ping @jonathanong

lib/request.js
```diff
@@ -17,7 +17,7 @@ const only = require('only');
  * Prototype.
  */
```
felixfbecker: The comment `Prototype` should go since this is not a prototype object anymore, but a class (function) that _has_ a prototype.

felixfbecker:
This is great imo. It is in line with the new `new Koa()` syntax, it is easier to understand (especially the getter/setter syntax) and it may make it easier to test koa apps without using supertest by simply creating new contexts / requests / responses.

jonathanong:
- [ ] can we add docs? specifically, how would someone extend the context/request/response with their own methods?
- [ ] why are the tests removed? can we rework the tests?

other than that, I'm +0

simplyianm:
I actually disagree with extending the base class -- things like that
should go in the middlewares to make execution more predictable. What are
the benefits to allowing it?

jonathanong:
well for one, people are using it: https://github.com/koajs/koa/issues/652

middleware is actually an anti-pattern in koa unless it's something you really want to be executed on every request (like checking etags and setting the status as 304 if the body is fresh). everything should be a function that you can just `await` ex. `const body = await parse.json(ctx)`. 

middleware:

- injects globals into your context
- is order dependent, meaning switching middleware order might cause undesirable behaviors or bugs
- usually executes logic on every request, which may be inefficient, or executes logic based on some "filter", which makes it harder to debug

for example, instead of having a body parsing middleware that always sets `this.request.body`, just `const body = await parse.json(ctx)`. the benefit of this is that sometimes you can do it before authentication and sometimes after, depending on the requirements of the route. changing order like this is difficult with middleware. if done properly, you should be able to see "all" the logic of a route in a single function instead of going through each middleware in different files.

now, back to context. the two reasons for extending context is:

1. allow users to build frameworks on top of koa
2. allow adding functions like `parse.json()` to `context` so you don't have to `require()` a billion things in every file, as well as defining all the globals in one place (vs. multiple middleware).

of course, this is prone to user error, so we never really recommended it strongly, but it's a feature that shouldn't be removed.

jonathanong:
similar philosophy: https://github.com/zeithq/micro/issues/8

PlasmaPower:
I don't like this PR because it won't fix much given that delegate is still used (multiple inheritance isn't supported in ES6).

felixfbecker:
@PlasmaPower that is true, but I don't think it has something to do with multiple inheritance. delegate only defines getters that _delegate_ to another object. It could be replaced by writing the getters into the class manually.

PlasmaPower:
@felixfbecker could you explain what you mean by "writing the getters into the class manually"?

felixfbecker:
@PlasmaPower Well instead of doing this
```js
delegate(Context.prototype, 'response')
  .method('attachment')
  .method('redirect')
  .method('remove')
  .method('vary')
  .method('set')
  .method('append')
  .access('status')
  .access('message')
  .access('body')
  .access('length')
  .access('type')
  .access('lastModified')
  .access('etag')
  .getter('headerSent')
  .getter('writable');
```
you simply write the getters and methods into the class manually:
```js
class Context {
  
  attachment(filename) {
    this.response.attachment(filename);
  }
  
  get status() {
    return this.response.status;
  }

  set status(status) {
    this.response status = status;
  }
```
This is what delegate does. It is a lot more verbose to write it manually. I personally have never been a fan of this delegation because it is not clear for someone who reads the code wehter `body` delegates to request or response for example, so I always use `response` / `request` explicitly anyway. But using delegate here is not a sin, the class is still much cleaner (I think you confused delegate with a mixing function).

PlasmaPower:
@felixfbecker yeah but that's a pain compared to simply extending Response and Request.

felixfbecker:
@PlasmaPower Always favor composition over inheritance ;)

fundon:
Drops `delegates` module will get better performances.

jonathanong:
hi, I'm going to close this for now. I don't think we need to move every single thing to ES6 — there's not much benefit in switching to ES6 classes.

thanks!

PR #606, opened 2015-11-25
Fix issue when app.use() is called with empty value

yanickrochon:
PR supersedes #590

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -99,7 +99,7 @@ module.exports = class Application extends Emitter {
    */
 
   use(fn) {
-    debug('use %s', fn._name || fn.name || '-');
+    debug('use %s', fn && (fn._name || fn.name) || '-');
     if (typeof fn !== 'function') throw new TypeError('middleware must be a function!');
     if (isGeneratorFunction(fn)) throw new TypeError('Support for generators has been removed. See the documentation for examples of how to convert old middleware https://github.com/koajs/koa#example-with-old-signature');
     this.middleware.push(fn);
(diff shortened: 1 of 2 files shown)


dead-horse:
d134fff9e82d050e81bf579a693ed2114e7946c6 landed

yanickrochon:
The point of keeping `debug` before the `if` is to know what is being passed. At the moment, an error is thrown, however there is no debug information telling what is being passed. My PR did solve that without losing any information.

dead-horse:
it will just debug `use -` when passing `null` or `undefined` to `app.use`. Throw an error is enough.

yanickrochon:
Well, then, the line should be `debug('use %s', fn && (fn._name || fn.name) || JSON.stringify(fn));` or something. The point of debugging is to debug, otherwise it would be called `info()` or `notice()`. Justifying throwing an error without actually know what value caused the error makes the `debug` call pointless in these cases.

juliangruber:
because of https://github.com/koajs/koa/commit/d134fff9e82d050e81bf579a693ed2114e7946c6#diff-5372f626ee15242f1e2c6eb31655b4faL103 the debug call will always see a function only

yanickrochon:
Some middleware are dynamically generated functions. In some applications, this may result in `undefined` to be returned and passed to `app.use`. There *is* a `debug` call, but it is after checking `fn`. So, in case of error being thrown, `debug` will not specify what value is actually being passed. Unless the error message is modified with `throw new TypeError('middleware must be a function, ` + String(fn) + ' given!');` or something, some information is missing.

PR #608, opened 2015-11-25
V2.x

Pana:
Update docs

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -113,7 +113,7 @@ app.use(convert(function *(next){
 
 
 ## Babel setup
-For node 4.0 and babel 6.0 you can setup like this
+For Node 4.0 and Babel 6.0 you can setup like this
 
 ```bash
 // install babel and required presets
--- a/Readme.md
+++ b/Readme.md
@@ -51,7 +51,7 @@ Middleware normally take two parameters (ctx, next), ctx is the context for one 
 
 Here we write an logger middleware with different function.
 
-####  common function
+###  common function
 ```js
 
 app.use((ctx, next) => {
@@ -63,9 +63,10 @@ app.use((ctx, next) => {
 });
 
 ```
-// Note: only code in `next().then`'s callback will run after downstream finished
 
-#### ___async___ functions (Babel required)
+Note: only code in `next().then`'s callback will run after downstream finished
+
+### ___async___ functions (Babel required)
 
 ```js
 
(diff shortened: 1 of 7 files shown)


tejasmanohar:
i'm +1 on most of this. i'll review some more tomorrow since a lot changed. things aren't perfect, but the docs (especially wording of thing) are something that'll evolve over time

Pana:
My english is not good, some sentence may need change, Thanks for point out

tejasmanohar:
@Pana - no worries. thx for contributing :)

Pana:
I think koa's most outstanding feature is no callback, better error handling. By the way is there an better way to call promise then callback ?

tejasmanohar:
>By the way is there an better way to call promise then callback ?
Not sure what you mean here. Code examples? :)

Pana:
I mean is there a better way to express `then callback`

tejasmanohar:
>next is an function, invoke it will run the downstream middleware and get an promise, you can run some code in it's then callback. 

to

>`next` is an function that is invoked to execute the downstream middleware. It returns a Promise

Promise = then-able -> doesn't need to be stated, imo... but if you think it does ->

>`next` is an function that is invoked to execute the downstream middleware. It returns a "then-able" Promise

or

>`next` is an function that is invoked to execute the downstream middleware. It returns a Promise with a with a then function for running code after completion.

But yeah, wouldn't call it a callback since that term generally refers to a function argument that runs at a specific time (usually after the main task).

Pana:
Thx

Readme.md
```diff
@@ -41,12 +39,21 @@ $ npm install koa
  - [Workshop](https://github.com/koajs/workshop) - A workshop to learn the basics of koa, Express' spiritual successor.
  - [Introduction Screencast](http://knowthen.com/episode-3-koajs-quickstart-guide/) - An introduction to installing and getting started with Koa
 
-## Example
+
+## Middleware
+Koa is an middleware framework, it can take 3 different kind function as middleware:
```
jmealo: This should read "3 different kinds"

jonathanong:
oh shit this is awesome guys. i'll try to take a look more tomorrow!

jonathanong:
shall we just merge this now and continuously improve?

tejasmanohar:
@jonathanong yea, way better than what's there now.

ps: wtf github? i don't think i pushed that :P
wtf github

PR #612, opened 2015-12-02
fix: should not destroy streams

dead-horse:
use black-hole-stream to make sure stream's data has been read.

since user may set `this.body` to a http `IncomingMessage`, if we destroy this stream, it will destroy the http socket that make `keep-alive` failed, and may affect other http requests reusing this socket.
I think maybe we shouldn't destroy streams, only need to insure streams' data will be read and don't leak.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -6,6 +6,7 @@
  */
 
 var contentDisposition = require('content-disposition');
+var BlackHoleStream = require('black-hole-stream');
 var ensureErrorHandler = require('error-inject');
 var getType = require('mime-types').contentType;
 var onFinish = require('on-finished');
@@ -15,6 +16,8 @@ var typeis = require('type-is').is;
 var statuses = require('statuses');
 var destroy = require('destroy');
 var assert = require('assert');
+var Stream = require('stream');
+var http = require('http');
 var path = require('path');
 var vary = require('vary');
 var extname = path.extname;
@@ -161,8 +164,15 @@ module.exports = {
     }
 
     // stream
-    if ('function' == typeof val.pipe) {
-      onFinish(this.res, destroy.bind(null, val));
+    if (val instanceof Stream) {
+      onFinish(this.res, function(){
+        // don't destroy http IncomingMessage, keep `keep-alive` conncetion alive.
+        if (val instanceof http.IncomingMessage) {
+          if (val.readable) val.pipe(new BlackHoleStream());
+        } else {
+          destroy(val);
+        }
+      });
       ensureErrorHandler(val, this.ctx.onerror);
 
       // overwriting
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
   "license": "MIT",
   "dependencies": {
     "accepts": "^1.2.2",
+    "black-hole-stream": "0.0.1",
     "co": "^4.4.0",
     "composition": "^2.1.1",
     "content-disposition": "~0.5.0",
@@ -43,14 +44,18 @@
     "vary": "^1.0.0"
   },
   "devDependencies": {
+    "agentkeepalive": "~2.0.3",
     "babel": "^5.0.0",
+    "freeport": "^1.0.5",
     "istanbul": "^0.4.0",
     "make-lint": "^1.0.1",
     "mocha": "^2.0.1",
+    "pedding": "^1.0.0",
     "should": "^6.0.3",
     "should-http": "0.0.3",
     "supertest": "^1.0.1",
-    "test-console": "^0.7.1"
+    "test-console": "^0.7.1",
+    "urllib": "^2.6.0"
   },
   "engines": {
     "node": ">= 0.12.0",
--- a/test/application.js
+++ b/test/application.js
@@ -2,9 +2,13 @@
 'use strict';
 
 var stderr = require('test-console').stderr;
+var Agent = require('agentkeepalive');
 var request = require('supertest');
 var statuses = require('statuses');
+var freeport = require('freeport');
+var pedding = require('pedding');
 var assert = require('assert');
+var urllib = require('urllib');
 var http = require('http');
 var koa = require('..');
 var fs = require('fs');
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -47,14 +47,15 @@
...
     "should-http": "0.0.3",
     "supertest": "^1.0.1",
-    "test-console": "^0.7.1"
+    "test-console": "^0.7.1",
+    "urllib": "^2.5.0"
   },
   "engines": {
-    "node": ">= 0.12.0",
-    "iojs": ">= 1.0.0"
+    "node": ">= 0.12.0"
```
JacksonTian: node >= 4

dead-horse: can we drop the support of node < 4 in koa@1 ? /cc @tj @jonathanong

JacksonTian: oh, master is v1.

dead-horse:
ping @jonathanong @fengmk2

fengmk2:
LGTM

.travis.yml
```diff
@@ -1,10 +1,8 @@
 language: node_js
 node_js:
   - "0.12"
-  - "1"
-  - "2"
-  - "3"
   - "4"
+  - "5"
```
juliangruber: this should be its own pull request

package.json
juliangruber: this should be its own pull request

test/application.js
```diff
@@ -872,6 +876,69 @@ describe('app.respond', function(){
       .get('/')
       .expect(404, done);
     })
+
+    it('should insure stream do not leak', function(done){
```
juliangruber: insure => ensure

dead-horse:
@juliangruber all fixed

test/application.js
```diff
@@ -872,6 +876,69 @@ describe('app.respond', function(){
       .get('/')
       .expect(404, done);
     })
+
+    it('should ensure stream do not leak', function(done){
+      done = pedding(3, done);
+      var app = koa();
+      let stream1 = fs.createReadStream(__filename);
+      let stream2 = fs.createReadStream(__filename);
+      stream1.once('end', done);
```
juliangruber: i don't understand why this one would end as well? it's never read from

juliangruber:
hmmm, this seems more like an edge case to me. are there other cases that would benefit from this change?

test/application.js
dead-horse: we'll destroy(consume) all stream's that set to `this.body` here https://github.com/koajs/koa/pull/612/files#diff-d86a59ede7d999db4b7bc43cb25a1c11R166

dead-horse:
The most serious problem now is http keepalive sockets will be destroyed, and we must fix this issue. Not sure any other side effect  if we destroy the streams.

juliangruber:
are you sure destroying a http stream will affect other requests on the same socket? sounds more like a node bug to me

dead-horse:
If you destroy an `IncommingMessage`, it will destroy the `socket` behind: https://github.com/nodejs/node/blob/master/lib/_http_incoming.js#L92

That will affect other `IncommingMessage`'s that reusing the same socket and will cause `socket hang up` error. This is reported in https://github.com/aliyun/oss-nodejs-sdk/issues/40 (sorry this issue is in chinese).

dead-horse:
any suggestions? I'll merge this PR if no response in the next day. :)

juliangruber:
Hey man, no need to put a deadline on it. If you need it urgently use a fork or find a userland solution.

dead-horse:
I think this is just a bug fix. so if anyone doubt this patch, I'll leave it here and fix in our own framework for now. :)

juliangruber:
your patch fails if the body stream is endless, like for example:

```js
this.body = fs.createReadStream('/dev/urandom');
```

or more real-worldy

```js
this.body = serverSentEvents(createUpdateStream());
```

in the cases above it's required to destroy the stream.

juliangruber:
and trying to read it all out will never finish

dead-horse:
although i don't think pipe an endless stream to response meaningful. but maybe there is a solution has less impact. detect the stream in `destroy` and don't destroy `IncommingMessage`(a breaking change in `destroy`).

dead-horse:
@juliangruber how about this?

juliangruber:
too edge-casy. koa should not need to worry about this.

Just to be sure, this would work for you, right?

```js
httpStream.destroy = blackHole.bind(null, httpStream);
this.body = httpStream;
```

dead-horse:
it is a common usage that use koa as a http proxy, and anybody who use `this.body = yield request('http://foo.bar/'); //keep-alive` will hit this "edge case", and it is really hard to figure out why. so I think we should fix it in framework level, because it is a bug in koa, and we can't let everyone to use `httpStream.destroy = blackHole.bind(null, httpStream);` hack.

jonathanong:
maybe a better solution is to just wrap the http stream in another stream so you never set the http stream as the actual `.body=`.

```js
const PassThrough = require('stream').PassThrough

app.use(function * (next) {
  this.body = httpStream.on('error', this.onerror).pipe(PassThrough())
})
```

juliangruber:
> would need docs on this though... super edge case

what about documenting that a body stream can be `.destroy`ed and showing how to guard against that

tejasmanohar:
>what about documenting that a body stream can be `.destroy`ed and showing how to guard against that
 Show all checks

+1 @juliangruber

lib/response.js
```diff
@@ -161,8 +164,15 @@ module.exports = {
     }
 
     // stream
-    if ('function' == typeof val.pipe) {
-      onFinish(this.res, destroy.bind(null, val));
+    if (val instanceof Stream) {
+      onFinish(this.res, function(){
+        // don't destroy http IncomingMessage, keep `keep-alive` conncetion alive.
+        if (val instanceof http.IncomingMessage) {
+          if (val.readable) val.pipe(new BlackHoleStream());
```
jonathanong: why do you need black hole stream? just `val.resume()` to dump it

jonathanong:
anyways prefer going the documentation route. will close this when i or someone else adds documentation

dead-horse:
It's ok to add some warning in documentation.

halt-hammerzeit:
Didn't understand a thing from the documentation on what's happening here.
It's just like "warning, never set a stream as ctx.body".

jiajianrong:
@catamphetamine, they were warning not to set a http-request-alike-stream which has keep-alive enabled. Because the framework will close the stream by default, which will break keep-alive feature. 
Keep-alive means to reuse tcp socket connection for two or more http connections.

Generally, it's OK with `ctx.body = fs-read-stream`. 
Just be careful with `ctx.body = request('host:port/path')`.

rhinoceros:
@dead-horse  
cnpmjs.org中目前用的koa的版本是"koa": "^1.2.0", 如果使用oss作为nfs， 在从oss上get的时候会有一堆的sockets hang up。

有什么好的办法能避免这类错误吗？

https://github.com/cnpm/cnpmjs.org/issues/1401

谢谢

image

zhangwilling:
@dead-horse  马哥好，昨天夜里见到这个 issue 忘了跟你打招呼了，白天来踩一下。

----- 骄傲的用中文交流

PR #629, opened 2015-12-23
Allow to use other flow control libraries. Defaults to co.

khc:
Recent Sequelize upgrade depends on bluebird@3 and yielding two or more times inside a generator causes a "_a promise was created in a handler but was not returned from it_" warning.

For example the code:
```javascript
app.use( function* ( next ) {
 var post = yield this.db.Post.findById( this.params.postId );
 var author = yield post.getAuthor();
 this.body = JSON.stringify( author );   
} ); 
```
causes the aforementioned warning.

Bluebird documentation states, that `Promise.coroutine` should be used with `yield`. My change allows to use other flow control solution, for example mentioned Bluebird's `coroutine`.

Example use:
```javascript
var coroutine = require( 'bluebird' ).coroutine;
var bluebird_co = require( 'bluebird-co/manual' );
coroutine.addYieldHandler( bluebird_co.toPromise );

var app = require('koa')();
app.co = coroutine;
```

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -50,6 +50,7 @@ function Application() {
   this.context = Object.create(context);
   this.request = Object.create(request);
   this.response = Object.create(response);
+  this.co = co.wrap;
 }
 
 /**
@@ -121,7 +122,7 @@ app.use = function(fn){
 app.callback = function(){
   var fn = this.experimental
     ? compose_es7(this.middleware)
-    : co.wrap(compose(this.middleware));
+    : this.co(compose(this.middleware));
   var self = this;
 
   if (!this.listeners('error').length) this.on('error', this.onerror);


tj:
Going forward Koa won't be using Co (async/await) at all so I'm -1 for this, it's not something we'd stick with. It'll also break middleware if they depend on some of the features Co provides.

khc:
True. But it's an option, not default behavior. And for now, before `koa@2` release, using Sequelize or Mongoose generates a lot of warnings.

tj:
I'm not really familiar with the lib, but the warnings sound sort of unrelated, I don't see a reason why it shouldn't work with or without Co involved

khc:
Well, it works with `co`, just generates warnings. Sometimes a lot of warnings, and console becomes bloated with stack traces. The resulting json/html/whatever is perfectly fine. Response times are also comparable, no issues there. As I said earlier, it's just an option to prevent those pesky warnings.

jonathanong:
@khc if you're talking about bluebird warnings, just disable them. they're too opinionated and not very accurate. https://github.com/tj/co/pull/256#issuecomment-168475913

PR #635, opened 2016-01-12
chore(package): update should to version 8.1.0

juliangruber:
someone needs to make up for api changes in should, everything now needs to be called as a function, eg `is.a.function()` instead of `is.a.function`


jonathanong:
personally, i'd rather just remove `should`.

PR #644, opened 2016-01-18
[CI] Configure tests for Circle CI as well

ide:
I've seen Circle CI get a lot of praise and they support some nice features like being able to SSH into a test box for a short while after the test has run. Getting the tests to run on Circle is pretty straightforward - there is a YAML file whose format is documented here: https://circleci.com/docs/configuration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,5 +3,5 @@ node_js:
   - "4"
   - "5"
 sudo: false
-script: "make test-travis"
+script: "make test-ci"
 after_script: "npm install coveralls@2 && cat ./coverage/lcov.info | coveralls"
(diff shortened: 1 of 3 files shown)


tj:
unnecessary IMO, there's at least a few dozen CI services out there, I actually prefer semaphore, but travis is fine too

ide:
Where I actually wanted to go with this is to use Circle's Docker support to install a wrk image for performance testing (whether this works is tbd...). As for why Circle instead of something else, several engineers in my network vouched mostly for Circle and a few other CI providers to a lesser extent, Travis being notably absent. I don't have skin in this game and you guys maintain Koa so I won't be offended/upset/whatever if you don't want this though.

tj:
benchmarking would be awesome some day, but I think there's too much variability in multi-tenant vm-based CIs. I'd love that in general if we can prove (ish) that it would be useful but I'm not super convinced it would work well. closing for now

cheers

PR #645, opened 2016-01-19
docs: mention es2016 when discussing async functions

tejasmanohar:


tejasmanohar:
also, off-topic from this PR but thinking it may be helpful to mention something about the release plan for koa v2 in the README. maybe this will reduce all the questions in issues about if Koa v2 is ready to use, when it's coming out, is it ok to still use koa v1, etc.? something like

> The current plan is to release Koa v2 once V8 and then, Node implements `async/await` in a stable release.

and _maybe_ add a line

> The migration from Koa v1 with co to v2 with async/await is very straight-forward and mostly find & replace. [... link to some information about it ?]

thoughts? @jonathanong

targos:
ES7 is not really a thing anymore. It would be more correct to say ES2015/2016

tejasmanohar:
@targos good point, force-push fixed.

PR #650, opened 2016-01-23
Use new operator to create a Koa v2 app

julien-f:
Without it one would get: `TypeError: Class constructors cannot be invoked without 'new'`.


dead-horse:
thanks!

julien-f:
@yanhao691025518 Feel free to open another PR to fix the docs.

PR #661, opened 2016-02-22
JSDoc question/suggestion on optional parameters

bartolkaruza:
Hi, Webstorm keeps giving me warnings on the 'redirect' method, because the JSDoc specified two input parameters, both required. There is a JSDoc standard for optional parameters. What is your view on using these in KOA documentation?
http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -247,7 +247,7 @@ module.exports = {
    *    this.redirect('http://google.com');
    *
    * @param {String} url
-   * @param {String} alt
+   * @param {String} [alt]
    * @api public
    */
 


targos:
LGTM

juliangruber:
+-0, is this the only place this change needs to be applied? consistency is important.

bartolkaruza:
A few more places I found looking beyond my own use of Koa;

https://github.com/koajs/koa/blob/master/lib/response.js#L250
https://github.com/koajs/koa/blob/master/lib/response.js#L278
https://github.com/koajs/koa/blob/master/lib/response.js#L299

I can add all of these to the PR if you want to go ahead with this. Limiting it to the '@api public' functions only.

tj:
:+1:

jonathanong:
:+1: for adding it everywhere!

jonathanong:
@bartolkaruza i will merge this, but we would love correct JSDocs everywhere!

Issue #663, opened 2016-02-26
Request .get should return undefined instead of '' as falsy value for missing headers.


brownstein:
Currently, Koa's request object returns an empty string instead of undefined for undefined headers. I'm requesting that this change to undefined to avoid confusion for greater compatibility with Express's convention.

This change is primarily motivated by issue #30 in [koa-oauth-server](https://github.com/thomseddon/koa-oauth-server), a module which adapts [node-oauth2-server](https://github.com/thomseddon/node-oauth2-server) for Koa. Node Oauth2-Server checks against an undefined literal, causing a false positive when encountering an empty string. While I understand that it is not Koa's responsibility to support 3rd-party modules, I believe that this change has a low level of risk and makes Koa more usable in general in addition to solving this integration issue.

jonathanong:
^ see https://github.com/koajs/koa/pull/664#issuecomment-189978557

brownstein:
Closing this, as it is arguably a potentially breaking change.

PlasmaPower:
@brownstein doesn't this still apply to Koa 2?

PlasmaPower:
An idea if we don't want to make it breaking: add an argument `fallbackValue` which is the value that it falls back to. That argument can default to `''`. You could set this to `undefined` if you want this new behavior.

tj:
no poitn having a fallback param IMO, easy enough to just `|| v`

PlasmaPower:
It would be used to distinguish between undefined and a blank string while still using Koa methods.

PlasmaPower:
This option would be (in my opinion) a better version of a flag to toggle this behavior.

brownstein:
@PlasmaPower I really like that idea, as the alternative could lead to downstream module maintainers proxying requests for compatibility.

PR #664, opened 2016-02-26
Modified request.get to return undefined

brownstein:
This modifies request.get to return undefined for undefined headers, instead of an empty string. This preserves falsiness, but prevents ambiguity relating to empty versus missing headers, and makes the API more similar to Express to reduce module integration pain.

Issue request documented [here](https://github.com/koajs/koa/issues/663).

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -593,9 +593,9 @@ module.exports = {
     switch (field = field.toLowerCase()) {
       case 'referer':
       case 'referrer':
-        return req.headers.referrer || req.headers.referer || undefined;
+        return req.headers.referrer || req.headers.referer;
       default:
-        return req.headers[field] || undefined;
+        return req.headers[field];
     }
   },
 
--- a/lib/request.js
+++ b/lib/request.js
@@ -593,9 +593,9 @@ module.exports = {
     switch (field = field.toLowerCase()) {
       case 'referer':
       case 'referrer':
-        return req.headers.referrer || req.headers.referer || '';
+        return req.headers.referrer || req.headers.referer || undefined;
       default:
-        return req.headers[field] || '';
+        return req.headers[field] || undefined;
     }
   },
 


jonathanong:
anyone have opinions?

tj:
¯\_(ツ)_/¯ – might as well remove the `|| undefined` if we're just leaving it as whatever. I kinda like empty values to at least stay the same type, but it doesn't really matter to me

lib/request.js
```diff
@@ -593,9 +593,9 @@ module.exports = {
     switch (field = field.toLowerCase()) {
       case 'referer':
       case 'referrer':
-        return req.headers.referrer || req.headers.referer || '';
+        return req.headers.referrer || req.headers.referer || undefined;
       default:
-        return req.headers[field] || '';
+        return req.headers[field] || undefined;
     }
```
lourd: Wouldn't this return `undefined` anyway if the `|| undefined` were omitted? Of course nothing wrong with being explicit.

lourd: TJ beat me to it :hurtrealbad:

brownstein: Good call - removed the || altogether.

brownstein:
@tj Good call, fixed. Regarding the build, I don't _think_ this change is causing a regression, but I'm a little concerned by the failing tests. Any concerns on your end?

dead-horse:
all the getters are return empty string when not exist(https://github.com/koajs/koa/blob/master/lib/request.js#L244). and this change may cause some unpredictable bug in users' code(even in koa https://github.com/koajs/koa/blob/master/lib/request.js#L333).

IMO we'd best keep it as-is, at least treat it as a breaking change.

fengmk2:
Empty string is better than `undefined`.

I saw a lot of `TypeError: Cannot read property 'trim' of undefined`.

jonathanong:
yeah agree with ^. it would be a breaking change and there really isn't any benefit. 

if you want us to reconsider:

- make sure tests pass
- make a PR against the v2.x branch

thanks!

PR #666, opened 2016-02-29
proxy-addr + forwarded

jonathanong:
not much benefit so i'm gonna put this on hold

ref: https://github.com/koajs/koa/issues/281

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -6,7 +6,9 @@
  */
 
 var contentType = require('content-type');
+var proxyAddr = require('proxy-addr');
 var stringify = require('url').format;
+var forwarded = require('forwarded');
 var parse = require('parseurl');
 var qs = require('querystring');
 var typeis = require('type-is');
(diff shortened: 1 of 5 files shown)


lib/request.js
```diff
@@ -395,9 +397,8 @@ module.exports = {
 
   get ips() {
     var proxy = this.app.proxy;
-    var val = this.get('X-Forwarded-For');
-    return proxy && val
-      ? val.split(/\s*,\s*/)
+    return proxy
+      ? forwarded(this.req)
```
dougwilson: Depending on what you want `ips` to be, ideally they should get cut off at the trust endpoint. You can remove a dependency with using https://github.com/jshttp/proxy-addr#proxyaddrallreq-trust
```js
return proxyAddr.all(this,req, this.app._proxyTrust);
```

jonathanong: oh shoot. nice.

PR #667, opened 2016-02-29
Add support for flushing headers

### LINKED ISSUE #657: Flush headers option?
I know that this _could_ cause problems, e.g. if someone tries to add headers after those were flushed, but that is a user error.

Sample code:
```javascript
	api.use(route.get('/myendpoint', async ctx => {
		ctx.type = 'text/plain; charset=utf-8';
		ctx.set('My-Queue-Size', queue.size);
		ctx.status = 200;
		// I want to flush headers here
		ctx.body = (await my_method()).stdout;
	}));
```

It would be useful if there was an convenient way to flush the headers before the `await` is fullfilled. Currently, that is impossible due to `respond` being a noop if headers were already sent: [lib/application.js#L184](https://github.com/koajs/koa/blob/2.0.0-alpha.3/lib/application.js#L184).

It's possible to work-around that by creating a PassThrough stream:
```javascript
	api.use(route.get('/myendpoint', async ctx => {
		ctx.type = 'text/plain; charset=utf-8';
		ctx.set('My-Queue-Size', queue.size);
		ctx.status = 200;
		ctx.body = new stream.PassThrough();

		let child;
		let closed = false;
		ctx.res.on('close', () => {
			closed = true;
			if (child) child.kill('SIGKILL');
		});

		setImmediate(async () => {
			ctx.res.flushHeaders();
			child = await my_method();
			child.on('exit', () => { child = undefined; })
			if (closed) child.kill('SIGKILL');
			child.stdout.pipe(ctx.body);
		})
	}));
```

But that doesn't look very nice.

PlasmaPower:
Resolves #657.
This is my first PR to this project, so if I'm doing anything wrong please let me know :smile:.

### DIFF
--- a/docs/error-handling.md
+++ b/docs/error-handling.md
@@ -1,0 +1,44 @@
+# Error Handling
+
+## Try-Catch
+
+  Using generators means that you can try-catch `next`. For example,
+  this example prepends all error messages with "Error: "
+
+  ```js
+  app.use(function*(next){
+    try {
+      yield next;
+    } catch (error) {
+      error.message = 'Error: ' + error.message;
+      throw error;
+    }
+  });
+  ```
+
+### Default Error Handler
+
+  The default error handler is essentially a try-catch at
+  the very beginning of the middleware chain. To use a
+  different error handler, simply put another try-catch at
+  the beginning of the middleware chain, and handle the error
+  there. However, the default error handler is good enough for
+  most use cases. It will use a status code of `err.status`,
+  or by default 500. If `err.expose` is true, then `err.message`
+  will be the reply. Otherwise, a message generated from the
+  error code will be used (e.g. for the code 500 the message
+  "Internal Server Error" will be used). All headers will be
+  cleared from the request, but any headers in `err.headers`
+  will then be set. You can use a try-catch, as specified
+  above, to add a header to this list.
+
+## The Error Event
+
+  Error handlers can be specified with `app.on('error')`.
+  If no error handler is specified, a default error handler
+  is used. Error handlers recieve all errors that make their
+  way back through the middleware chain, if an error is caught
+  and not thrown again, it will not be handled by the error
+  handler. If not error event handler is specified, then
+  `app.onerror` will be used, which simply log the error if
+  `error.expose` is true and `app.silent` is false.
--- a/lib/context.js
+++ b/lib/context.js
@@ -117,8 +117,9 @@ var proto = module.exports = {
       return;
     }
 
-    // unset all headers
+    // unset all headers, and set those specified
     this.res._headers = {};
+    this.set(err.headers);
 
     // force text/plain
     this.type = 'text';
--- a/test/context/onerror.js
+++ b/test/context/onerror.js
@@ -52,6 +52,40 @@ describe('ctx.onerror(err)', function(){
     })
   })
 
+  it('should set headers specified in the error', function(done){
+    var app = koa();
+
+    app.use(function *(next){
+      this.set('Vary', 'Accept-Encoding');
+      this.set('X-CSRF-Token', 'asdf');
+      this.body = 'response';
+
+      throw Object.assign(new Error('boom'), {
+        status: 418,
+        expose: true,
+        headers: {
+          'X-New-Header': 'Value'
+        }
+      })
+    })
+
+    var server = app.listen();
+
+    request(server)
+    .get('/')
+    .expect(418)
+    .expect('Content-Type', 'text/plain; charset=utf-8')
+    .expect('X-New-Header', 'Value')
+    .end(function(err, res){
+      if (err) return done(err);
+
+      res.headers.should.not.have.property('vary');
+      res.headers.should.not.have.property('x-csrf-token');
+
+      done();
+    })
+  })
+
   describe('when invalid err.status', function(){
     describe('not number', function(){
       it('should respond 500', function(done){


test/application.js
```diff
@@ -412,9 +412,12 @@ describe('app.respond', function(){
...
         res.write('Hello');
-        setTimeout(function(){
-          res.end("Goodbye")
-        }, 0);
+        yield new Promise(function(resolve){
+          setTimeout(function(){
+            res.end("Goodbye");
+            resolve();
+          }, 0);
+        });
```
PlasmaPower: This is necessary because before `respond` wouldn't do anything. Now we have to make sure that it doesn't close the socket too soon.

lib/response.js
```diff
@@ -520,5 +522,12 @@ module.exports = {
       message: this.message,
       header: this.header
     };
+  },
+
+  /**
+   * Flush any set headers, and begin the body
+   */
+  flushHeaders: function(){
+    this.res.write('');
```
jonathanong: we should be using `res.writeHead()` and setting the current status code, which may or may not be `200`

test/response/flushHeaders.js
```diff
@@ -0,0 +1,43 @@
...
+    .get('/')
+    .expect(200)
+    .expect('Body', done);
+  })
+
+  it('should allow a response afterwards', function(done){
+    var app = koa();
+
+    app.use(function *(next){
+      this.res.statusCode = 200;
```
jonathanong: we should test with status codes other than `200`

test/application.js
jonathanong: the previous version should still work.

jonathanong: > Now we have to make sure that it doesn't close the socket too soon. 

can you give me a code example?

PlasmaPower: It doesn't, I tested. It seems to consistently not work too.

jonathanong: > the previous version should still work.

i mean the functionality shouldn't change, though i see how it gets complicated. this might be a breaking change, then

PlasmaPower: A valid point in that this might be a breaking change. However, this code relies on the fact that `respond` will do nothing if headers have been sent, which is exactly what this change tries to stop. We could have a variable that would enable this new functionality, but that seems like a hack.

lib/response.js
jonathanong: thought this still works... just seems a little odd. not sure if there's any significant differences

jonathanong:
i think we just need more tests :)

@ChALkeR care to review?

jonathanong:
also, docs!

lib/response.js
PlasmaPower: Using `writeHead` gives
```
TypeError: Cannot read property 'toString' of undefined
     at ServerResponse.writeHead (_http_server.js:190:44)
    at Object.module.exports.flushHeaders (/home/lee/koa/lib/response.js:533:16)
```
but only for the first test. Any ideas why?

dougwilson: That error means somewhere in the code path there was a `res.statusCode = undefined` or `delete res.statusCode`.

jonathanong:
and i think we should throw if the user does `this.status=` after flushing headers unless node already does.

lib/response.js
PlasmaPower: @dougwilson that was my first thought too, but if I add `this.res.statusCode = 200;` in directly before the `writeHead` it gives me the same error.

dougwilson: @PlasmaPower, what is the exact version of Node.js you are using, so I can know for sure what is at line 190 in `_http_server.js`?

PlasmaPower: @dougwilson I'm on node v5.7.0. Where are you getting the source BTW? I couldn't find it with a quick google search.

PlasmaPower:
@jonathanong `this.status=` after flushing now throws, how do I test this though? The error is thrown in the generator, so should I just put the assert there? If the assert fails, it won't give a nice error message though.

lib/response.js
dougwilson: @PlasmaPower https://github.com/nodejs/node/blob/v5.7.0/lib/_http_server.js#L190

The error is that you are not providing the first argument, `statusCode`, to the `writeHead` method.

PlasmaPower:
@jonathanong Docs are done, any more ideas for tests?

PlasmaPower:
Oh, also, should I squash this into one commit before merge?

lib/response.js
PlasmaPower: @dougwilson That's really odd. Can you confirm that the following `flushHeaders` function produces that error?

```
    try {
      this.res.statusCode = 200;
      this.res.writeHead();
    } catch (e) {
      console.error(e.stack);
    }
```

PlasmaPower:
Oh, also, it would appear I don't need to set the current status code, as that's already done by the status setter. I've added a test for other status codes.

lib/response.js
dougwilson: Hi @PlasmaPower, the error is that you are not providing the first argument, `statusCode`, to the `writeHead` method. You code needs to be the following:
```js
try {
  this.res.writeHead(this.res.statusCode);
} catch (e) {
  console.error(e.stack);
}
```

PlasmaPower: Oh oops, I totally missed that.

PlasmaPower:
Since this is a breaking change, even if only for the use case it fixes, should I rewrite this for the 2.x branch?

jonathanong:
@ChALkeR any opinions?

jonathanong:
@PlasmaPower yeah i'm thinking it's best to rewrite from v2.x

PlasmaPower:
I'll make a PR for koa v2 then, it shouldn't be too hard.

jonathanong:
:+1:

mufumbo:
how do you guys use this? we tried to flushHeaders but the headers are only flushed at the end and not as soon as we call the method. Can you please confirm the version that this patch was released to?

PlasmaPower:
@mufumbo The `flushHeaders` method calls the response object's `writeHead`. If the method exists at all, it should work as expected. Do you have evidence that it doesn't?

PR #668, opened 2016-02-29
Add support for headers in errors

### LINKED ISSUE #571: Way to set headers on HttpErrors
Currently, the default error handler in koa checks for the status property on errors and sets the status accordingly. However, many errors in the HTTP spec are defined with required or recommended headers to be send along with the status. For example, 405 Method Not Allowed should always be sent with an _Allow_ header, a 503 Service Unavailable with a _Retry After_.

Currently, it is not possible to set this. You could write your own error handler that checks a `headers` property on the error object, but third party middleware can't rely on that. As a result, middleware modules on npm (like koa-router etc.) have to set status and headers on the context response manually instead of throwing an error, which means it is not possible to catch them and therefor destroys one of the best features of koa - error handling with try / catch.

It would be nice if the koa error handler would check for the existence of a `headers` property on the object and set the headers in that hash (with normalized casing). http-errors already supports passing a properties hash in the constructor, so it could look like this:
```js
throw new HttpError(405, {headers: {'Allow': 'GET, POST'}})
throw new HttpError(503, {headers: {'Retry-After': 60}})
```
This would then allow middlewares from npm to actually throw errors that can be catched to add some additional custom text etc.
Thoughts?

PlasmaPower:
Resolves #571. I'd like to document this, but I'm not quite sure where.

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -117,8 +117,9 @@ var proto = module.exports = {
       return;
     }
 
-    // unset all headers
+    // unset all headers, and set those specified
     this.res._headers = {};
+    this.set(err.headers);
 
     // force text/plain
     this.type = 'text';
--- a/docs/error-handling.md
+++ b/docs/error-handling.md
@@ -1,0 +1,44 @@
+# Error Handling
+
+## Try-Catch
+
+  Using generators means that you can try-catch `next`. For example,
+  this example prepends all error messages with "Error: "
+
+  ```js
+  app.use(function*(next){
+    try {
+      yield next;
+    } catch (error) {
+      error.message = 'Error: ' + error.message;
+      throw error;
+    }
+  });
+  ```
+
+### Default Error Handler
+
+  The default error handler is essentially a try-catch at
+  the very beginning of the middleware chain. To use a
+  different error handler, simply put another try-catch at
+  the beginning of the middleware chain, and handle the error
+  there. However, the default error handler is good enough for
+  most use cases. It will use a status code of `err.status`,
+  or by default 500. If `err.expose` is true, then `err.message`
+  will be the reply. Otherwise, a message generated from the
+  error code will be used (e.g. for the code 500 the message
+  "Internal Server Error" will be used). All headers will be
+  cleared from the request, but any headers in `err.headers`
+  will then be set. You can use a try-catch, as specified
+  above, to add a header to this list.
+
+## The Error Event
+
+  Error handlers can be specified with `app.on('error')`.
+  If no error handler is specified, a default error handler
+  is used. Error handlers recieve all errors that make their
+  way back through the middleware chain, if an error is caught
+  and not thrown again, it will not be handled by the error
+  handler. If not error event handler is specified, then
+  `app.onerror` will be used, which simply log the error if
+  `error.expose` is true and `app.silent` is false.
--- a/test/context/onerror.js
+++ b/test/context/onerror.js
@@ -52,6 +52,40 @@ describe('ctx.onerror(err)', function(){
     })
   })
 
+  it('should set headers specified in the error', function(done){
+    var app = koa();
+
+    app.use(function *(next){
+      this.set('Vary', 'Accept-Encoding');
+      this.set('X-CSRF-Token', 'asdf');
+      this.body = 'response';
+
+      throw Object.assign(new Error('boom'), {
+        status: 418,
+        expose: true,
+        headers: {
+          'X-New-Header': 'Value'
+        }
+      })
+    })
+
+    var server = app.listen();
+
+    request(server)
+    .get('/')
+    .expect(418)
+    .expect('Content-Type', 'text/plain; charset=utf-8')
+    .expect('X-New-Header', 'Value')
+    .end(function(err, res){
+      if (err) return done(err);
+
+      res.headers.should.not.have.property('vary');
+      res.headers.should.not.have.property('x-csrf-token');
+
+      done();
+    })
+  })
+
   describe('when invalid err.status', function(){
     describe('not number', function(){
       it('should respond 500', function(done){


lib/context.js
```diff
@@ -117,8 +117,26 @@ var proto = module.exports = {
...
     }
 
-    // unset all headers
+    // unset all headers not explicitly kept
+    // while it's inconvenient to use setHeader and getHeader,
+    // res._headers isn't documentented and can't be relied on much
+    // for example: the keys are always stored in lowercase
+    var errHeaders = err.headers ? Object.keys(err.headers) : [];
+    var newHeaders = {};
+    for (var i = 0; i < errHeaders.length; i++) {
```
jonathanong: wish we didn't have to support node < 4. then we could do for-of. sigh

jonathanong:
yeah, hmmm... we need a doc just on error handling

jonathanong:
looks good to me.

people who would find interest in this, would you like to review? @niftylettuce @menems @dead-horse @tejasmanohar @cesarandreu @pensierinmusica @felixfbecker @ruimarinho @fixe

lib/context.js
dead-horse: `for (var key in err.headers)` is ok and have the similar performance with a small number of key.

lib/context.js
```diff
@@ -117,8 +117,26 @@ var proto = module.exports = {
...
+      var header = errHeaders[i];
+      if (err.headers[header] === true) { // keep the header
+        newHeaders[header] = this.res.getHeader(header);
+      } else {
+        newHeaders[header] = err.headers[header];
+      }
+    }
     this.res._headers = {};
+    var newHeaderKeys = Object.keys(newHeaders);
+    for (var i = 0; i < newHeaderKeys.length; i++) {
```
dead-horse:
```js
this.set(newHeaders);
```

menems:
hi , I think the ability to add new header on error is a good point , but i am not totally agree about keeping already set header in that way. May be In addition we can have a global var like `app.keepHeaders = true` that keep all headers on error.

As I said, the problem with that approach is,  the module responsible of keeping header have to know which headers are already set. (it's not always the case) and the goal is remove some boilerplate i think .

But I can't judge that because i haven't make a pull request for that issue (missing time sorry)
So maybe other have another opinion?

PlasmaPower:
@menems Yeah this is a bit hacky. Maybe have `err.headers` and `err.keepHeaders`?

dead-horse:
I'd rather remove keep header hack like `err.headers.kept = true`, people can just set `err.headers.kept = this.response.get('kept')`. 

Keep this feature simple and obvious.

PlasmaPower:
@dead-horse Sounds good. I'll implement that in a bit. Would you like something to solve menems' problem, that the error thrower might not know which headers to keep? I think that a `ctx.keepHeadersOnError` object would be a good idea. Maybe a new name though, because that sounds like a boolean.

PlasmaPower:
I also think that a `ctx.keepHeadersOnError` would be a good idea because it could help when errors are thrown in another library.

dead-horse:
not really sure how much should we do in the default error handler, because lots of people already use third party error handlers that don't implement like this may cause some unexpected behavior. so we'd better keep the rules clearly enough and make third party error handlers follow these rules.  

`ctx.keepHeadersOnError` maybe good, but I can't think any use case that we really need this for now.

menems:
@dead-horse my use case, and i think it 's a standard one. I have a cors middleware and an authentication behavior, if i use `ctx.throw(401)`. the headers of the response will be reseting and all cors headers too, so the client will scold me :) because the client want the cors headers even if you don't have permission to access a resource. that because on context error function headers is by default reset.

My only solution is to catch the error and set status to bypass the context.onerror.
````
app.use( (ctx, next) => next.catch(e => ctx.status = e.status))
```
It's a little annoying i think because status will be set by default ctx.onerror handler.

PlasmaPower:
@menems the default error handler already sets the status: https://github.com/koajs/koa/blob/master/lib/context.js#L135

menems:
@PlasmaPower Yes I know but it remove headers too https://github.com/koajs/koa/blob/master/lib/context.js#L121

PlasmaPower:
@menems Oh, I thought that you were then throwing the error again after setting the status. Speaking of that, that might actually eliminate the use case for `ctx.keepHeadersOnError`:

```js
app.use(function*(next){
  try {
    yield next;
  } catch (err) {
    err.headers = err.headers || [];
    err.headers.push('x-my-header');
    throw err;
  }
})
```

lib/context.js
thelinuxlich: Besides, for-of is too slow at the moment

PlasmaPower:
We seem to have [a wiki page on error handling](https://github.com/koajs/koa/wiki/Error-Handling), should I make an edit there once this merges to document this?

I'm not quite sure on why we have [koajs.com](http://koajs.com), [the github wiki](https://github.com/koajs/koa/wiki), and `/docs/`. Should we try to consolidate these?

jonathanong:
koajs.com is built from the docs. we should put it in the docs!

tj:
hahaha good call

PlasmaPower:
@jonathanong Oh, I see. I thought they weren't the same since a recent change wasn't in there. I'll add a docs page for error handling then.

PlasmaPower:
I've added documentation for this.

menems:
@PlasmaPower : ok so if I understand the goal of this pr is to pass header on error and throw it again?
ok no problem i'm going to continu to catch my error like this for keeping headers on response :D
```
app.use( (ctx, next) => next.catch(e => ctx.status = e.status))
```
ps: I'm still thinking this middleware should be unnecessary with a global var that keep headers for status != 5XX

PlasmaPower:
@menems this is more aimed at koa libraries I think. They shouldn't implement their own error handlers. In your case, you should probably just make an `app.on('error')` error handler that does what you want. It will override `ctx.onerror`.

menems:
@PlasmaPower : nop, i tried that before but you can override effectively `app.onerror`, but not `ctx.onerror`because `ctx.onerror`just delegate the error to `app.onerror`but continue to remove headers and other stuff. other solutions are to extends Context object and override the ctx.onerror or, context can be like app ans extends eventEmitter and have the same behaviour of application like here: https://github.com/koajs/koa/blob/v2.x/lib/application.js#L120

PlasmaPower:
I see. Yeah since `ctx.onerror` is basically a try-catch at the start of the chain, then your middleware is practically an error handler. I guess that makes this the recommended way to make an error handler:

```
app.on(async (ctx, next) => {
  try {
    await next();
  } catch (err) {
    err.status = err.status || 500;
    ctx.body = 'An error occured: ' + err.status;
    ctx.status = err.status;
  }
})
```

thelinuxlich:
If you do that, aren't you deoptimizing the code?

PlasmaPower:
@thelinuxlich What do you mean?

thelinuxlich:
try/catch deoptimizes V8 code

PlasmaPower:
@thelinuxlich I wasn't aware of that. However, doing a bit of research, it would appear that it will not apply to the next call, so the impact is minimal. Is this the case? Also, if try-catch is to be avoided, you can always just do `.catch` on a promise:

```js
app.on((ctx, next) => {
  return next().catch(function(err){
    err.status = err.status || 500;
    ctx.body = 'An error occured: ' + err.status;
    ctx.status = err.status;
  })
})
```

PlasmaPower:
`koa-compose@next` try-catches middleware anyway, so I'm not sure any of that matters. https://github.com/koajs/compose/blob/next/index.js#L40

thelinuxlich:
hmmm interesting

jonathanong:
tried to see if removing the try/catch was faster... it was not. https://github.com/koajs/compose/pull/49

jonathanong:
i believe it only deoptimizes that function, anyway

PlasmaPower:
> it's actually faster w/ try/catch!

I propose adding 5 more try-catches for optimal performance /s

felixfbecker:
+1 on more try/catches. Maybe add a finally block aswell?

jonathanong:
anyways, this LGTM

PlasmaPower:
I need to fix the docs before merging, it should be done soon. I was a big confused on `ctx.onerror` and `app.onerror`.

PlasmaPower:
Okay they should be good now.

dead-horse:
LGTM

PlasmaPower:
Small change to reflect current implementation: `it('should keep headers specified in the error'` -> `it('should set headers specified in the error'`.

PR #672, opened 2016-03-03
lint: import eslint from v2.x

jonathanong:


PlasmaPower:
Would I be correct to assume that travis isn't checking eslint? Should this be fixed?

test/application.js
```diff
@@ -27,7 +26,7 @@ describe('app', function(){
     request(app.listen())
     .get('/')
     .end(function(){});
-  })
+  });
 
   it('should not .writeHead when !socket.writable', function(done){
```
PlasmaPower: Why isn't there a space before the `{` like in the diffs above (`/lib`)? Is it because this is part of a function call?

jonathanong: i just copied whatever was in v2.x. yeah, they should be the same

jonathanong:
yeah, need to add the script to travis

test/application.js
PlasmaPower: Which are we standardizing on, and shouldn't it be specified in the eslintrc?

jonathanong:
@stevepeak hey how do i switch this over to codecov? i can't figure out why coveralls is a status check and not codecov

jonathanong:
gonna try opening a new PR

PR #673, opened 2016-03-03
use codecov instead of coveralls

jonathanong:


PlasmaPower:
> coverage/coveralls — Waiting for status to be reported

Yeah, umm, about that :P

PR #675, opened 2016-03-04
Add support for flushing headers

PlasmaPower:
This is a breaking change, as seen by the test change, but it is very unlikely to break things. Previous discussion was at #668. Any more ideas for tests?


PlasmaPower:
Branch coverage dropped a branch from 218 to 217, but statement coverage remained the same. However, codecov didn't seem to process this PR, so I don't know which branch it is.

jonathanong:
ah crap... because i need to add codecov to v2.x as well. damnit

PlasmaPower:
I thought that might be the case. Also, coveralls still seems to be running, I'm assuming we want to remove that?

test/response/flushHeaders.js
```diff
@@ -0,0 +1,98 @@
...
+      }
+      try {
+        ctx.status = 200;
+      } catch (err) {
+        ctx.body += 'ctx.status fail ';
+      }
+      try {
+        ctx.length = 10;
+      } catch (err) {
+        ctx.body += 'ctx.length fail';
```
jonathanong: interesting way to make sure things are asserted... haha

PR #677, opened 2016-03-09
context.state can inherit app.context.state

Gerhut:
In some middlewares like [koa-views](/queckezz/koa-views/), `ctx.state` will automatically inject to view engine. Therefore, if I want to inject some global functions to `ctx.state`, at present I have to build a middleware for it. Although there are some other ways to implement that, I think give `ctx.state` an access at application level will make context self less polluted.

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -48,7 +48,7 @@ function Application() {
   this.middleware = [];
   this.proxy = false;
   this.context = Object.create(context);
-  this.context.state = {};
+  this.state = {};
   this.request = Object.create(request);
   this.response = Object.create(response);
 }
@@ -163,8 +163,8 @@ app.createContext = function(req, res){
     secure: request.secure
   });
   context.accept = request.accept = accepts(req);
-  if (typeof this.context.state === 'object') {
-    context.state = Object.create(this.context.state);
+  if (typeof this.state === 'object') {
+    context.state = Object.create(this.state);
   } else {
     context.state = {};
   }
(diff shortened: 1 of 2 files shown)


dead-horse:
why not use `app.state` ?

Gerhut:
`app.state` sure is OK.

In fact I just want to prevent the overwriting of `app.context.state`. On the other hand, `app.context.state` may be more similar with the current behavior, looks like a bugfix; `app.state` inheritance may cause existing code have a different behavior if they had already used `app.state`, looks more like a new feature.

PlasmaPower:
I don't like the idea of passing the application object to a library, instead of the library returning middleware which can be `app.use`d (or similar like `koa-router`). I think the latter approach works just fine, and I'm not sure how this would make `ctx.state` any less polluted. In fact, the latter approach's flexibility allows for a more limited `ctx.state` - if you have a library mounted on `/example` then only requests to `/example` have the library's state additions.

Gerhut:
@PlasmaPower 

Setting such like `app.state` is not the work of library but the work of application(business), just instead of setting `app.context` directly. In my understand, `context` is for library and `ctx.state` is for business. Code for business should control the whole app. So giving the `state` access at application level may be needed.

jonathanong:
@koajs/owners is this something we want to add? i'm -0

juliangruber:
-1 from me, koa doesn't need to be responsible for this

fengmk2:
Should be impl this in your middleware.

I'm -1

jonathanong:
@Gerhut thanks, but :(

PR #679, opened 2016-03-13
Remove unused http variable (v2.x)

PlasmaPower:
ESLint now runs fine. This is why we need eslint in travis.

### DIFF
--- a/benchmarks/middleware.js
+++ b/benchmarks/middleware.js
@@ -1,7 +1,6 @@
 
 'use strict';
 
-const http = require('http');
 const Koa = require('..');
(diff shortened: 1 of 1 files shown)


jonathanong:
oh shoot. is it not running in travis?

jonathanong:
how come this isn't failing for me locally? brb `rm -rf node_modules; npm i`

jonathanong:
oh because we aren't linting that folder

PlasmaPower:
@jonathanong oops - was looking at master not v2.x when checking if eslint was run.

PR #683, opened 2016-03-14
convert generator-mw with deprecation warning

fl0w:
As discussed with @PlasmaPower in #533 (from [issuecomment-196293199](https://github.com/koajs/koa/issues/533#issuecomment-196293199) to [issuecomment-196326368](https://github.com/koajs/koa/issues/533#issuecomment-196326368))

Enable support for old-style middleware using `koa-convert` with an added deprecation warning.

Left tests as is for now to illustrate impact on current coverage.


PlasmaPower:
You need to fix the existing generator test [here](https://github.com/koajs/koa/blob/v2.x/test/application/use.js#L63). I would recommend also testing that the middleware handles a request as expected.

fl0w:
Aye, as noted that in the description (that I haven't updated the tests yet). I wasn't sure if this was a direction you were willing to take this; as we discussed deprecating usage in `koa-compose` - and I was unsure if `util.deprecate` wrapping was done to taste. I'm on the run and will try to get to it as soon as possible.

PlasmaPower:
I'm fine with the util.deprecate logging, the one thing is that it won't log the name of the middleware that is using a generator, but hardly anybody names their generators anyway and it should be fairly easy to spot.

fl0w:
I'm trying to figure out how I can catch and/or assert that a util.deprecate output was made. Any pointers?

jonathanong:
if we're okay with including `koa-convert` (which we weren't the last time i remember discussing it), we should use https://github.com/dougwilson/nodejs-depd because it has a better UX - it tells you where you called the deprecated function

fl0w:
@jonathanong saw same usage from cookie (throws deprecation during tests btw). Updating PR as we speak for depd and to include tests.

fl0w:
@tejasmanohar I reverted to "old style" in master (one line, w/ body). Figured there's a reason that specific line exceeds ~72 character limit.

fl0w:
Oops, there seems to be some linting issues. I assumed linting occurred while doing `npm run test`. One second.

lib/application.js
```diff
@@ -100,7 +104,7 @@ module.exports = class Application extends Emitter {
 
   use(fn) {
     if (typeof fn !== 'function') throw new TypeError('middleware must be a function!');
-    if (isGeneratorFunction(fn)) throw new TypeError('Support for generators has been removed. See the documentation for examples of how to convert old middleware https://github.com/koajs/koa/tree/v2.x#old-signature-middleware-v1x');
+    if (isGeneratorFunction(fn)) { fn = convert(fn); deprecate('Support for generators will been removed. See the documentation for examples of how to convert old middleware https://github.com/koajs/koa/tree/v2.x#old-signature-middleware-v1x'); }
```
PlasmaPower: You should probably split this up into two lines now that it isn't a single statement.

PlasmaPower: Also, we might want to change that link/not link to it since it just says to use koa-convert.

tejasmanohar: +1 for both

tejasmanohar:
What's the goal of doing this? I'd prefer we don't take the responsibility of koa-convert in Koa itself.

PlasmaPower:
@tejasmanohar to make the new koa less of a breaking change. It would follow the Ember and React style where you deprecate something in v+1 and remove it in v+2. In addition, we wouldn't be taking responsibility, it would be *less* of a breaking change, this wouldn't be guaranteed to not break things. As for taking responsibility of the core of koa-convert, it's not that bad, it's pretty much just `co.call` wrapped in a function.

tejasmanohar:
@PlasmaPower I understand, but since most but not everything will work out-of-the-box with koa-convert ([example](https://github.com/koajs/mount/issues/42)), we're bound to receive lots of issues about why X works in Koa v1 but not v2. A clear differentiation between what signature and return values v1 and v2 accept makes things a lot simpler to wrap your head around and support.

fl0w:
Updated deprecation message accordingly.

@tejasmanohar I understand that concern; I am new so my input isn't necessarily valid. I drew inspiration from Ember. This PR arose from the bikeshedding in #533 regarding pro's and con's for releasing `v2` before async/await is natively implemented. What this "solves" is being able to release v2 as a stepping stone to an eventual v3 which would release in sync with a native async/await solution.

People want/use `v2` already (with and without a build process). There's numerous examples where people already use `v2` and do not understand how to use generator middleware in `v2` ([example](https://github.com/koajs/jwt/issues/34)). However, this might just be a documentation thing - which I'm willing to take a stab at as well if allowed.

I made this PR to illustrate a potential solution as discussed in #533, which actually was about introducing a deprecation message in koa-convert or koa-compose. I figured I'll alleviate @PlasmaPower from this suggestion while you Koa.js hackers make a decision.

jonathanong:
@koajs yes or no? i don't mind adding this for v2 and removing it in v3.

lourd:
Correct me if I'm wrong, but if this is added in then there wouldn't be any breaking changes, right? You could publish the current v2 branch as a new minor version (1.3), and then 2.x will be the one that removes deprecation warnings.

I'm not one for strict semver doctrine, but I think that makes more sense than bumping the major version twice for one breaking change.

PlasmaPower:
@lourd The flushing headers support breaks an edge case, and deprecating generators means deprecating the entire codebase of v1.

dead-horse:
+1 for add generator function support in v2.

fl0w:
Two questions
1. Do you want me to add an entry about this to README.md in this PR or should that be done in a separate PR?
2. The depd message; is it sufficient or should we add a more warning e.g. "you milage may vary, strongly recommend v2 middleware"

PlasmaPower:
@fl0w A small note about this in the README might be helpful, maybe modify the note about koa-convert to include this. I think the depd message is sufficient.

PlasmaPower:
Somehow this branch is out of date from `v2.x`, you should probably rebase.

PlasmaPower:
A problem with this is that then generators would work with `app.use`, but not `router.use` or `app.use(mount(`. Should it be implemented in other places using middleware too? That might get a bit messy, but it's nicely wrapped up in koa-convert.

jonathanong:
yeah, docs in the readme are always helpful. 

for depd, as long as you say something along the lines of, "generator usage is deprecated in v2.x and will be removed in v3.x. please use the new koa v2 signature w/ async functions".

lib/application.js
```diff
@@ -100,7 +104,11 @@ module.exports = class Application extends Emitter {
 
   use(fn) {
     if (typeof fn !== 'function') throw new TypeError('middleware must be a function!');
-    if (isGeneratorFunction(fn)) throw new TypeError('Support for generators has been removed. See the documentation for examples of how to convert old middleware https://github.com/koajs/koa/tree/v2.x#old-signature-middleware-v1x');
+    if (isGeneratorFunction(fn)) {
+      deprecate('Support for generators will been removed. ' +
```
jonathanong: let's add that it'll be removed in `v3` and have a link to the docs

jonathanong:
> A problem with this is that then generators would work with app.use, but not router.use or app.use(mount(. Should it be implemented in other places using middleware too? That might get a bit messy, but it's nicely wrapped up in koa-convert.

yeah, that'll have to be done in those middleware

otherwise, LGTM!

fl0w:
Ok - give me a few hours (on the run) and I'll:
 * add a paragraph to README
 * update depd message
 * rebase

fl0w:
I would prefer #686 merged before this so I can update README and tests accordingly.

jonathanong:
@fl0w #686 has been merged!

fl0w:
@jonathanong updated.

I apologise however. I could have sworn that PR introduced markdown linting which I wanted to conform with.

PlasmaPower:
@fl0w it did, but the linter would force us to make some style changes that we didn't want so we disabled it.

PR #685, opened 2016-03-15
Lint before running test

robinpokorny:
To ensure code style lint should be run before testing the code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "Koa web app framework",
   "main": "lib/application.js",
   "scripts": {
-    "test": "make test"
+    "test": "make lint test"
   },
   "repository": "koajs/koa",
   "keywords": [


PlasmaPower:
We already have it running in travis, at least for v2.x.

robinpokorny:
I am aware of that but that could be quite late and frustrating for the developer. Style should be checked as soon as possible—but that is my personal opinion.

robinpokorny:
And this is a PR to v2. As long as I know we do not lint v1.

PlasmaPower:
@robinpokorny we recently started linting v1. Also, I'm not sure many people even run `npm test` as opposed to `make test`.

robinpokorny:
@PlasmaPower, I see. Then this is probably useless.

The idea was that developers might not be aware of the linter, committing wrongly-styled test-passing code and discover the style does not pass only after the CI runs.

Maybe there could be a *Contributing* section/file where one would learn what should be done before submitting a PR. What do you think?

PlasmaPower:
I agree that forgetting to lint is an issue, I often forget myself. As for a contributing file, I'm not sure that would help much more than travis and a friendly comment, especially since that I haven't seen linting fail much.

jonathanong:
i'm -1 because if i need to lint, i'll run `npm run lint` or whatever. i'll know linting fails in CI

robinpokorny:
I see, slowing the test down is a good argument.

Then I would just add the `npm run lint` @jonathanong mentioned and which does not exist now. WDYT?

fl0w:
I would much rather read a contributor doc that explains what's expected and having CI follow those same guidelines.

jonathanong:
@robinpokorny i'm fine with an `npm run lint` command.

@fl0w yes we need a contributor doc :)

PR #686, opened 2016-03-15
Correct code style of JavaScript in Markdown

robinpokorny:
Use the same code style in Markdown files as in the rest of the codebase.

**Updated:** Due to some problems described in the comments it was not possible to automate the linter. This could be reconsidered when eslint/eslint-plugin-markdown#31 and eslint/eslint-plugin-markdown#46 are solved.

A follow-up of #680.

Also, I added the `.yml` to `.eslintrc` as usage without extension is now [deprecated](http://eslint.org/docs/user-guide/configuring#configuration-file-formats).

---

~~Use [eslint-plugin-markdown](https://github.com/eslint/eslint-plugin-markdown) to lint Javascript examples in Markdown.~~

~~The `.eslintrc.md.yml` configuration extends configuration for code with this exceptions:~~
 * ~~disables *Require file to end with single newline*~~
 * ~~allows use of undefined variables (mostly `app`) and defining variables without using them.~~

~~Unfortunately, `yield` outside a generator function results in a parsing error. Therefore I wrap those in anonymous generator (two occurrences).~~


docs/api/request.md
```diff
@@ -105,7 +105,7 @@ const ct = this.request.type;
   Get request charset when present, or `undefined`:
 
 ```js
-this.request.charset
+this.request.charset;
```
PlasmaPower: I don't like these changes, any way we can fix this?

docs/api/request.md
```diff
@@ -117,10 +117,10 @@ this.request.charset
 
   For example "color=blue&size=small":
 
-```js
+```json
 {
-  color: 'blue',
-  size: 'small'
+  "color": "blue",
+  "size": "small"
```
PlasmaPower: I prefer this without the extra quotes, will the markdown linter not let us do that?

package.json
```diff
@@ -51,6 +51,7 @@
     "eslint": "~2.2.0",
     "eslint-config-standard": "^5.1.0",
     "eslint-plugin-babel": "^3.1.0",
+    "eslint-plugin-markdown": "^1.0.0-beta.1",
```
jonathanong: when's the stable version being releaseD?

docs/api/request.md
robinpokorny: Well, there is no nice solution. It is discussed on the plugin page: eslint/eslint-plugin-markdown#46.

These are possible solutions:

* It could be left unhighlighted.
* It could be marked as a different but similar language, e.g. Typescript.
* It could be disabled with `// eslint-disable-line` or `// eslint-disable-next-line`.
* Semicolons could be optional in Markdown.
* Some files could be unlinted.

Is there one that would be acceptable?

package.json
robinpokorny: Hopefully soon: eslint/eslint-plugin-markdown#31

jonathanong: i've had enough problems dealing with betas w/ babel-eslint. i'd say to at least wait until the stable version comes out.

but a lot of these changes are kind of annoying as pointed out by @PlasmaPower

docs/api/request.md
robinpokorny: Unfortunately, this is not valid JS code:

```
{
  color: 'blue',
  size: 'small'
}
```

It is closer to JSON, but then it misses the quotation marks.

PlasmaPower: Why is it not valid JS? No semicolon?

robinpokorny: The object notation represents an expression. However, JS code consist of statements. It is possible to have *Expression Statements* (e.g. `'use strict';`) but these may not ‘start with an opening curly brace because that might make it ambiguous with a Block’, see http://es5.github.io/#x12.4. Wrapping it in parentheses would make it valid, though.

robinpokorny:
@PlasmaPower, @jonathanong So I propose to remove the annoying changes (one line with a semicolon, object notation), also remove the linter until it is stable and preferably has a way to be disabled somewhere.

PlasmaPower:
@robinpokorny agreed, sounds like a good plan of action.

robinpokorny:
PR is updated please review it again.

PlasmaPower:
You might want to change the title of the PR.

docs/api/context.md
```diff
@@ -12,7 +12,7 @@
   snippet:
 
 ```js
-app.use(function *(){
+app.use(function * () {
```
PlasmaPower: I thought we used `*()` in our code base, am I incorrect?

robinpokorny: In v2 we only use it [once](https://github.com/koajs/koa/blob/v2.x/test/application/use.js#L66) without space. Usage in docs is [inconsistent](https://github.com/koajs/koa/pull/686/files/340dd4f1a3c0b715e259bbcbea9d05aeb79619c1#diff-0f569af1b88af54931448ef15c1c26b1R151).

In v1 it is sometimes [with space](https://github.com/koajs/koa/blob/master/test/request/href.js#L29) but [mostly](https://github.com/koajs/koa/blob/master/test/application.js#L17) without.

In standard.js it is [defined](https://github.com/feross/eslint-config-standard/blob/master/eslintrc.json#L41) to have [spaces around tha star](http://eslint.org/docs/rules/generator-star-spacing) and that is what I used.

I can unify it if it is clear what to use.

jonathanong:
LGTM. anyone else have comments?

juliangruber:
lgtm as well

PR #691, opened 2016-03-22
remove outdated wiki reference

fl0w:
Kind of weird. The wiki reference is outdated, either update or remove the reference.

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -159,8 +159,7 @@ http.createServer(app.callback()).listen(3001);
 
 ## app.use(function)
 
-  Add the given middleware function to this application. See [Middleware](https://github.com/koajs/koa/wiki#middleware) for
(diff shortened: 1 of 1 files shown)


PlasmaPower:
It looks fine to me. Sure, some of the middleware doesn't support v2, but we could add a "supports v2" column or something.

fl0w:
Alright, closing this then. I'll try to "PR" that wiki entry with a separate issue adding a support v2 column. I'll do an initial scan for compatibility.

PlasmaPower:
@fl0w okay, makes sense. Also, for the wiki it's not a PR, just an edit that automatically goes through.

fl0w:
@PlasmaPower aye, "PR" :) I figured I'd make changes to my fork and reference it for acceptance.

PR #693, opened 2016-03-24
Allow passing a wrapper-mw to compose

fl0w:
This was done to get the ball rolling to solve #219.

Opted for usage of the term `wrap`.

There's discussion on if the wrapping should/could be an array or not.
Related PR: https://github.com/koajs/compose/pull/51


fl0w:
I'm going to close this for now; discussion in koajs/compose

PR #694, opened 2016-03-24
Add ctx.wrappers array

PlasmaPower:
Goes along with koajs/compose#52.

### DIFF
--- a/docs/api/context.md
+++ b/docs/api/context.md
@@ -58,6 +58,15 @@ app.use(async (ctx, next) => {
 ctx.state.user = await User.find(id);
 ```
 
+### ctx.wrappers
+
+  Stores functions to wrap middleware (e.g. co.wrap could be used).
+  Note that koa itself does not implement the use of these wrappers, they are implemented by utilities such as koa-compose.
+
+```js
+ctx.wrappers.push(co.wrap);
+```
+
 ### ctx.app
 
   Application instance reference.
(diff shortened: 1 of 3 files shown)


PlasmaPower:
Copied from another conversation:

If we implement this, I'd be in favor of removing the deprecated generator to promise conversion in favor of `koa-convert` providing a wrapper to convert generators to promise-based middleware. I think that would be the more versatile approach, as some people might want to use generators with the new argument syntax until async/await arrives. The current approach would convert them as if they took ctx as this instead of an argument, and you'd have to extend the Koa Application to change that.

jonathanong:
also see https://github.com/koajs/compose/pull/51

jonathanong:
would really love more people to express their opinions, otherwise, I'm inclined to just merge it!

PlasmaPower:
This and #707 preform the same purpose different ways, so only one should be merged. Do you prefer this one?

miwnwski:
I'll just close this as it has little relevance today IMO. Maybe there's still a need for wrappers but I doubt it personally and the same functionality can be approached in user land. Also, over the years I haven't really seen any requests for wrappers. If I remember correctly, this was brought to light due to TJ's original issue to simplify mw profiling, though today this can be solved with async_hooks which seems more like native approach Koa should aim towards. Please reopen if I'm wrong :)

PR #698, opened 2016-03-25
Cache the request IP

### LINKED ISSUE #695: Sometimes request.ip returns undefined
After certain types of errors (e.g. `{ [Error: Parse Error] bytesParsed: 159, code: 'HPE_INVALID_METHOD' }`) the remote socket gets closed, and calling `.remoteAddress` on it will return `undefined`. 

Because koa's `request.ip` is lazy, this means it's sometimes returning  `undefined` which screws up some things (e.g. I record errors in a format that requires an ip, and would like to see the ip address of the bad requests).

I think this is quite unexpected behavior, and would be more sensible to strictly evaluate `.ip` to always be available

PlasmaPower:
Resolves #695

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -158,6 +158,7 @@ module.exports = class Application extends Emitter {
       keys: this.keys,
       secure: request.secure
     });
+    request.ip = request.ips[0] || req.socket.remoteAddress || '';
     context.accept = request.accept = accepts(req);
     context.state = {};
     return context;
--- a/test/helpers/context.js
+++ b/test/helpers/context.js
@@ -4,18 +4,18 @@
 const Stream = require('stream');
 const Koa = require('../..');
 
-module.exports = (req, res) => {
+module.exports = (req, res, app) => {
   const socket = new Stream.Duplex();
-  req = req || { headers: {}, socket: socket, __proto__: Stream.Readable.prototype };
-  res = res || { _headers: {}, socket: socket, __proto__: Stream.Writable.prototype };
-  req.socket = req.socket || socket;
-  res.socket = res.socket || socket;
+  req = Object.assign({ headers: {}, socket }, Stream.Readable.prototype, req);
+  res = Object.assign({ _headers: {}, socket }, Stream.Writable.prototype, res);
+  req.socket.remoteAddress = req.socket.remoteAddress || '127.0.0.1';
+  app = app || new Koa();
   res.getHeader = k => res._headers[k.toLowerCase()];
   res.setHeader = (k, v) => res._headers[k.toLowerCase()] = v;
   res.removeHeader = (k, v) => delete res._headers[k.toLowerCase()];
-  return (new Koa()).createContext(req, res);
+  return app.createContext(req, res);
 };
 
-module.exports.request = (req, res) => module.exports(req, res).request;
+module.exports.request = (req, res, app) => module.exports(req, res, app).request;
 
-module.exports.response = (req, res) => module.exports(req, res).response;
+module.exports.response = (req, res, app) => module.exports(req, res, app).response;
--- a/lib/request.js
+++ b/lib/request.js
@@ -367,19 +367,6 @@ module.exports = {
   },
 
   /**
-   * Return the remote address, or when
-   * `app.proxy` is `true` return
-   * the upstream addr.
-   *
-   * @return {String}
-   * @api public
-   */
-
-  get ip() {
-    return this.ips[0] || this.socket.remoteAddress || '';
-  },
-
-  /**
    * When `app.proxy` is `true`, parse
    * the "X-Forwarded-For" ip address list.
    *
--- a/test/request/ip.js
+++ b/test/request/ip.js
@@ -1,32 +1,49 @@
 
 'use strict';
 
-const request = require('../helpers/context').request;
+const assert = require('assert');
+const Stream = require('stream');
+const Koa = require('../..');
+const Request = require('../helpers/context').request;
 
 describe('req.ip', () => {
   describe('with req.ips present', () => {
     it('should return req.ips[0]', () => {
-      const req = request();
-      req.app.proxy = true;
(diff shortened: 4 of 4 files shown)


lib/application.js
```diff
@@ -158,6 +158,11 @@ module.exports = class Application extends Emitter {
       keys: this.keys,
       secure: request.secure
     });
+    if (req && req.headers) {
+      request.ip = request.ips[0] || req.socket.remoteAddress || '';
+    } else {
+      request.ip = ''; // We are in a badly written test :P
```
PlasmaPower: Should I also check the environment? I wasn't quite sure on this one.

test/helpers/context.js
```diff
@@ -4,7 +4,7 @@
 const Stream = require('stream');
 const Koa = require('../..');
 
-module.exports = (req, res) => {
+module.exports = (req, res, app) => {
```
PlasmaPower: Needed to set `app.proxy` before the context is created.

lib/application.js
RHavar: I don't think using '' is particularly better than `undefined`. I'd remove the || ''

PlasmaPower: That's the current behavior, changing that would be breaking for no reason. Also, it's better because it keeps the type.

RHavar:
Nice work, if no one objects to it -- it'd be nice to backport this fix to koa 1 as well

dougwilson:
Since removeAddress calls into native code, I wonder what the performance hit this will add to every request, for those bot even getting the IP in their code.

PlasmaPower:
Travis is running benchmarks now, let me check

lib/application.js
RHavar: Well I think the nice thing would be to guarantee that every request has an ip. That removes the burden from users of the library to wonder (and check) if it's  ''. I don't really see any reason why a request should be processed that can't get an ip

PlasmaPower:
There doesn't seem to be a noticeable difference on my machine, on Travis it's 10% faster but I'd assume that's not reliable.

lib/application.js
PlasmaPower: We can't guarantee that in this section of code though, if `req.socket.remoteAddress` is undefined what do we do? Make up an IP?

dougwilson:
So sounds negligible.

lib/application.js
RHavar: Or error out? Perhaps we shouldn't even be continuing.

PlasmaPower: A getter shouldn't error.

RHavar: I don't think an error breaks any behavior, because the request will merely not be created and not processed. From a middlware's perspective, it simply never gets called

kraghack:
Shouldn't the stable major-point release of Koa (v1) be patched as well instead of just the v2 version that's not going to be stable until async/await lands in v8?

lib/application.js
PlasmaPower: Oh, sorry I thought this was still a getter for request.ip. I guess throwing an error might work then. Anyone else have an opinion?

PlasmaPower:
@kraghack I'm pretty sure v2 is going to be stable sooner than async/await.

lib/application.js
PlasmaPower: Another problem is I'd have to be checking `req && req.headers` anyway because of badly written tests.

dead-horse: we'd better fix those test cases rather than add some check here IMO.

PlasmaPower: Fixing the test cases wouldn't be easy, there are a lot of them and with the above mentioned change we'd have to add both headers and ip. However, I should be able to "fill in" missing attributes.

Also @dead-horse do you think we should throw here if there's no IP? That would essentially mean the request doesn't get processed.

PlasmaPower:
Okay I changed a bunch of things up but it can't be backported to 0.12 easily now.

jonathanong:
looks fine to me. anyone else? @Koajs

lib/application.js
dead-horse: I'm not sure if we should throw here(don't know what's the impact when can't get IP), but at lease we should give more detailed information here rather than just an assert.

jonathanong:
@PlasmaPower i think the tests will pass again if you rebase. i upgraded the eslint packages

lib/application.js
PlasmaPower: > don't know what's the impact when can't get IP

The request won't be handled. An app error will occur I think.
If we're not throwing, what do you think the default should be @dead-horse? Undefined? An empty string?

PlasmaPower:
Yep, tests seem to be passing now.

lib/application.js
dead-horse: after reading https://nodejs.org/api/net.html#net_socket_remoteaddress , I'm -1 on throw here, lose `ip` here mainly cause by client side error(etc: client disconnected), but assert here may record as a server side error(500). 

I'd like default to empty string so the return type of `request.ip` keep the same.

PlasmaPower: Well I guess we could try-catch createContext and if it fails return. A request where the client immediately disconnected shouldn't be processed IMO.

dead-horse: maybe no need to do to many things in the framework level. and in the real world, most of node server is behind a `nginx` server and rarely in this situation.

PlasmaPower: This could remove rate limiting if the attacker doesn't need the result of the request. I'd prefer we don't handle the request. However, the impact is minimal so I'll change it back to `|| ''` unless anybody else has a comment.

PlasmaPower:
It now defaults to '' at the request of @dead-horse.

dead-horse:
LGTM

fengmk2:
+1

PlasmaPower:
I just rebased, tests are still passing, this should be ready to merge.

luckydrq:
+1. Should we also make a PR for koa v1.x then?

PR #699, opened 2016-03-27
Correct ctx.assert documentation parameter order (v2.x)

PlasmaPower:
Essentially #696 for v2.x


fl0w:
Wasn't this fixed in https://github.com/koajs/koa/pull/696/files?

PlasmaPower:
Yes but that was for master, not v2.x.

fl0w:
@PlasmaPower hah, my bad. the iPhone app didn't do your patch justice :+1:

PR #702, opened 2016-04-03
Move to container travis builds

PlasmaPower:
These allow for caching, and besides the old ones are deprecated. The only disadvantage to using caching is that once wrk is cached, it won't be updated again until the cache is manually cleared. Hopefully that won't be too big of an issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,14 @@
+sudo: false
 language: node_js
 node_js:
   - "4"
   - "5"
+cache:
+  directories:
+    - wrk/bin
 before_script:
-  - sudo apt-get install build-essential libssl-dev git
-  - git clone https://github.com/wg/wrk.git
-  - cd wrk && make && sudo cp wrk /usr/local/bin && cd ..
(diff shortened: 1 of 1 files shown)


PlasmaPower:
The travis build is now passing! I can't confirm that it will work once the cache exists as PRs can't write to the cache, but I'm pretty sure it will.

.travis.yml
```diff
@@ -1,11 +1,14 @@
...
   - "4"
   - "5"
+cache:
+  directories:
+    - wrk/bin
 before_script:
-  - sudo apt-get install build-essential libssl-dev git
-  - git clone https://github.com/wg/wrk.git
-  - cd wrk && make && sudo cp wrk /usr/local/bin && cd ..
+  - "[ ! -f wrk/bin/wrk ] && rm -rf wrk && git clone https://github.com/wg/wrk.git && make -C wrk && mkdir wrk/bin && mv wrk/wrk wrk/bin || true"
```
PlasmaPower: This is because travis always creates the cached directory `wrk/bin` even if it doesn't have a cache.

jonathanong:
:+1: awesome!!!

PR #704, opened 2016-04-05
Rename request.ips to request.forwardedFor

PlasmaPower:
The name is somewhat deceiving IMO. Another option would be to always add `this.socket.remoteAddress` to the beginning of request.ips, and then have request.ip simply return `this.ips[0]`. However, that would be a breaking change, and this is not.

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -158,7 +158,7 @@ module.exports = class Application extends Emitter {
       keys: this.keys,
       secure: request.secure
     });
-    request.ip = request.ips[0] || req.socket.remoteAddress || '';
+    request.ip = request.forwardedFor[0] || req.socket.remoteAddress || '';
     context.accept = request.accept = accepts(req);
     context.state = {};
     return context;
--- a/lib/context.js
+++ b/lib/context.js
@@ -192,4 +192,5 @@ delegate(proto, 'request')
   .getter('stale')
   .getter('fresh')
   .getter('ips')
+  .getter('forwardedFor')
   .getter('ip');
--- a/lib/request.js
+++ b/lib/request.js
@@ -12,6 +12,7 @@ const qs = require('querystring');
 const typeis = require('type-is');
 const fresh = require('fresh');
 const only = require('only');
+const deprecate = require('depd')('koa');
 
 /**
  * Prototype.
@@ -378,12 +379,24 @@ module.exports = {
    * @api public
    */
 
-  get ips() {
+  get forwardedFor() {
     const proxy = this.app.proxy;
     const val = this.get('X-Forwarded-For');
     return proxy && val
       ? val.split(/\s*,\s*/)
       : [];
+  },
+
+  /**
+   * Deprecated alias of forwardedFor
+   * @return {Array}
+   * @api public
+   */
+
+  get ips() {
+    deprecate('This is a deprecated alias of forwardedFor. ' +
+              'Please use forwardedFor instead.');
+    return this.forwardedFor;
   },
 
   /**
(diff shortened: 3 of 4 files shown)


PlasmaPower:
Just a second, looks like I forgot to pull from v2.x before creating this branch and now it has conflicts.

PlasmaPower:
The new travis container builds seem to be working great! wrk was cached and didn't need to be rebuilt.

jonathanong:
not happy with `forwardedFor` because i believe you could forward more than just IPs (i.e. protocols). i'm also not the best person to ask about that.

PlasmaPower:
@jonathanong well there's the forwarded header, and the x-forwarded-header. This currently works off of the latter. A parser for the former would be much more complicated.

jonathanong:
oops yeah. i mean `.forwarded = []`. 

either way, this would be a breaking change (unless we just add an alias, which is kind of pointless), so i would hold on this.

PlasmaPower:
@jonathanong I did add a deprecated alias, and this is for v2.x.

jonathanong:
ok let's just get more feedback from people then. i'm +0

dougwilson:
I'm :-1: on this change for two reasons:

1. I think it's kinda ugly looking.
2. But mainly, the contents of `.forwardedFor` here are wrong. For example, the socket address should not be in `.forwardedFor`, since the requests was not forwarded for the actual server's socket, that doesn't make any sense. For this PR to make sense technically, it should only include the IPs which the request was forwarded for in the `.forwardedFor` property (i.e. the current address at `[0]` should not be there). This means that `.ip != .forwardedFor[0]` to make sense.

PlasmaPower:
@dougwilson .forwardedFor simply gets and parses the `X-Forwarded-For` header, that makes sense right? Is there something I'm missing about that header?

As for .ip, it gets it's value from forwardedFor, which makes sense because if the request is forwarded for a client the IP should be that of the client. Unless you're talking about req.remoteAddress, in which case I don't think that's the case in this PR.

dougwilson:
@PlasmaPower, I see that indeed `.ips` was just the parse of the header, so I take back my `#2` statement.

PlasmaPower:
@dougwilson Yeah I'd like to just prepend the actual IP to .ips and have .ip just be .ips[0], but unfortunately that would be a breaking change.

What do you mean about this being ugly looking? Are you talking about the property name? It might be a tad long, but it gets the point across and shouldn't be used often.

fl0w:
@PlasmaPower I prefer "the other solution" you mentioned in top issue post, it's breaking, but now is the perfect time to deprecate.

fengmk2:
-1, `request.ips` is more easy to understand and look nice. `ips` meaning IP list, and forwardedFor meaning more than IP address.

PlasmaPower:
What does everybody think about my other solution, adding `req.remoteAddress` to the end of .ips and having .ip simply be .ips[0]? Also, if we're caching .ip, should we be caching .ips?

fl0w:
@PlasmaPower I'm +1. But I think it should be to prepend remoteAddress, not append, right? Otherwise it'd be .ips[ips.length-1]?

PlasmaPower:
@fl0w, no the contacted proxy is the last in the chain of proxies, so it should be the last item in the list, and .ip should favor X-Forwarded-For which comes first in the list if app.proxy is on.

fl0w:
@PlasmaPower I stand corrected. Either way, +1. But that would mean deprecation in v2, and removal of old behaviour in v3? What's the impact on current codebase?

PlasmaPower:
@lf0w In this case, I'd be in favor of changing it in v2 without deprecation because .ips is a very nice name. In addition, it shouldn't be used by the end user much (if at all) and I think overall the breakage should be small.

PR #707, opened 2016-04-12
Added onUsed and wrapper support

PlasmaPower:
Yet another wrapper implementation. Created this to play with the idea. Requires koajs/compose#53 to work.

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -21,8 +21,6 @@ const assert = require('assert');
 const Stream = require('stream');
 const http = require('http');
 const only = require('only');
-const convert = require('koa-convert');
-const deprecate = require('depd')('koa');
 
 /**
  * Expose `Application` class.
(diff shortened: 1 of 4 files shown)


jonathanong:
what's the pros and cons with https://github.com/koajs/koa/pull/694? seems quite convoluted.

@koajs anyone have opinions on this or https://github.com/koajs/koa/pull/694?

PlasmaPower:
@jonathanong Pros:

- It will be a lot easier for profilers to generate a tree of middleware
- Middleware can be converted before any request arrives, speeding things up
- The `onUsed` API is more general than just wrapping, and could be used for other things

Cons:

- The most complicated option

PlasmaPower:
@jonathanong I added tests for nested middleware, look good?

PlasmaPower:
I've rebased this to v2.x, but actual server tests will still fail until koajs/koa-compose#53 is merged.

jonathanong:
would really love more people to express their opinion :)

miwnwski:
I'll just close this as it has little relevance today IMO. Maybe there's still a need for wrappers but I doubt it personally and the same functionality can be approached in user land. Also, over the years I haven't really seen any requests for wrappers. If I remember correctly, this was brought to light due to TJ's original issue to simplify mw profiling, though today this can be solved with async_hooks which seems more like native approach Koa should aim towards. Please reopen if I'm wrong :)

PR #711, opened 2016-04-19
fix: v1 artifact "this" reference

cr1s:
As above


fl0w:
LGTM

cr1s:
There a couple of other this references in the v.2x docs, that I missed on the 1st pass.

PR #723, opened 2016-04-29
fix to pass the test at node v6.x

iyuq:
I reopened the request and squashed the 3 commits into one commit.

### DIFF
--- a/test/request/query.js
+++ b/test/request/query.js
@@ -2,12 +2,13 @@
 'use strict';
 
 const context = require('../helpers/context');
+const should = require('should');
 
 describe('ctx.query', () => {
   describe('when missing', () => {
     it('should return an empty object', () => {
       const ctx = context({ url: '/' });
-      ctx.query.should.eql({});
+      should(ctx.query).empty;
     });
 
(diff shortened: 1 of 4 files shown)


juliangruber:
can you explain why this change is necessary for node 6? (btw +1 from me, I've never liked the implicit globals available through mocha)

iyuq:
@juliangruber  because node v6.x's querystring is create by Object.create(null), it doesn't have the prototype of Object.prototype, so if call ctx.query.should it would throw a error.

iyuq:
@juliangruber  see The object returned by querystring.parse() no longer inherits from Object.prototype [#6055](https://github.com/nodejs/node/pull/6055).

juliangruber:
thank you!

iyuq:
@juliangruber you're welcome! but why haven't it added to the master branch? I have also opened a pull request to the master branch.

PlasmaPower:
@iyuq I'm pretty sure that PR got closed as a duplicate of this one by accident :P it shouldn't be though, they are different branches.

iyuq:
@juliangruber  so I should reopen that PR again?

PR #726, opened 2016-05-03
remove the local should and change to a better way

iyuq:
Since I don't like the local should either, I think this is better to use this to avoid local should. It can also pass test on Node v6.0.0. If this make sence, I will also open another PR to the v2.x branch.

### DIFF
--- a/test/request/querystring.js
+++ b/test/request/querystring.js
@@ -3,7 +3,6 @@
 
 var context = require('../context');
 var parseurl = require('parseurl');
-var should = require('should');
 
 describe('ctx.querystring', function(){
   it('should return the querystring', function(){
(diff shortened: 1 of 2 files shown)


test/request/querystring.js
```diff
@@ -33,8 +32,9 @@ describe('ctx.querystring=', function(){
     ctx.querystring = 'page=2&color=blue';
     ctx.url.should.equal('/store/shoes?page=2&color=blue');
     ctx.search.should.equal('?page=2&color=blue');
-    should(ctx.query).have.property('page', '2');
-    should(ctx.query).have.property('color', 'blue');
+    Object.keys(ctx.query).should.eql(['page', 'color']);
```
PlasmaPower: Why this assert? The next ones will throw an error if the keys don't exist anyway.

iyuq:
unless the keys is set enumerable:false, it will be in the Object.keys. But I'm not sure the order of the two keys, it seems V8 order them in the order they define. I check this to ensure the query object doesn't contain any other key. @PlasmaPower

jonathanong:
i don't see the benefit of moving from a local `should` to a global `should`. i'd actually prefer if we only used local `should`. even more, i'd prefer if we ditched `should` in favor of node.js' built-in `assert`.

anyone else have opinions?

jonathanong:
also, this should be against `v2.x` as that will be maintained longer

tj:
node's assert is brutal, that's my 0.02 haha, I like tests you can actually read, but most importantly assert's error messages are really poor

jonathanong:
only going to merge bug fixes to v1.x now. thanks!

PR #733, opened 2016-05-16
redirect should add mountPath prefix

coderhaoxin:
Maybe it's not a good idea to do this in koa, any thought?

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -251,7 +251,9 @@ module.exports = {
 
   redirect: function(url, alt){
     // location
+    const mountPath = this.ctx.mountPath;
     if ('back' == url) url = this.ctx.get('Referrer') || alt || '/';
+    if (mountPath && url.indexOf('/') == 0) url = mountPath + url;
     this.set('Location', url);
 
     // status
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
   "devDependencies": {
     "babel": "^5.0.0",
     "istanbul": "^0.4.0",
+    "koa-mount": "^1.3.0",
     "make-lint": "^1.0.1",
     "mocha": "^2.0.1",
     "should": "^6.0.3",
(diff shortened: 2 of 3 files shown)


lib/response.js
```diff
@@ -251,7 +251,9 @@ module.exports = {
 
   redirect: function(url, alt){
     // location
+    const mountPath = this.ctx.mountPath;
```
dead-horse: -1.

override `redirect` in `koa-mount` is much better than add this logic here, or write a `koa-mount-redirect` for this feature alone.

juliangruber:
-1 from me as well, at least for this implementation. Does anyone know how express handles this?

I'm thinking either koa-mount should patch .redirect, or koa should offer a more generic way to handle this without patching.

juliangruber:
actually I'm also -1 to this logic, if I do `.redirect('/foo')` i expect it to redirect to `DOMAIN/foo`, instead of `DOMAIN/prefix/foo`. That's what absolute paths should do. Otherwise, use relative paths.

juliangruber:
I'd suggest adding a method like `.subRedirect(path)` via your middleware that contains this logic.

dougwilson:
> Does anyone know how express handles this?

Express does pretty much the same as Koa, both just passing through whatever URL was provided.

coderhaoxin:
> override redirect in koa-mount is much better than add this logic here, or write a koa-mount-redirect for this feature alone.

It's not a good idea to override method in middlewares IMO, also not worth to write a new module.

> I'd suggest adding a method like .subRedirect(path) via your middleware that contains this logic.

+1 for this!

coderhaoxin:
> I'd suggest adding a method like .subRedirect(path) via your middleware that contains this logic.

Actually, I think relative path is enough. https://github.com/koajs/mount/issues/45

peey:
I don't think this should be done in koa. 

Overriding in koa-mount would be a better solution and more in line with the behavior on how it already overrides the original path

PR #736, opened 2016-05-19
fix typo

### LINKED ISSUE #735: typo: on koajs.com website main page
Under "Error Handling":

"If an error in the req/res cycle and..."

should read:
"If an error is in the req/res cycle and..."

SherylHohman:
fixes #735


dead-horse:
see https://github.com/koajs/koa/pull/738#issuecomment-220766234

PR #738, opened 2016-05-19
possible typo

### LINKED ISSUE #737: possible typo..
Documentation, at http://koajs.com/#response

**should:**
response.socket  
Request socket.

**read:**
response.socket  
Response socket.

**?**
Or is it correct as is ?

SherylHohman:
Is documentation correct? 
or should it be: ?

from:
response.socket

Request socket.

to:
response.socket

Response socket.

if so:
fixes #737

### DIFF
--- a/docs/api/response.md
+++ b/docs/api/response.md
@@ -17,7 +17,7 @@
 
 ### response.socket
 
-  Request socket.
+  Response socket.
 
 ### response.status
 


dead-horse:
in each http request we only have one `socket`: https://nodejs.org/dist/latest-v4.x/docs/api/http.html#http_message_socket

so in koa, both `request.socket` and `response.socket` are point to it.

PR #743, opened 2016-05-26
add maxAge and overwrite options api/context doc to master branch.

iyuq:
add maxAge and overwrite options to ctx.cookies.set(name, value, [options]).


iyuq:
@dead-horse  I add this to master also.

PR #744, opened 2016-05-26
2.x change three console.error statements into one.

iyuq:
1. change three console.error statements into one, this should improve some performance for the console.error is a sync io operation, reduce the number of times to call to console.error should improve some performance although this will not cause realtime problem.
2. user template string in multiple lines of strings to avoid + operator.

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -105,9 +105,9 @@ module.exports = class Application extends Emitter {
   use(fn) {
     if (typeof fn !== 'function') throw new TypeError('middleware must be a function!');
     if (isGeneratorFunction(fn)) {
-      deprecate('Support for generators will been removed in v3. ' +
-                'See the documentation for examples of how to convert old middleware ' +
-                'https://github.com/koajs/koa/tree/v2.x#old-signature-middleware-v1x');
+      deprecate(`Support for generators will been removed in v3. 
+                See the documentation for examples of how to convert old middleware 
+                https://github.com/koajs/koa/tree/v2.x#old-signature-middleware-v1x`);
       fn = convert(fn);
     }
     debug('use %s', fn._name || fn.name || '-');
--- a/lib/application.js
+++ b/lib/application.js
@@ -178,9 +178,7 @@ module.exports = class Application extends Emitter {
     if (this.silent) return;
 
     const msg = err.stack || err.toString();
-    console.error();
-    console.error(msg.replace(/^/gm, '  '));
-    console.error();
+    console.error(`\n${msg.replace(/^/gm, '  ')}\n`);
   }
 
 };
--- a/test/application/onerror.js
+++ b/test/application/onerror.js
@@ -48,7 +48,7 @@ describe('app.onerror(err)', () => {
 
     const output = stderr.inspectSync(() => app.onerror(err));
 
-    output.should.eql(['\n', '  Foo\n', '\n']);
+    output.should.eql(['\n  Foo\n\n']);
 
     done();
   });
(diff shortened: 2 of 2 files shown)


juliangruber:
the decprecation notice now should look weird, right? with all the extra white space. have you tried causing it?

juliangruber:
also, are you sure those 2 extra console log statements cause a real performance penalty for you?

iyuq:
@juliangruber the deprecation notice is weird. So the second commit is wrong. The two extra console log statements will not cause a real performance penalty, but I'm pretty sure that merge them into one will  
improve the performance. It will not cause any read problem. Should it be merged?

lib/application.js
```diff
@@ -178,9 +178,7 @@ module.exports = class Application extends Emitter {
     if (this.silent) return;
 
     const msg = err.stack || err.toString();
-    console.error();
-    console.error(msg.replace(/^/gm, '  '));
-    console.error();
+    console.error(`\n${msg.replace(/^/gm, '  ')}\n`);
```
jonathanong: i don't really see this as cleaner

lib/application.js
```diff
@@ -105,9 +105,9 @@ module.exports = class Application extends Emitter {
   use(fn) {
     if (typeof fn !== 'function') throw new TypeError('middleware must be a function!');
     if (isGeneratorFunction(fn)) {
-      deprecate('Support for generators will been removed in v3. ' +
-                'See the documentation for examples of how to convert old middleware ' +
-                'https://github.com/koajs/koa/tree/v2.x#old-signature-middleware-v1x');
+      deprecate(`Support for generators will been removed in v3.
```
jonathanong: the intention before was that there was no indentation or line break before. now there is. do you guys think this is uglier?

test/application/onerror.js
```diff
@@ -48,7 +48,7 @@ describe('app.onerror(err)', () => {
 
     const output = stderr.inspectSync(() => app.onerror(err));
 
-    output.should.eql(['\n', '  Foo\n', '\n']);
+    output.should.eql(['\n  Foo\n\n']);
```
jonathanong: this looks cleaner to me tho

PlasmaPower: It's only due to the (much less) clean change above. However, my comment on it may be a good compromise.

lib/application.js
PlasmaPower: How about `console.error('\n' + msg.replace(/^/gm, '  ') + '\n')`? I think it's a nice compromise.

lib/application.js
PlasmaPower: I prefer the style before. I don't think the indentation looks good since the first line isn't indented.

PlasmaPower:
Also, doing this for performance is a fine example of premature optimization.

lib/application.js
iyuq: I can accept this way, it seems to be more readable.

test/application/onerror.js
iyuq: @juliangruber Yes, with the change above, this change will be need. But if without that, this will fail the test.

lib/application.js
iyuq: @jonathanong yes, this is a mistake. I found it also.

iyuq:
@jonathanong how about changing to  `console.error('\n', ${msg.replace(/^/gm, '  ')}, '\n');`, it seems more readable.If this is okay, I can open a new pull request to only pull this change.

lib/application.js
dougmoscrop: `console.error('\n', msg.replace(/^/gm, ' '), '\n')` is another option

jonathanong:
this has been inactive for a while. please feel free to open a new PR with recommended changes. thanks!

PR #749, opened 2016-06-01
http.createServer -> https.create... to illustrate "multiple addresses" comment

jonboiser:
Just saw this and assumed the intent was to use node's `https` module for the second listener.


jonathanong:
cool. can you add an explanation of `options`? people might get confused because it's not defined.

jonboiser:
Add some clarification.

PR #751, opened 2016-06-06
swap co to bluebird-co for better performance

capaj:
here are my benchmark results before the swap:

```
make[1]: Entering directory '/home/capaj/git_projects/koa/benchmarks'

  1 middleware
  12657.20

  5 middleware
  11990.98

  10 middleware
  12123.40

  15 middleware
  11384.56

  20 middleware
  11351.09

  30 middleware
  10703.54

  50 middleware
  9655.15

  100 middleware
  7821.55


  1 async middleware
Experimental ES7 Async Function support is deprecated. Please look into Koa v2 as the middleware signature has changed.
  10373.59

  5 async middleware
Experimental ES7 Async Function support is deprecated. Please look into Koa v2 as the middleware signature has changed.
  8331.53

  10 async middleware
Experimental ES7 Async Function support is deprecated. Please look into Koa v2 as the middleware signature has changed.
  6510.46

  15 async middleware
Experimental ES7 Async Function support is deprecated. Please look into Koa v2 as the middleware signature has changed.
  5154.40

  20 async middleware
Experimental ES7 Async Function support is deprecated. Please look into Koa v2 as the middleware signature has changed.
  4722.30

  30 async middleware
Experimental ES7 Async Function support is deprecated. Please look into Koa v2 as the middleware signature has changed.
  3623.42

  50 async middleware
Experimental ES7 Async Function support is deprecated. Please look into Koa v2 as the middleware signature has changed.
  2465.00

  100 async middleware
Experimental ES7 Async Function support is deprecated. Please look into Koa v2 as the middleware signature has changed.
  1268.31

```

and after the swap:
```
make[1]: Entering directory '/home/capaj/git_projects/koa/benchmarks'

  1 middleware
  18730.10

  5 middleware
  17620.69

  10 middleware
  16914.27

  15 middleware
  16391.59

  20 middleware
  15858.33

  30 middleware
  14465.84

  50 middleware
  13257.24

  100 middleware
  9867.00


  1 async middleware
Experimental ES7 Async Function support is deprecated. Please look into Koa v2 as the middleware signature has changed.
  10567.71

  5 async middleware
Experimental ES7 Async Function support is deprecated. Please look into Koa v2 as the middleware signature has changed.
  8454.45

  10 async middleware
Experimental ES7 Async Function support is deprecated. Please look into Koa v2 as the middleware signature has changed.
  6357.48

  15 async middleware
Experimental ES7 Async Function support is deprecated. Please look into Koa v2 as the middleware signature has changed.
  5422.08

  20 async middleware
Experimental ES7 Async Function support is deprecated. Please look into Koa v2 as the middleware signature has changed.
  4656.78

  30 async middleware
Experimental ES7 Async Function support is deprecated. Please look into Koa v2 as the middleware signature has changed.
  3561.25

  50 async middleware
Experimental ES7 Async Function support is deprecated. Please look into Koa v2 as the middleware signature has changed.
  2516.37

  100 async middleware
Experimental ES7 Async Function support is deprecated. Please look into Koa v2 as the middleware signature has changed.
  1311.89

make[1]: Leaving directory '/home/capaj/git_projects/koa/benchmarks'

```

Thanks for your consideration!

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -21,7 +21,7 @@ var assert = require('assert');
 var Stream = require('stream');
 var http = require('http');
 var only = require('only');
-var co = require('co');
+var co = require('bluebird-co');
 
 /**
  * Application prototype.
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
   "license": "MIT",
   "dependencies": {
     "accepts": "^1.2.2",
+    "bluebird": "^3.4.0",
     "bluebird-co": "^2.1.2",
     "composition": "^2.1.1",
     "content-disposition": "~0.5.0",
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
   "license": "MIT",
   "dependencies": {
     "accepts": "^1.2.2",
-    "co": "^4.4.0",
+    "bluebird-co": "^2.1.2",
     "composition": "^2.1.1",
     "content-disposition": "~0.5.0",
     "content-type": "^1.0.0",


PlasmaPower:
Seems to be a considerable improvement - :+1:

capaj:
@PlasmaPower yeah, the simple case with just one middleware is faster by 47 percent. Not bad for 3 line change :grinning:

novacrazy:
The only negative to putting bluebird-co into as largely used library as this is that it kind of pollutes the bluebird coroutine handlers. Plus requiring bluebird in the first place. 

However, the performance improvements being as good as they are, a solution would be to do something like allowing `require('koa/native')` or `require('koa/legacy')` for anyone that is insistent on using the native Promise implementation.

I do like native Promises, but they're still not as good as Bluebird for now. Supposedly they'll get much better in future versions of JavaScript, though. 

Just thought I'd give my thoughts on this. I would love to see bluebird-co in Koa, but those concerns might annoy people.

capaj:
@novacrazy maybe koa could use https://github.com/kevinbeaty/any-promise approach where by default it would have `bluebird-co` and you could just opt to supply any `co` implementation you'd need.

PlasmaPower:
@capaj No such package exists that I know of though, I guess we could create `any-co` :P. I'm not sure how useful that would be though over a simple require toggle.

tj:
Friendly reminder that almost no one needs even 300,000 requests a minute haha. I'm all down for perf improvements but bluebird is behemoth, not a fan. I'm sure we could just optimize Co or similar.

PlasmaPower:
@tj But is it web scale? /s

Seriously though, I do think we should be concerned with the performance improvement this brings, it probably represents a 40% decrease in needed processing power, which means that this update would save about 40% on servers (it's not linear, I know, but you get the point).

tj:
Totally, but without the baggage of a giant library would be definitely preferable IMO

PlasmaPower:
@tj Makes sense. Maybe we could check add bluebird-co as an optional dependency.

novacrazy:
Well the reason I made bluebird-co is because I was "parallelizing" (making it use async and streams wherever possible) hundreds if not thousands of things, on a website and service I was writing. That's where the performance improvements (and even JIT and cache hit improvements) of Bluebird or bluebird-co came into play. I was able to reduce TTFB times on the website by about 5-30% (quite a variance just depending on the problem complexity) just through that.

I kind of tried to express that in the stupid "complex_generator" benchmark, where Bluebird and bluebird-co really shine through.

tj:
Maybe there's a version of Bluebird that cuts out 95% of the unnecessary stuff haha, can't say I've looked

novacrazy:
Another option is to do the opposite of the require native or legacy thing. Make bluebird-co and bluebird optional or peer dependencies and require a specific version that uses them instead.

However, it'd be hard for people to know that has been added unless it's pretty up-front in the README or docs.

capaj:
@tj I think a lot of projects already use bluebird-mine certainly are. I do agree that it's a big dependency. I am sure that you could just cut all the non standard API and get much more lightweight `bluebird`. That would probably be the best since bluebird is the fastest promise lib.
`bluebird-light` wouldn't be a bad name.

magicdawn:
Aren't we going to remove generator support in the comming soon when async/await become native ?

tj:
@magicdawn that's the plan yep.

It would be cool to have some latency benchmarks, that's definitely more important than an already absurdly large rps metric (for 99% of the cases at least)

magicdawn:
@tj Gotcha

BTW cann't find a `bluebird-light` or `bluebird-lite` 😂

dead-horse:
most web server's in production is much more complicated than 'Hello World', we are running 20+ application's(each have a cluster with 10~100 servers), most of these server's maximal qps is below than 1000(lots of stuffs limit the performance). Improve performance in this level is not the most important thing for now, koa now is fast enough.

BTW, anybody test koa with co and bluebird's promise implement? If native Promise is much slower than bluebird, people who want to use bluebird can just use bluebird as global promise implementation by `global.Promise = require('bluebird')`. and in the and we'll migrate to async await when it being native(it may be much slower than generator in the first implementation).

capaj:
@dead-horse good point- I benched with regular co and got even better results:

```
make[1]: Entering directory '/home/capaj/git_projects/koa/benchmarks'

  1 middleware
  17925.78

  5 middleware
  17100.42

  10 middleware
  17008.89

  15 middleware
  16345.61

  20 middleware
  15055.35

  30 middleware
  14195.20

  50 middleware
  12745.32

  100 middleware
  9936.54


  1 async middleware
Experimental ES7 Async Function support is deprecated. Please look into Koa v2 as the middleware signature has changed.
  16802.68

  5 async middleware
Experimental ES7 Async Function support is deprecated. Please look into Koa v2 as the middleware signature has changed.
  13556.76

  10 async middleware
Experimental ES7 Async Function support is deprecated. Please look into Koa v2 as the middleware signature has changed.
  11220.10

  15 async middleware
Experimental ES7 Async Function support is deprecated. Please look into Koa v2 as the middleware signature has changed.
  8990.07

  20 async middleware
Experimental ES7 Async Function support is deprecated. Please look into Koa v2 as the middleware signature has changed.
  7863.99

  30 async middleware
Experimental ES7 Async Function support is deprecated. Please look into Koa v2 as the middleware signature has changed.
  6106.40

  50 async middleware
Experimental ES7 Async Function support is deprecated. Please look into Koa v2 as the middleware signature has changed.
  4035.58

  100 async middleware
Experimental ES7 Async Function support is deprecated. Please look into Koa v2 as the middleware signature has changed.
  2150.38

make[1]: Leaving directory '/home/capaj/git_projects/koa/benchmarks'

```

It is obvious that just monkey patching Promise to bluebird's promise is better way to achieve best performance. Closing this PR.

tj:
Weird! 

We could just chuck it on the site for now if anyway wants/needs the boost

capaj:
@tj see https://github.com/koajs/koa/pull/753
Would that be a good place to put it?

joshburgess:
This is super interesting. Many people are already using Bluebird globally for promises anyway. 

Would love to see a clustered & Bluebird-ified koa benchmark added to this repo for comparison purposes: 

[https://github.com/mroth/phoenix-showdown](https://github.com/mroth/phoenix-showdown)

fl0w:
Isn't there an `any-co`-library that does what `any-promise` does? Maybe that's an option for now and the future?

andyhu:
Here is the npm package I've published in case anyone who wish to use koa with bluebird-co:
http://npmjs.com/package/koa-bluebird-co

PlasmaPower:
@andyhu doing `global.Promise = require('bluebird')` before requiring Koa is faster than that BTW. Also, you forgot to update the git link in your `package.json` for that (and Readme needs updating of course).

novacrazy:
The reason `bluebird-co` might appear slower than `co` plus `bluebird` is because `bluebird-co` uses Bluebird's `Promise.coroutine` as the base, whereas `co` does that from scratch. Bluebird does a lot more work with stack traces, promise cancellation, and other stuff.

All I wanted with `bluebird-co` was to add similarly behaving and highly optimized yield handlers via `Bluebird.coroutine.addYieldHandler` to do what `co` provides, and provide a small interface to further extend it. That last part being key.

Like TJ said, so little time is spent on either solution in practice that it doesn't usually matter.

PR #753, opened 2016-06-06
mention bluebird promise performance boost

capaj:


PlasmaPower:
If this gets merged, you might also want to create this for v2.x, where it will have an even bigger impact.

jonathanong:
iirc correctly v2 uses `any-promise`. i'm not really willing to recommend people replace globals...

fl0w:
Also, this won't work for anyone transpiling using babel which would require `require('babel-runtime/core-js/promise').default = require('bluebird')`.

tj:
nothing wrong with replacing globals really, it's effectively the same thing as a replacing a singleton

eriktrom:
Seems reasonable until native implementations are faster. It's also a good doc in general, I say merge it. If anyone is opposed to globals, just say "yeah we know it's a global monkeypatch, and x,y and z are the reasons..."

I hate when I forget to require bluebird, just in general.  Bigger than koa this does seems to be, but a good place to mention the topic it is nonetheless (as koa is popular and people may actually read these docs, spread word, etc)

eriktrom:
> If this gets merged, you might also want to create this for v2.x, where it will have an even bigger impact.

Disagree. Very few people read .md files outside of master branch on github.

fl0w:
@eriktrom: @PlasmaPower was talking about performance wise - as v2.x uses Promises for flow control.

jonathanong:
no longer relevant with v2.x. thanks

capaj:
@jonathanong I disagree that this is not relevant anymore. There has been some improvement in native promises speed, but bluebird is still way faster even when you're not using generators.

tj:
When someone shows me that they handle over 9000 or whatever request/s on a single node which somehow magically provides HA, we can talk about adding Bluebird.

capaj:
@tj I am not saying promises are the bottleneck. You're right, the performance gains are small and it's worthless to waste precious readme.md markdown space on this.

My disagreement was merely with a comment https://github.com/koajs/koa/pull/753#issuecomment-282463545

PR #754, opened 2016-06-09
Updated AUTHORS

PlasmaPower:
It's now alphabetically sorted, with info at the top on how to update it. Should we only keep names with spaces in here? (as in, not usernames, just real names) It would leave out some contributors, but we already have the contributors button at the top of the repo; this serves a different purpose.

Also, we have different combinations of usernames, emails, and real names, but there's no good way to unique them independently I think.

I'll make this for v2.x once it's merged for master.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,70 +1,95 @@
-TJ Holowaychuk <tj@vision-media.ca>
-Jonathan Ong <jonathanrichardong@gmail.com>
-dead_horse <dead_horse@qq.com>
-Yiyu He <dead_horse@qq.com>
-fengmk2 <fengmk2@gmail.com>
-Julian Gruber <julian@juliangruber.com>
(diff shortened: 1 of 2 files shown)


PlasmaPower:
Should I make this into a shell script? Maybe have Travis check that it's correct? A pre-commit hook would be optimal, but unfortunately we can't ship those with the repo of course.

Maybe we could use a GitHub webhook? Might not be a bad idea - whenever we get a commit on master, pull, regen authors, add, commit (it won't do anything if there's nothing to commit), and push.

juliangruber:
you could move the update script into the package.json's scripts section

PlasmaPower:
@juliangruber Good idea! I've done that. I was looking for a good place to put it, a bin folder didn't seem necessary.

juliangruber:
LGTM now :)

PR #760, opened 2016-06-23
acceptsLanguages method fails to detect locale

marr:
Illustrates that the arguments received during an `acceptLanguages(langs)` call does not appear to behave as it should.


jonathanong:
hi, can you open this issue here? https://github.com/jshttp/negotiator

marr:
@jonathanong done https://github.com/jshttp/negotiator/pull/46

PR #776, opened 2016-07-14
Replace co with bluebird-co

andyhu:
By replacing `co` with `bliebird-co`, the benchmark is nearly doubled!
See result below:

### Benchmarks with koa 1.2.0:
image

### Benchmarks with koa 1.2.0 (modified version, replaced co with bluebird-co)
image

### Benchmarks with koa 2.0.0-alpha3:
image

Here is the benchmark for `co` vs `bluebird` `coroutine`:
https://github.com/novacrazy/bluebird-co/tree/master/benchmark

This is a huge gain.

Here is the npm package I published with this pr:
http://npmjs.com/package/koa-bluebird-co

### DIFF
--- a/benchmarks/experimental/async.js
+++ b/benchmarks/experimental/async.js
@@ -2,6 +2,7 @@
 'use strict';
 
 var http = require('http');
+var Promise = require('bluebird');
 var koa = require('../..');
 var app = koa();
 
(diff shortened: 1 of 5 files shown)


andyhu:
Sorry this might be duplicated with #751

PlasmaPower:
Yep! If you look into it, `co` is actually faster, but Bluebird promises gave the performance increases.

PR #778, opened 2016-07-18
remove last comma in object

sodawy:


PlasmaPower:
Should we add an eslint rule to enforce this?

magicdawn:
eslint +1

jonathanong:
yeah, style changes are pointless w/o rules to enforce it

PlasmaPower:
Oh, master doesn't have an eslintrc? I thought it did. This is already fixed in v2.x BTW.

PR #781, opened 2016-07-21
fix(response): correct response.writable logic

### LINKED ISSUE #779: pipeling request will lost!
### Test result:
```
=== Http server ===
Request count: 1 , Error listeners: 1
Request count: 2 , Error listeners: 1
HTTP/1.1 200 OK
Date: Wed, 20 Jul 2016 02:56:47 GMT
Connection: keep-alive
Content-Length: 1

1HTTP/1.1 200 OK
Date: Wed, 20 Jul 2016 02:56:47 GMT
Connection: keep-alive
Content-Length: 1

2

=== Toa server ===
Request count: 1 , Error listeners: 1
Request count: 2 , Error listeners: 1
HTTP/1.1 200 OK
Content-Type: text/plain; charset=utf-8
Content-Length: 1
X-Powered-By: Toa
Date: Wed, 20 Jul 2016 02:56:47 GMT
Connection: keep-alive

1HTTP/1.1 200 OK
Content-Type: text/plain; charset=utf-8
Content-Length: 1
X-Powered-By: Toa
Date: Wed, 20 Jul 2016 02:56:47 GMT
Connection: keep-alive

2

=== Koa server ===
Request count: 1 , Error listeners: 2
Request count: 2 , Error listeners: 2
HTTP/1.1 200 OK
Content-Type: text/plain; charset=utf-8
Content-Length: 1
Date: Wed, 20 Jul 2016 02:56:47 GMT
Connection: keep-alive

1
```

### Reason

When server emit a `request`, `responce` instance may not have `socket` for the time being, [_http_server.js](https://github.com/nodejs/node/blob/master/lib/_http_server.js#L487), `this.writable === false` in [application.js](https://github.com/koajs/koa/blob/master/lib/application.js#L199)

dead-horse:
closes #779

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -478,16 +478,19 @@ module.exports = {
 
   /**
    * Checks if the request is writable.
-   * Tests for the existence of the socket
-   * as node sometimes does not set it.
    *
    * @return {Boolean}
    * @api private
    */
 
   get writable() {
+    // can't write any more after response finished
+    if (this.res.finished) return false;
+
     var socket = this.res.socket;
-    if (!socket) return false;
+    // There are already pending outgoing res, but still writable
+    // https://github.com/nodejs/node/blob/v4.4.7/lib/_http_server.js#L486
+    if (!socket) return true;
     return socket.writable;
   },
 
--- a/test/application.js
+++ b/test/application.js
@@ -6,6 +6,7 @@ var request = require('supertest');
 var statuses = require('statuses');
 var assert = require('assert');
 var http = require('http');
+var net = require('net');
 var koa = require('..');
 var fs = require('fs');
 var AssertionError = assert.AssertionError;
--- a/test/response/writeable.js
+++ b/test/response/writeable.js
@@ -1,19 +1,105 @@
 
 'use strict';
 
-var response = require('../context').response;
+var should = require('should');
+var koa = require('../../');
+var net = require('net');
 
 describe('res.writable', function(){
-  it('should return the request is writable', function(){
-    var res = response();
-    res.writable.should.be.ok;
+  describe('when continuous requests in one persistent connection', function() {
+    function requestTwice(server, done) {
+      let port = server.address().port;
+      var buf = new Buffer('GET / HTTP/1.1\r\nHost: localhost:' + port + '\r\nConnection: keep-alive\r\n\r\n');
+      var client = net.connect(port);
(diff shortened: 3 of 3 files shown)


lib/response.js
```diff
@@ -478,16 +478,19 @@ module.exports = {
...
    * Checks if the request is writable.
-   * Tests for the existence of the socket
-   * as node sometimes does not set it.
    *
    * @return {Boolean}
    * @api private
    */
 
   get writable() {
+    // can't write any more after response finished
```
juliangruber: this looks like it could be a handy little node module

dead-horse: haha, I think support by node core is more reasonable.

https://github.com/nodejs/node/blob/master/lib/_http_outgoing.js#L61 `response.writable` is never changed.

fengmk2:
LGTM

lib/response.js
```diff
@@ -478,16 +478,19 @@ module.exports = {
...
+    if (this.res.finished) return false;
+
     var socket = this.res.socket;
-    if (!socket) return false;
+    // There are already pending outgoing res, but still writable
+    // https://github.com/nodejs/node/blob/v4.4.7/lib/_http_server.js#L486
+    if (!socket) return true;
```
jonathanong: interesting. if the socket doesn't exist, couldn't it also be that the connection was closed?

jonathanong:
don't 100% understand it but writing to a broken connection shouldn't raise any issues, so it's ok to me

lib/response.js
dead-horse: socket will always assigned to res (https://github.com/nodejs/node/blob/v4.4.7/lib/_http_server.js#L517). even it is closed,(but socket.writable will set to false).

dead-horse: wait. not sure if [this](https://github.com/nodejs/node/blob/v4.4.7/lib/_http_server.js#L433) will make res.socket not exists, but this too extreme anyway.

lib/response.js
dougwilson: In case anyone is wondering, this change seems fine to me, and of course checking the flag really doesn't quality to have it's own module. FWIW, the following I found to work on all Node.js versions as far back as 0.8: https://github.com/jshttp/on-finished/blob/v2.3.0/index.js#L70

It is literally the same code that is in here now, but as a single line :)

PR #782, opened 2016-07-22
fix(response): correct response.writable logic

dead-horse:
related #781

### DIFF
--- a/test/response/writeable.js
+++ b/test/response/writeable.js
@@ -1,19 +1,94 @@
 
 'use strict';
 
-const response = require('../helpers/context').response;
+const Koa = require('../../');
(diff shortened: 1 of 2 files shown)


test/response/writeable.js
```diff
@@ -1,19 +1,95 @@
...
+          if (ctx.writable) return done(new Error('ctx.writable should not be true'));
+          done();
+        });
+      });
+      const server = app.listen();
+      requsetClosed(server);
+    })
+  })
+
+  describe('when resposne finished', () => {
```
fl0w: typo `response`

dead-horse: oops

PR #792, opened 2016-08-03
Use triple equals

ratson:


fundon:
Should include this https://github.com/koajs/koa/blob/v2.x/.eslintrc.yml#L14

jonathanong:
style changes aren't accepted with rules to enforce them. you'll also have to go through the gauntlet of all contributors to judge whether such a change is worth it :)

tj:
-1 :D haha `typeof` is always a string blah blah usual speech

PlasmaPower:
I'd like this change, if only for the fact that we could enforce it with eslint.

dominhhai:
:-1:  for this. `typeof` may be a `String`.
Force JSers to use static type is not good idea, thought.

fundon:
👍 for this.

Why not?

* Express.js does it. https://github.com/expressjs/express/blob/master/lib/request.js#L64

* Lints or Styles recommend it. https://github.com/airbnb/javascript#comparison-operators--equality

tj:
I counter the "why not?" with a "why?" haha. I won't bother bikeshedding, but since we're not going to bikeshed there's no point to change it IMO, it doesn't harm anything but pull-requests like this are kind of just noise

fundon:
@tj Thank you for your detailed explanation.

PR #795, opened 2016-08-10
fix: res.flushHeaders()

### LINKED ISSUE #788: Clarify documentation how to flush headers early using koa
We tried all possible google links about "koa flush headers early" and only could found two issues about "flushHeaders" function. 

We tried that function but it doesn't work as expected. The expected behavior is:
1) call flushHeaders()
2) all headers that were set are sent over the internet to the client
3) if the server tries to set other headers it throws
4) when the body is set, it's streamed to the client

A basic use-case is to stream "preload" headers for alerting the client to download CSS and font files, so they will be ready by the time slow api and db calls are finished. 

For some reason we were able to do this with express and can't figure out how to do it with Koa. Probably just missing documentation, it will help a lot! Thanks.

jonathanong:
closes #788 

a better way to test this would be nice...

/cc @mufumbo @PlasmaPower @renatopeterman

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -534,6 +534,6 @@ module.exports = {
    * Flush any set headers, and begin the body
    */
   flushHeaders() {
-    this.res.writeHead(this.res.statusCode);
+    this.res.flushHeaders(this.res.statusCode);
   }
 };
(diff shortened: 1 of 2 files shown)


renatopeterman:
Hi @jonathanong! 
Thanks for the fix, it's working now!

renatopeterman:
Is it ready to be merged?

jonathanong:
great. yup, after i fix tests for node < 6

PR #804, opened 2016-08-30
fix: add named function for request and response handlers

fengmk2:
Named function is good for APM and avoid anonymous functions everywhere.

image

image

PR #805, opened 2016-08-31
fix: add named arrow function for request and response handlers

fengmk2:
cherry-pick #804

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -128,13 +128,16 @@ module.exports = class Application extends Emitter {
 
     if (!this.listeners('error').length) this.on('error', this.onerror);
 
-    return (req, res) => {
+    const handleRequest = (req, res) => {
       res.statusCode = 404;
       const ctx = this.createContext(req, res);
       const onerror = err => ctx.onerror(err);
+      const handleResponse = () => respond(ctx);
       onFinished(res, onerror);
-      fn(ctx).then(() => respond(ctx)).catch(onerror);
+      fn(ctx).then(handleResponse).catch(onerror);
     };
+
+    return handleRequest;
   }
 
   /**


lib/application.js
```diff
@@ -125,15 +125,18 @@ module.exports = class Application extends Emitter {
 
   callback() {
     const fn = compose(this.middleware);
+    const self = this;
```
fengmk2: self come back again...

lib/application.js
```diff
@@ -125,15 +125,18 @@ module.exports = class Application extends Emitter {
...
 
-    return (req, res) => {
+    return function handleRequest(req, res) {
       res.statusCode = 404;
-      const ctx = this.createContext(req, res);
+      const ctx = self.createContext(req, res);
       const onerror = err => ctx.onerror(err);
       onFinished(res, onerror);
-      fn(ctx).then(() => respond(ctx)).catch(onerror);
+      fn(ctx).then(function handleResponse() {
```
fengmk2: > 137:20  error  Unexpected function expression  prefer-arrow-callback

eslint fail, maybe keep arrow function is better.

jonathanong:
```js
const handleResponse = () => {}

fn(ctx).then(handleResponse)
```

i think that'll name the function. not 100% sure, though

juliangruber: yeah, it would suck to give up arrows for this

fl0w: @jonathanong assertion was correct. Preassigning the arrow retains the naming.

```js
const Koa = require('koa') // @2
const app = new Koa()
const myMiddleware = (ctx, next) => { ctx.body = 'foo bar' }
app.use(mw)
app.listen(2020)
```

screen shot 2016-08-31 at 11 27 25

While a regular:
```js
app.use((ctx, next) => {
  ctx.body = 'foo bar'
})
```
yields:
screen shot 2016-08-31 at 11 30 25

lib/application.js
```diff
@@ -128,13 +128,16 @@ module.exports = class Application extends Emitter {
 
     if (!this.listeners('error').length) this.on('error', this.onerror);
 
-    return (req, res) => {
+    const handleRequest = (req, res) => {
```
fengmk2: @jonathanong "named arrow function" back

fengmk2:
named arrow function `const ab = () => 42; console.log(ab.name);` only work on node >= 6.

jonathanong:
> named arrow function const handleResponse = () => {} only work on node >= 6.

o i c. i'm thinking maybe when we actually release koa v2, we only support node >= 8 since that's where `await` is native. or maybe this is a feature that can only work on node >= 6 and let it be known that it does not work on node v4

iamchenxin:
@jonathanong Its a nice idea `only support node >= 8`. Backward compatibility is not always good.

jonathanong:
going to merge this and only support node 7.6+ :)

PR #807, opened 2016-09-06
fix: subdomains should be [] if the host is an ip

### LINKED ISSUE #775: this.subdomains is error parsed
when host is 127.0.0.1 this.subdomains=> [127,0,0]

welladamm:
Fixes #775

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -417,7 +417,13 @@ module.exports = {
 
   get subdomains() {
     var offset = this.app.subdomainOffset;
-    return (this.host || '')
+    var ip = /^(\d{1,3}\.){3}\d{1,3}$/gim;
+    var host = this.host || '';
+
+    // loose check if the host is an ip address
+    if (ip.test(host)) return [];
+
+    return host
       .split('.')
       .reverse()
       .slice(offset);
--- a/lib/request.js
+++ b/lib/request.js
@@ -5,6 +5,7 @@
  * Module dependencies.
  */
 
+var net = require('net');
 var contentType = require('content-type');
 var stringify = require('url').format;
 var parse = require('parseurl');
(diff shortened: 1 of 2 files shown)


lib/request.js
```diff
@@ -417,7 +417,13 @@ module.exports = {
 
   get subdomains() {
     var offset = this.app.subdomainOffset;
-    return (this.host || '')
+    var ip = /^(\d{1,3}\.){3}\d{1,3}$/gim;
```
PlasmaPower: This doesn't accommodate IPv6. Also, I don't like the use of a regex here, is there a better way to do this?

PlasmaPower: If you do keep the regex, remove the modifiers, they are all irrelevant.

PlasmaPower:
:+1: on premise, but I don't like the current implementation.

lib/request.js
welladamm: I don't think IPv6 is a big concern for the moment. However I did think about leaving a TODO in the comment, but, hey, using IPv4 address as the host is ugly already, who'd bother with IPv6 anyway.

welladamm: AFAIK, this regex might just be a pragmatic solution, but I'm open to better suggestions.

welladamm: You are right about the modifers. I'll fix this when I'm on my laptop.

tj: Looks like there's an is-ip module in npm, might as well delegate!

dougwilson: Isn't this part of the code Node.js `net` library? https://nodejs.org/dist/latest-v4.x/docs/api/net.html#net_net_isip_input

PlasmaPower: Looks like exactly what we need.

welladamm:
Shoot, totally forgot there's a native implementation. My bad.

dead-horse:
LGTM

dead-horse:
@welladamm make a PR to v2.x ?

welladamm:
@dead-horse Will do.

PR #808, opened 2016-09-07
fix: subdomains should be [] if the host is an ip

welladamm:
Closes: #775

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -5,6 +5,7 @@
  * Module dependencies.
  */
 
+const net = require('net');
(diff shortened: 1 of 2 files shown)


lib/request.js
```diff
@@ -404,7 +405,9 @@ module.exports = {
 
   get subdomains() {
     const offset = this.app.subdomainOffset;
-    return (this.host || '')
+    const hostname = this.hostname;
```
dead-horse: v1.x also should use `hostname`.

Issue #813, opened 2016-09-12
Use res.end instead of this.body When redirect


fundon:
Maybe should use `res.end` instead of `this.body=` ASAP.

```js
  redirect: function(url, alt){
    // location
    if ('back' == url) url = this.ctx.get('Referrer') || alt || '/';
    this.set('Location', url);

    // status
    if (!statuses.redirect[this.status]) this.status = 302;

    // html
    if (this.ctx.accepts('html')) {
      url = escape(url);
      this.type = 'text/html; charset=utf-8';
      this.body = 'Redirecting to <a href="' + url + '">' + url + '</a>.';
      return;
    }

    // text
    this.type = 'text/plain; charset=utf-8';
    this.body = 'Redirecting to ' + url + '.';
  },
```

juliangruber:
can you explain a bit more please?

fundon:
The `this.body=` has logic to much.
But when redirect, we clearly know what we want to do.

yorkie:
I think a benchmark would be more helpful to see what's the difference between them @fundon.

fundon:
@yorkie Ok,  I will add a benchmark.

fundon:
# Benchmark in Koa v2

## Change `redirect()` to below:

```js
  redirect(url, alt) {
    // location
    if ('back' == url) url = this.ctx.get('Referrer') || alt || '/';
    this.set('Location', url);

    // status
    if (!statuses.redirect[this.status]) this.status = 302;

    // html
    if (this.ctx.accepts('html')) {
      url = escape(url);
      this.type = 'text/html; charset=utf-8';
      this.res.end(`Redirecting to <a href="${url}">${url}</a>.`);
      return;
    }

    // text
    this.type = 'text/plain; charset=utf-8';
    this.res.end(`Redirecting to ${url}.`);
  },
```

## app example

```js
'use strict'

const Koa = require('koa')

const app = new Koa()

app.use((ctx, next) => {
  ctx.redirect('https://github.com')
})

app.listen(7002)
```

## results

### new redirect

```
 λ ~ > wrk -c 50 -t 8 -d 10 http://127.0.0.1:7002
Running 10s test @ http://127.0.0.1:7002
  8 threads and 50 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     8.50ms    3.34ms  43.70ms   84.25%
    Req/Sec   715.63    194.54     1.03k    60.75%
  57061 requests in 10.03s, 13.06MB read
[...]

PR #814, opened 2016-09-12
response: use res.end instead of this.body=

fundon:
#813

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -264,13 +264,13 @@ module.exports = {
     if (this.ctx.accepts('html')) {
       url = escape(url);
       this.type = 'text/html; charset=utf-8';
-      this.body = `Redirecting to <a href="${url}">${url}</a>.`;
+      this.res.end(`Redirecting to <a href="${url}">${url}</a>.`);
       return;
     }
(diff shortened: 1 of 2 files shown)


juliangruber:
the tests fail?

fundon:
The last case has a little problem. Should I change the describing?

lib/response.js
```diff
@@ -264,13 +264,13 @@ module.exports = {
     if (this.ctx.accepts('html')) {
       url = escape(url);
       this.type = 'text/html; charset=utf-8';
-      this.body = `Redirecting to <a href="${url}">${url}</a>.`;
+      this.res.end(`Redirecting to <a href="${url}">${url}</a>.`);
```
dead-horse: if you just want to skip all the logic in `this.body=`, maybe use `this._body=` is much better than `this.res.end`, it can break people's code if people want to set some header after this.redirect like this:

```js
  await next;
  ctx.set('cache-control', 'maxage=600');
})
```

fengmk2:
This change broken too much and the middleware flow was stop on redirect response.

PR #829, opened 2016-10-10
doc: Change usage of reserved word `static` in guide

saadq:
Very minor, but I think `static` is a reserved word so it might be better to use `serve` instead, I think that's what `koa-static` uses in its examples as well.


juliangruber:
cool, thanks!

PR #832, opened 2016-10-14
Add flag for cache body type

fundon:


fl0w:
What would the (side?)-effects be to use use Symbols for the flags?

fengmk2:
Better performance?

tj:
Hrm seems super overkill to me personally

jonathanong:
If we're improving performance we should see a benchmark.

But yeah, seems overkill to me too

dead-horse:
it will be a breanking change if developers already use `ctx.flag` for some other usage. also export `flag` is useless here.

fundon:
I closed it.

@dead-horse Just you said, `ctx.xxx` would be overridden.

Issue #703, opened 2016-04-04
Switch to a new testing system


PlasmaPower:
# Frameworks

| Framework | Simultaneous Testing | Multiple Processes For Multiple Files | Global Error Handling | Builtin Babel | NPM Downloads/Month |
| --- | :-: | :-: | :-: | :-: | :-: |
| Mocha (current) |  |  | ✓ |  | ![NPM Downloads/Month](https://img.shields.io/npm/dm/mocha.svg?style=flat-square) |
| Ava | ✓ | ✓ |  | ✓ | ![NPM Downloads/Month](https://img.shields.io/npm/dm/ava.svg?style=flat-square) |
| Jest | ✓ | ✓ |  | ✓ | ![NPM Downloads/Month](https://img.shields.io/npm/dm/jest-cli.svg?style=flat-square) |
# Request Libraries

| Request Library | Uses Promises | Builtin Response Testing | Response Testing Covers Common Use Cases | Reasonable Defaults For Testing | NPM Downloads/Month |
| --- | :-: | :-: | :-: | :-: | :-: |
| Superagent |  |  |  | ✓ | ![NPM Downloads/Month](https://img.shields.io/npm/dm/superagent.svg?style=flat-square) |
| Supertest (current) |  | ✓ |  | ✓ | ![NPM Downloads/Month](https://img.shields.io/npm/dm/supertest.svg?style=flat-square) |
| Supertest-As-Promised | ✓ | ✓ |  | ✓ | ![NPM Downloads/Month](https://img.shields.io/npm/dm/supertest-as-promised.svg?style=flat-square) |
| Assert-Request | ✓ | ✓ | ✓ | ✓ | ![NPM Downloads/Month](https://img.shields.io/npm/dm/assert-request.svg?style=flat-square) |
| Request |  |  |  | ✓ | ![NPM Downloads/Month](https://img.shields.io/npm/dm/request.svg?style=flat-square) |
| Request-Promise | ✓ |  |  |  | ![NPM Downloads/Month](https://img.shields.io/npm/dm/request-promise.svg?style=flat-square) |
| Node-Fetch | ✓ |  |  | ✓ | ![NPM Downloads/Month](https://img.shields.io/npm/dm/node-fetch.svg?style=flat-square) |
| Requisition | ✓ |  |  | ✓ | ![NPM Downloads/Month](https://img.shields.io/npm/dm/requisition.svg?style=flat-square) |
# Notes
- `(current)` indicates that Koa currently uses it.
[...]

PR #833, opened 2016-10-16
NPM scripts as Make alternative for Windows

DaAwesomeP:
I know there's been a lot of debate here about this, so I tried to do this without affecting the workflows of people who want to use Make: I didn't touch the Makefile or Travis file. Make on Windows is annoying to set up (NodeJS also doesn't work well with MSYS2), and I don't see why not use NPM scripts in that case.

They can be used in the following way:

``` bash
npm test
npm make-test  # will actually just run "make test"
npm run test-cov
npm run test-travis
npm run lint
```

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -138,7 +138,7 @@ Check out an example in koa's [test](test/babel/index.js).
 ## Running tests
 
 ```
-$ make test
+$ npm test
 ```
 
 ## Authors
--- a/package.json
+++ b/package.json
@@ -4,7 +4,11 @@
   "description": "Koa web app framework",
   "main": "lib/application.js",
   "scripts": {
-    "test": "make test"
+    "make-test": "make test",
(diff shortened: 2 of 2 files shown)


jonathanong:
we had a discussion about this and we were on the fence

fl0w:
@DaAwesomeP Why would "average Joe" run Koa's test suite though?

jonathanong:
i might be switching to windows so i might end up merging windows support. however, i wouldn't merge this in its current form since 300-character commands are not reasonable. we need to move them to `mocha.opts` or something

dougwilson:
> Why would "average Joe" run Koa's test suite though?

To contribute to Koa.

DaAwesomeP:
> 300-character commands are not reasonable. we need to move them to mocha.opts or something

@jonathanong I definitely agree. I'll add a `mocha.opts` file to the `test/` directory. Should I edit the Make rules as well to use this?

jonathanong:
https://github.com/koajs/koa/issues/703 gonna switch over to jest, which will make the test command 1 line and will make this easier to do!

PR #838, opened 2016-10-21
make response.append more smartly.

iyuq:
change response.append to set headers field to * when field is * or val contains *, also only add val of case-insensitive.
inspired by [vary](https://github.com/jshttp/vary).

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -460,14 +460,26 @@ module.exports = {
 
   append(field, val) {
     const prev = this.get(field);
-
+    let rlt = [];
     if (prev) {
-      val = Array.isArray(prev)
-        ? prev.concat(val)
-        : [prev].concat(val);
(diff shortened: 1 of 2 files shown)


dead-horse:
we already have a [response.vary](https://github.com/koajs/koa/blob/master/lib/response.js#L229) method, and `response.append` here is different with append in vary.

iyuq:
@dead-horse  I know that, but shouldn't other header's append to act like this?

dougwilson:
> but shouldn't other header's append to act like this?

Not really, no. The `Vary` header logic does this, because that is what is specified in the RFC for the `Vary` header. A general header append, like this method, is actually specified to allow anything, so there is no reason that you cannot send back
```
X-Header: 5
X-Header: *
X-Header: foo
```
The `*` behavior is specific to the `Vary` header.

Issue #846, opened 2016-10-28
mounting into an express app


nickb1080:
I'm exploring migrating a large production Express app to koa v2 (so all of this applies to the promise API of v2). The koajs.com docs say for `app.callback()`: 

> "You may also use this callback function to mount your koa app in a Connect/Express app." 

However, it seems like you must mount a koa app at the "bottom" of your Express app since koa will never call `next()` to yield control back to express (like in the 404 case). I can't see an obvious work around for this currently. If the function returned by `callback` returned the promise, I think something like the following would work:

``` js
const koaCallback = koaApp.callback();
expressApp.use(function (req, res, next) {
  koaCallback(req, res)
    .then(function () {
      // not sure exactly how to detect this but seems doable
      if (!koaSentResponse(res)) {
        next() // let express keep processing this request
      }
   });
});
```

Perhaps somewhat related -- what's the use case for setting `ctx.respond` to `false`?

PlasmaPower:
You probably want `res.finished`.

I'm not quite sure what `ctx.respond = false` would be used for, as a custom responder at the beginning of the chain should do the same job. It shouldn't be used, it's not officially supported and it's bound to break stuff.

PlasmaPower:
I think the docs for you're looking at would assume that Koa would be used to handle a subdirectory like `/api/...` or similar, as opposed Koa being the default and Express filling in the gaps.

nickb1080:
Huh. The [`ctx.respond = false` line](https://github.com/koajs/koa/blob/v2.x/lib/application.js#L194) even has a comment saying "allow bypassing koa". Just for some background, the use case here is I want to move routes in an existing application over to koa incrementally, without affecting clients.

PlasmaPower:
@nickb1080 Yep, I figured so. I don't think you want `ctx.respond = false` here though, I think you're initial solution is good with `if (!res.finished)` (instead of `if (!koaSentResponse(res))`). You could also manually assign routes to Koa of course, but that might be a pain depending on how many routes you have.

PlasmaPower:
Wait, I just realized you're going to get a 404 response from Koa of course. I'll see how that can be prevented.

PlasmaPower:
It looks like `ctx.respond = false` might be useful after all. Try making your first middleware `return next().then(() => ctx.respond = (ctx.body == null));` or similar (double equal sign as `undefined == null` but `'' != null`).

nickb1080:
[...]

PR #847, opened 2016-10-28
Return middleware chain promise from `callback()`

PlasmaPower:
Without this, there's no way of hooking into after Koa is done with the response. This can be extremely important if used with `ctx.respond = false`, as in #846 (see there for use case). While this is technically a breaking change, I don't expect it to break anything.

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -169,6 +169,12 @@ this.cookies.set('name', 'tobi', { signed: true });
 app.context.db = db();
 ```
 
+## app.usedAsMiddleware
+
+  Disables the default 404 status code and empty response.
+  This is very useful if you're using Koa as part of another framework, where Koa tries
+  to handle a request first and then falls back to the legacy framework.
+
 ## Error Handling
 
   By default outputs all errors to stderr unless __NODE_ENV__ is "test" or `app.silent` is `true`.
--- a/lib/application.js
+++ b/lib/application.js
@@ -130,7 +130,10 @@ app.callback = function(){
   if (!this.listeners('error').length) this.on('error', this.onerror);
 
   return function handleRequest(req, res){
-    res.statusCode = 404;
+    // don't set the default status code if Koa is part of a larger framework
+    if (!self.usedAsMiddleware) {
+      res.statusCode = 404;
+    }
     var ctx = self.createContext(req, res);
     onFinished(res, ctx.onerror);
     return fn.call(ctx).then(function handleResponse() {
@@ -215,6 +218,8 @@ function respond() {
 
   // status body
   if (null == body) {
+    // don't send back the default empty response if Koa is part of a larger framework
+    if (this.app.usedAsMiddleware) return;
     this.type = 'text';
     body = this.message || String(code);
     this.length = Buffer.byteLength(body);
--- a/lib/application.js
+++ b/lib/application.js
@@ -133,7 +133,7 @@ app.callback = function(){
     res.statusCode = 404;
     var ctx = self.createContext(req, res);
     onFinished(res, ctx.onerror);
-    fn.call(ctx).then(function handleResponse() {
+    return fn.call(ctx).then(function handleResponse() {
       respond.call(ctx);
     }).catch(ctx.onerror);
   }
--- a/test/application.js
+++ b/test/application.js
@@ -1110,3 +1110,41 @@ describe('app.response', function(){
     .expect(204, done);
   })
 })
+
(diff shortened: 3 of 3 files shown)


dougwilson:
i.m.o. it's not really a breaking change :)

PlasmaPower:
Yeah, the one situation I could see this breaking anything is `if (handler(req, res))`, but that shouldn't happen (hopefully). http://xkcd.com/1172/

PlasmaPower:
> Current coverage is 100% (diff: 100%)
> Merging #847 into master will not change coverage

I don't think that's how diffs work... :smile:

jonathanong:
can we add a test for the use-case if we add this? not opposed to it.

nickb1080:
This is great! One thing to be aware of is that even if you do `ctx.respond = false` and then use the promise-returning API from this PR to do further processing after koa finishes with a request, there is still the fact that koa has set `res.statusCode = 404`. It's not difficult to reset that value, but you risk overwriting something in the koa app *explicitly* setting status code to `404`. Is there a way for koa to internally mark a request as "maybe send 404 if `respond` is reached" without actually changing the state of the `res` object -- seems important especially in this context, because once control is transferred back to the enclosing app, the enclosing app sees the `404` attached by koa.

jonathanong:
> Is there a way

yes :) suggestions welcomed. i haven't thought about this yet, so feel free to make suggestions.

PlasmaPower:
Maybe `app.usedAsMiddleware`? It would remove the default `statusCode` setting and return early [here](https://github.com/PlasmaPower/koa/blob/8836cd30d140ec285e3ceadbeadeda1f1796af24/lib/application.js#L218) (don't `res.end()` if there's nothing to send).

PlasmaPower:
Okay, I've just added in `app.usedAsMiddleware`. Let me know how it looks and I can port it over to the v2 PR.

jonathanong:
im ok with it. anyone else? @koajs/owners

though maybe we could use a name with only 1-2 words :P

PlasmaPower:
Yeah a better name for the app variable would be appreciated.

PlasmaPower:
I can rebase and pull the changes into the v2 branch if you're ready to merge.

nickb1080:
Any updates on this?

PlasmaPower:
If a better config name is needed, maybe `disableDefaultResponse`?

EduardoRFS:
why not `àpp.middlewareMode` or `app.mode = 'middleware'`

PlasmaPower:
Because that describes a common use case, not its function.

nickb1080:
So as I look at this again, it occurs to me that instead perhaps we can be more generic. `usedAsMiddleware` is still a pretty specific name and purpose. I propose instead `app.defaultStatus` with an initial value of `404`. Thus we can end up with something like the following:

```
const app = new Koa();
console.log(app.defaultStatus); // logs "404"
app.defaultStatus = null; // override it

// main app logic and routes here
app.use(async ctx => {
  if (ctx.url !== "/user") return;
  const user = await db.getUser();
  ctx.body = user;
  ctx.status = 200; // <-- set status here
});

// final middleware, responsible for just checking if something upstream
// set a status. if so, we want to let Koa respond. otherwise, we will assume
// the outer framework will handle the request
app.use(async ctx => {
  if (ctx.status == null) {
    ctx.respond = false;
  }
});
```

This also lets us drop checking of `app.usedAsMiddleware` in the `respond` function, and we can let `ctx.respond` do it's original job.

(This all assumes we are returning the promise as already implemented in this middleware)

Thoughts?

nickb1080:
@PlasmaPower @jonathanong any thoughts on the comment above? would love to get support for embedding Koa into existing applications merged ASAP. Happy with the current implementation if people feel it's superior to what I proposed above.

jonathanong:
@nickb1080 yes, that looks feasible. It's been a while since I looked at this issue. if all we want to merge is `return fn.call(ctx).then(...)`, then I'm okay with that

nickb1080:
@jonathanong How do you want to proceed? Do you want me to open a PR against `v2.x` with the single small change of adding `return`?

fl0w:
@nickb1080 why not master?

PlasmaPower:
Why does the branch `v2.x` still exist? It seems to be just a previous point on master.

nickb1080:
Ah! I hadn't noticed the switchover had already happened.

PlasmaPower:
Yeah I agree. Closing this in favor of #848.

PR #850, opened 2016-11-06
this.is() returns null when no body is passed

felixsanz:
It doesn't return `undefined`. It returns `null`. 

https://github.com/jshttp/type-is#each-type-can-be

> `null` will be returned if the request does not have a body.

### DIFF
--- a/docs/api/request.md
+++ b/docs/api/request.md
@@ -195,7 +195,7 @@ this.body = yield db.find('something');
 
   Check if the incoming request contains the "Content-Type"
   header field, and it contains any of the give mime `type`s.
-  If there is no request body, `undefined` is returned.
+  If there is no request body, `null` is returned.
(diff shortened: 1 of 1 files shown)


PlasmaPower:
[Yep.](https://github.com/jshttp/type-is/blob/master/index.js#L128) :+1:

juliangruber:
thanks!

jonathanong:
would anyone like to help running make and pushing https://github.com/koajs/koajs.com once in a while? :)

stojanovic:
@jonathanong I can do that

stojanovic:
Done, I sent PR
Btw, I think we can automate that at some point, it should be simple just a webhook from this repo

PlasmaPower:
@jonathanong Could the travis build for Koa itself do that?

jonathanong:
@PlasmaPower i guess we could...

PR #856, opened 2016-11-22
Clarify behavior of ctx.throw in docs

noinkling:


avindra:
Sorry, what does the `.expose = true` do? Is that what passes thru the error to the response sent back to the client?

noinkling:
@avindra Yes.

PR #864, opened 2016-12-01
Amend typo, request.is() return null|fasle|string.

iamchenxin:
Modifying the test for `null` from `==` to `===` to make sure it must be `null`.


fengmk2:
+1

PR #875, opened 2016-12-27
add test: req.host should work with IPv6 Host

i5ting:


jonathanong:
@dougwilson what do you think?

i haven't seen ipv6 host names in the wild

dougwilson:
The `Host: [::1]` and `Host: [::1]:3000` are the headers one would see in the wild. Including the `[` and `]` in the hostname result seems to be the correct results AFAICT.

test/request/hostname.js
```diff
@@ -11,6 +11,12 @@ describe('req.hostname', function(){
     req.hostname.should.equal('foo.com');
   })
 
+  it('should work with IPv6 Host', function(){
+    var req = request();
+    req.header.host = '[::1]:3000';
+    assert.equal(req.hostname, '[::1]');
+  })
+
```
dotnil: What should req.hostname be if req.host is `[::1]` without the port? Or is req.host guaranteed with port attached such as `[::1]:80`? 

Otherwise, if req.host were `[::1]`, this req.hostname would be  `[::`.

dougwilson: Good catch. Change needs to take into account that port is optional.

fengmk2: Can we convert the ipv6 address to real one? `[::1]:3000` look like not real.

dougwilson: `::1` is the short hand for the main loopback address (the IPv6 version of 127.0.0.1).

PR #877, opened 2017-01-03
Mention the --harmony-async-await flag

SamuelBolduc:
Starting with Node 7 the `--harmony-async-await` flag can be enabled to support the async / await syntax directly in Node without compiling using Babel. It is planned to be available flag-free starting with Node v8.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -53,7 +53,7 @@ app.use(async ctx => {
 });
 ```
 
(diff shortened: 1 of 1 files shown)


jonathanong:
I don't think we should be mentioning it as it's buggy in node@7 and you won't need it in node@8

Issue #883, opened 2017-01-13
koa failing on citgm AIX


gdams:
currently `citgm koa` fails on AIX because `make test` can't be used. AIX needs `gmake test` if there is a way an if statement could be added?

fl0w:
Relates to discussion in https://github.com/koajs/koa/issues/703

ilkkao:
Is it `make test` in package.json that causes node citgm to fail? That could be replaced with a small wrapper shell script that runs gmake for AIX and make for others. That would still not support Windows but that I think should be handled separately under #703

gdams:
@ilkkao that is exactly what we need

jonathanong:
let's merge with https://github.com/koajs/koa/issues/559

PR #884, opened 2017-01-17
Support UNIX variants where make is not an alias for gmake

ilkkao:
Assuming bash is always available, this could solve #883

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "Koa web app framework",
   "main": "lib/application.js",
   "scripts": {
-    "test": "make test",
+    "test": "if type 'gmake' 2> /dev/null; then gmake test; else make test; fi",
     "update-authors": "git log --format='%aN <%aE>' | sort -u > AUTHORS"
   },
   "repository": "koajs/koa",


juliangruber:
what about `alias make="gmake"` in your local shell config?

gdams:
@juliangruber that's not a particularly good solution, it would be good to make this work on aix without any extra configuration

tj:
I like @juliangruber's suggestion, I don't know what's going on here but seems like a hack if this were something every library needs to support.

ilkkao:
I was thinking of making `npm test` to work out of the box on AIX. The report said that on AIX you only have GNU make.  

I don't use or care AIX myself but as the official node releases are tested with https://github.com/nodejs/citgm and Koa is one of the selected test packages, it makes sense to not let koa to fail there. Koa is still only one of the 100 test modules so changing environment just for Koa might not be that straightforward either.

I was able to minimize the change. It's definitely still a hack but affects only one line now.

tj:
Nice the new change is not bad, though I still think delegating here is appropriate. Koa doesn't care if it's selected for some CI thing or not, that's their deal.

ilkkao:
That also makes sense. I'll just leave this here.

gibfahn:
The reason we have to use `gmake` (i.e. GNU make) on AIX is because koa's Makefile uses the GNU extensions to POSIX make. The ideal fix would be to go through the makefile and have it be POSIX make compliant, but if we're going to do that we might as well just rewrite the Makefile in node right?

Note that other non-Linux *nix variants will probably have the same issues as AIX.

tj:
Fair enough, at least it's a small change now, though I can't say I've typed `gmake` in my life, too much Ubuntu/OSX I guess haha.

gdams:
@ilkkao is this good to land now?

ilkkao:
yes, tested on mac and fedora 24.

gdams:
I can confirm that this works on AIX

gdams:
This is failing for me on linux and darwin:
```
error:   failure             | The canary is dead: 
error:   failing module(s)   |                     
error:   module name:        | dbca1f21c7eaa33a37c1b136c7f9b26a19bcc233
error:   version:            |                     
error:   error:              | The canary is dead: 
error:   error:              | undefined                                                                                          
error:                       | > koa@1.2.4 test /tmp/be2b8a64-bb03-4d48-83dd-270f748855c3/dbca1f21c7eaa33a37c1b136c7f9b26a19bcc233
error:                       | > if type 'gmake' 2> /dev/null; then gmake test; else make test; fi                                
error:                       |                                                                                                    
error:                       | gmake: not found                                                                                   
error:                       |                                                                                                    
error:                       | make: *** No rule to make target `test'.  Stop.                                                    
error:                       | npm ERR! Test failed.  See above for more details.                                                 
error:   done                | The smoke test has failed.
info:    duration            | test duration: 46857ms
```

ilkkao:
Looks like the working directory is somehow wrong. My test was:

```
~/Code> cd koa && npm test
```

If the temp directory is still around, can you check that

```
ls /tmp/be2b8a64-bb03-4d48-83dd-270f748855c3/dbca1f21c7eaa33a37c1b136c7f9b26a19bcc233
```

is koa repository root directory and includes `Makefile`?

gdams:
I'll double check

gdams:
yup works for me, I was installing from npm, not github

gdams:
has there been any progress with this?

PlasmaPower:
If #926 goes through, this won't be necessary as it doesn't use make.

jonathanong:
https://github.com/koajs/koa/issues/703 gonna switch over to jest, which will make the test command 1 line and will make this easier to do!

PR #891, opened 2017-01-31
Adds the ability to insert middleware at desired index in middleware stack

khc:
Because Koa runs middleware in sequence created by pushing middleware into app.middleware, sometimes there is a need to insert middleware at different position. For example when we use async/await or Promises when composing middleware.

In example below we can insert "first" middleware at first position after Promise has been resolved. The app will output "first", although "second" middleware was added first to the stack.

```javascript
const Koa = require('koa');
const app = new Koa();

// async/await
(async () => {
  var index = app.middleware.length;
  var resource = await getResource();
  app.use(ctx => { ctx.body = 'first'; }, index);
})();

// Promise
var index = app.middleware.length;
getResource().then( resource => {
   app.use(ctx => { ctx.body = 'first'; }, index);
} );

app.use(ctx => { ctx.body = 'second' });

function getResource() {
   return new Promise( (resolve,reject) => {
      setTimeout( () => {
         resolve(true);
      }, 500 );
   } );
}

app.listen(3000);
```

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -157,9 +157,9 @@ http.createServer(app.callback()).listen(3001);
   You may also use this callback function to mount your koa app in a
   Connect/Express app.
 
-## app.use(function)
+## app.use(function,[index])
 
-  Add the given middleware function to this application. See [Middleware](https://github.com/koajs/koa/wiki#middleware) for
+  Add the given middleware `function` to the application middleware stack or insert middleware `function` at desired `index` in stack. See [Middleware](https://github.com/koajs/koa/wiki#middleware) for
   more information.
 
 ## app.keys=
@@ -184,7 +184,7 @@ ctx.cookies.set('name', 'tobi', { signed: true });
 
 ## app.context
 
-  `app.context` is the prototype from which `ctx` is created from. 
+  `app.context` is the prototype from which `ctx` is created from.
   You may add additional properties to `ctx` by editing `app.context`.
   This is useful for adding properties or methods to `ctx` to be used across your entire app,
   which may be more performant (no middleware) and/or easier (fewer `require()`s)
--- a/lib/application.js
+++ b/lib/application.js
@@ -98,11 +98,12 @@ module.exports = class Application extends Emitter {
    * Old-style middleware will be converted.
    *
    * @param {Function} fn
+   * @param {number} [idx]
    * @return {Application} self
    * @api public
    */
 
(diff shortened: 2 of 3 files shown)


tj:
-1 from me. In the case of adding middleware via async/await, I would just synchronized when they're added, ex:

```js
async function createApp() {
  const app = koa()

  const a = await something()
  app.use(whatever(a))

  const b = await something()
  app.use(whatever(b))

  return app
}
```

I don't think it's something Koa has to deal with, and working with indexes can be pretty obscure.

jonathanong:
-1 it's easier to do a `Promise.all(middleware).then(middleware => middleware.forEach(fn => app.use(fn)))`

khc:
Ok. So I understand your point of view. When we have full control on middleware composing, you proposed some easier alternatives. But sometimes we don't have that control. For example we want to build a Koa module for everyone to use. An during this module initialisation we need something that we have to wait for and than pass to our middleware. Of course Koa exposes`app.middleware` so we can do our own slicing, but we also have to handle all the errors Koa handles inside `app.use()`.

Someone would like to use `OurModuleThatWaitsForSomething` in their application (and have it in the first place), and might do something like this:

```javascript
const Koa = require('koa');
const app = Koa(); 

new OurModuleThatWaitsForSomething( { app: app } );
new SomeRestModule( { app: app } ); 
new Pug( { app: app } );
```

So are we supposed to tell potential users of our module to rewrite their middleware composition? Or is there some other solution that I haven't thought of?

tj:
IMO those things could still just return middleware, instead of calling `app.use()` themselves, that way you have full control over where it gets inserted.

khc:
So I guess this works (let's assume, that we only have control over Module class):

```javascript
class Module {
   constructor() {
      this.options = {};
      new Promise( (resolve,reject) => {
         setTimeout( () => {
            resolve('first');
         }, 500 );
      } ).then( text => {
         this.options.text = text;
      } ) ;
      return this.middleware();
   }
   middleware() {
      return ctx => {
         ctx.body = this.options.text;
      }
   }
}
app.use( new Module() );
app.use(ctx => { ctx.body = { text: 'second' } } );
 ```
But if we needed some other middleware in our Module and be able to enable/disable it on some promised data, I still don't know how to achieve it without splicing app.middleware.

jonathanong:
we won't be merging this in v2.x. sorry

PR #896, opened 2017-02-05
Adding option to use µWebSockets

zanaca:
Creating an option to inform that you want to use koajs using µWebSockets http wrapper, instead of node default http module.


jonathanong:
there's no point in including this in koa. you can just do `require('uws').createServer(app.callback()).listen()`

PR #897, opened 2017-02-09
Fix malformed content-type header causing exception on charset get

ruimarinho:
A malformed content-type header may cause an exception when the charset getter is triggered. This is unexpected as the returning value should always be a string. This can be exploited by bad actors.

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -319,7 +319,13 @@ module.exports = {
     var type = this.get('Content-Type');
     if (!type) return '';
 
-    return contentType.parse(type).parameters.charset || '';
+    try {
+      type = contentType.parse(type);
+    } catch (e) {
+      return '';
+    }
+
+    return type.parameters.charset || '';
   },
 
   /**
--- a/test/request/charset.js
+++ b/test/request/charset.js
@@ -26,5 +26,11 @@ describe('req.charset', function(){
       req.header['content-type'] = 'text/plain; charset=utf-8';
       req.charset.should.equal('utf-8');
     })
+
+    it('should return "" if content-type is invalid', function(){
+      var req = request();
+      req.header['content-type'] = 'application/json; application/text; charset=utf-8';
+      req.charset.should.equal('');
+    })
   })
 })


lib/request.js
```diff
@@ -319,7 +319,14 @@ module.exports = {
     var type = this.get('Content-Type');
     if (!type) return '';
 
-    return contentType.parse(type).parameters.charset || '';
+    try {
+      type = contentType.parse(type);
+    } catch (e) {
+      if (e instanceof TypeError) return '';
+      throw e
```
dead-horse: maybe we can remove `if (e instanceof TypeError)` here since this line will never hit.

others LGTM

ruimarinho: Not sure I understood your comment. The error is specifically a TypeError - https://github.com/jshttp/content-type/blob/master/index.js#L88.

dead-horse: `contentType.parse` will only throw `TypeError`, so `if (e instanceof TypeError)` is useless here, and line 326 will never be executed.

Even if `contentType.parse` throw some other unexpected errors, we also should return `""` here.

dougwilson: I agree with @dead-horse

nunofgs: While I agree that `contentType.parse` only throws a `TypeError` right now, there could also be a concern that another error could be thrown in the future (for example: a _SyntaxError_ introduced by a bug).

Koa has a declared dependency of: `"content-type": "^1.0.0",`. Adding a new error type could be conceivable without a major bump which would, in turn, then break this implementation.

It's unlikely but not outside the realm of possibility :)

dead-horse: @nunofgs as I said

> Even if contentType.parse throw some other unexpected errors, we also should return "" here.

promag: Exception handling should always be specific. If a new error is thrown in the future then new error handling can be added here. +1 to keep as it is.

ruimarinho:
I've updated the PR based on @dead-horse's feedback, but I still think it should not swallow the errors.

jonathanong:
i personally like throwing, but i would want the error to be a `4xx` error so that my error handler could catch it and tell the client that something is malformed. i wonder if we should have a "strict" app mode that throws instead of silently failing, but i'm -1 on adding new features.

i'm okay with this PR though if everyone else agrees

lib/request.js
```diff
@@ -319,7 +319,13 @@ module.exports = {
     var type = this.get('Content-Type');
     if (!type) return '';
```
jonathanong: true, but i prefer this explicitness

dead-horse:
@ruimarinho can you pick this fix to v2.x ?

ruimarinho:
Done @dead-horse.

PR #899, opened 2017-02-13
remove unused app.name

amitport:
`app.name` is mentioned in the documentation, but doesn't seem to have any semantic meaning.

Application instances are not frozen and as far as I can tell there's no reason to force people to use `name` specifically (people may want to use `id` / `machineId`, etc., or even better, don't rely on koa app's namespace to pass this information)

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -110,7 +110,6 @@ app.listen(3000);
   Application settings are properties on the `app` instance, currently
   the following are supported:
 
-  - `app.name` optionally give your application a name
   - `app.env` defaulting to the __NODE_ENV__ or "development"
   - `app.proxy` when true proxy header fields will be trusted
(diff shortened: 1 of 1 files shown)


docs/api/index.md
```diff
@@ -110,7 +110,6 @@ app.listen(3000);
   Application settings are properties on the `app` instance, currently
   the following are supported:
 
-  - `app.name` optionally give your application a name
```
dotnil: I think the convention might already been established. What is the point of removing it?

amitport:
a minimal interface better 
-> less potential for confusion/forward-incompatibility-issues/pointless-discussions
-> seem to be the general philosophy with koa (which I like)
-> that being said, if it is indeed 'already been established', probably no need to remove it

jonathanong:
👍 for this. anyone else?

tj:
👍 I don't remember what it's for haha

PR #907, opened 2017-02-21
Don't access res._headers directly when helpers exist

ilkkao:
Node HTTP module `response._headers` is considered internal to node itself. Its value will
change in a backwards incompatible way in the future node releases.

Use the documented accessor functions instead when they are available.

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -115,8 +115,14 @@ var proto = module.exports = {
       return;
     }
 
-    // unset all headers, and set those specified
-    this.res._headers = {};
+    // first unset all headers
+    if (this.res.getHeaderNames) {
+      this.res.getHeaderNames().forEach(function(name) { this.removeHeader(name) }, this.res);
+    } else {
+      this.res._headers = {}; // Node < 8
+    }
+
+    // then set those specified
     this.set(err.headers);
 
     // force text/plain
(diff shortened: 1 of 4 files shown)


lib/context.js
```diff
@@ -116,7 +116,12 @@ var proto = module.exports = {
     }
 
     // unset all headers, and set those specified
-    this.res._headers = {};
+    if (this.res.getHeaderNames) {
+      this.res.getHeaderNames().forEach(function(name) { this.removeHeader(name) }, this.res);
```
ilkkao: This line gets tested when the tests a are run with node 8 (test/context/onerror.js). I did a manual test run using node master branch.

ilkkao: There's no `removeAllHeaders()`, hence the loop.

ilkkao: But then again coverage was 100%, it can't drop. Added a test.

jonathanong:
I want to merge this, but I want to test against node nightly. anyone know if there's an easy way to test against node nightly on travis?

dougwilson:
https://github.com/expressjs/express/commit/cd7d241a5d07eb8fa3df025e96759ee15800e661

jonathanong:
@dougwilson you are the best

jonathanong:
#913 91c403a541ba91e3ba10afdfa8b28adeae74fbbe

Issue #902, opened 2017-02-18
Don't set Content-Length


alexhultman:
This header is set automatically by the HTTP module (much more efficient) and does not need to be set in the library. Setting Content-Type is fine since that is not something the HTTP module can infer.

Akcbryant:
It appears that the HTTP node module isn't setting the content-length when koa doesn't set it explicitly.

jonathanong:
yeah, the only time it _could_ be set is if you do `.end(buf)` without a `.write()`. however, it's impossible to know the content length is if the response body is streamed with `.write()`s

Akcbryant:
This can probably be closed then.

PR #908, opened 2017-02-21
No longer set content-length

Akcbryant:
The content-length header is automatically set by the http module.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -147,14 +147,12 @@ module.exports = {
     // string
     if ('string' == typeof val) {
       if (setType) this.type = /^\s*</.test(val) ? 'html' : 'text';
-      this.length = Buffer.byteLength(val);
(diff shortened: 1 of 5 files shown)


dougwilson:
The http module does not set content-length. You can see that by the fact that the tests for it failed. If the length is not set by Koa, then the response will be in the chunked format, adding extra bytes on the site for the chunk boundaries (and the zero length chunk at the end).

dougwilson:
The test/application goes through the http module, so would see any headers added by it.

Akcbryant:
Ah, I see. Thanks, this means issue #902 can be closed then?

dougwilson:
Maybe. Hard to really know without understanding more about what drove the issue. Perhaps there is an issue somewhere that made it seem like it was from the content-length ? Not sure.

PR #911, opened 2017-02-25
updated engines in package.json for 7.6.0 as minimal

panva:


jonathanong:
did this on purpose so that people still using babel won't get annoyed with error messages. no big deal :P

panva:
I thought so, but considering the probably higher userbase of lts 4 and 6 i think this is for the better.

PR #919, opened 2017-02-26
feat: set err.headerSent before app error event emit

fengmk2:


test/context/onerror.js
```diff
@@ -86,6 +86,29 @@ describe('ctx.onerror(err)', () => {
     });
   });
 
+  it('should ignore error after headerSent', done => {
+    const app = new Koa();
+
+    app.on('error', err => {
+      err.message.should.equal('mock error');
+      err.headerSent.should.equal(true);
```
dead-horse: the only difference is we can get `err.headerSent` in app's error listener?

jonathanong: I think so

jonathanong:
More of a feature than a fix but 👍

dead-horse:
@fengmk2 make a PR to v1.x please.

test/context/onerror.js
fengmk2: yes

fengmk2:
@dead-horse ok

PR #920, opened 2017-02-27
Create v2 Migration document

JeffRMoore:
Give v2 migration its own document.  

Hopefully, this creates a place that can collect any other necessary documentation on migrating.  I've added some of the comments from #533, which is hard to find now because the issue is closed.  Slight updates to main readme to make it friendlier to those new to Koa with v2.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -80,7 +80,7 @@ app.use((ctx, next) => {
 
 ### GeneratorFunction
 
-To use generator functions, you must use a wrapper such as [co](https://github.com/tj/co) that is no longer supplied with Koa.
+To use generator functions, you must use a wrapper such as [co](https://github.com/tj/co).
 
 ```js
 app.use(co.wrap(function *(ctx, next) {
@@ -91,35 +91,13 @@ app.use(co.wrap(function *(ctx, next) {
 }));
 ```
 
-### Old signature middleware (v1.x) - Deprecated
+### KOA v1.x Middleware Signature
 
-**Old signature middleware (v1.x) support will be removed in v3**
+The middleware signature changed between v1.x and v2.x.  The older signature is deprecated.
 
-Koa v2.x will try to convert legacy signature, generator middleware on `app.use`, using [koa-convert](https://github.com/koajs/convert).
-It is however recommended that you choose to migrate all v1.x middleware as soon as possible.
+**Old signature middleware support will be removed in v3**
 
-```js
-// Koa will convert
-app.use(function *(next) {
-  const start = new Date();
-  yield next;
-  const ms = new Date() - start;
-  console.log(`${this.method} ${this.url} - ${ms}ms`);
-});
-```
-
-You could do it manually as well, in which case Koa will not convert.
-
-```js
-const convert = require('koa-convert');
-
-app.use(convert(function *(next) {
-  const start = new Date();
-  yield next;
-  const ms = new Date() - start;
-  console.log(`${this.method} ${this.url} - ${ms}ms`);
-}));
-```
+Please see the [Migration Guide](docs/migration.md) for more information on upgrading from v1.x.
 
 ## Babel setup
 
--- a/Readme.md
+++ b/Readme.md
@@ -91,7 +91,7 @@ app.use(co.wrap(function *(ctx, next) {
 }));
 ```
 
-### KOA v1.x Middleware Signature
+### Koa v1.x Middleware Signature
 
 The middleware signature changed between v1.x and v2.x.  The older signature is deprecated.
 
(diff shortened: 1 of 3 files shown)


Readme.md
```diff
@@ -91,35 +91,13 @@ app.use(co.wrap(function *(ctx, next) {
 }));
 ```
 
-### Old signature middleware (v1.x) - Deprecated
+### KOA v1.x Middleware Signature
```
jonathanong: We never use all caps. can we change this to `Koa`? :)

Readme.md
```diff
@@ -80,7 +80,7 @@ app.use((ctx, next) => {
 
 ### GeneratorFunction
 
-To use generator functions, you must use a wrapper such as [co](https://github.com/tj/co) that is no longer supplied with Koa.
+To use generator functions, you must use a wrapper such as [co](https://github.com/tj/co).
```
jonathanong: i think we should remove this section as it would only confuse people

jonathanong:
this is great! anyone else have recommendations?

Readme.md
JeffRMoore: I thought it was weak, but didn't know if the unbundling of co mattered.  Seems not.  Removed.

Readme.md
JeffRMoore: Oops.  Fixed.

Readme.md
JeffRMoore: Oh, wait!  I completely misread this.  My mistake.  I thought you were referring to the dependency section in the migration page which is where I moved the "no longer supplied text" to.

@jonathanong When you refer to "this section" are you referring to the "GeneratorFunction" section?

Should I add back the dependency section I just removed?

JeffRMoore: After thinking about it, I agree.  The overview document should focus on the recommended way, which I think now is async and leave the other two formats for deeper documentation.  I'd like to submit that as a separate PR, though, so as to not block the merging of this migration guide.

JeffRMoore: I went ahead and restored the Dependency changes section I mistakenly removed.  I don't know if its meaningful, I'm not actually a v1.x user, I was attracted by the v2.x release.

docs/migration.md
```diff
@@ -0,0 +1,132 @@
...
+An explicit check for the `test` environment was removed from error handling. 
+
+## Dependency changes
+
+- [co](https://github.com/tj/co) is no longer bundled with Koa.  Require or import it directly.
+- [composition](https://github.com/thenables/composition) is no longer used and deprecated.
+
+## v1.x support
+
+The v1.x branch is still supported and may also receive feature updates.  Except for this migration
```
fl0w: > may also receive feature updates

Is this misleading? I think I saw @jonathanong expressing no intent to add features to v1.x branch from this point on. Just to make sure.

jonathanong: i would say, "should not expect feature updates".

JeffRMoore: Updated to "should not expect feature updates."

Readme.md
jonathanong: yup, I meant the `GeneratorFunction` section. we shouldn't be recommending people to use generator functions, and if they do, it should be the v1.x version that should be migrated to async functions. this is a generation function with `(ctx, next)` arguments, which is a mix of v1.x and v2.x APIs and is awkward

docs/migration.md
```diff
@@ -0,0 +1,132 @@
...
+  const start = new Date();
+  yield next;
+  const ms = new Date() - start;
+  console.log(`${this.method} ${this.url} - ${ms}ms`);
+}));
+```
+
+## Upgrading middleware
+
+You will have to convert your generators to async functions with the new middleware signature:
```
jonathanong: i would show `next()` in the signature

docs/migration.md
```diff
@@ -0,0 +1,132 @@
...
+  const user = await Users.getById(this.session.user_id)
+  ctx.body = { message: 'some message' }
+})
+```
+
+Upgrading your middleware may require some work. One migration path is to update them one-by-one.
+
+1. Wrap all your current middleware in koa-convert
+2. Test
+3. npm outdated to see which koa middleware is outdated
```
jonathanong: wrap commands in backticks so they're easier to read `npm outdated`

docs/migration.md
```diff
@@ -0,0 +1,132 @@
...
+4. Update one outdated middleware, remove using koa-convert
```
jonathanong: same with modules `koa-convert`

jonathanong:
except for small fixes 👍

docs/migration.md
JeffRMoore: Agreed. Fixed.

PR #921, opened 2017-02-27
Add Troubleshooting documentation

JeffRMoore:
This pulls the really fantastic [Common Problems](https://github.com/koajs/koa/wiki/Common-Problems) wiki page into the main documentation as a troubleshooting guide to give it greater exposure.  I also linked to it from the main read me for the same reason.

I generalized the headings for the two 404 causes after surveying closed tickets to give more surface area in the document for problem diagnosis.

I added an example of a missing `await`, also based on survey of closed tickets.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -1,6 +1,7 @@
 # Troubleshooting
 
 - [Whenever I try to access my route, it sends back a 404](#Whenever-I-try-to-access-my-route-it-sends-back-a-404)
+- [My response or context changes have no effect](#My-response-or-context-changes-have-no-effect)
 - [My middleware is not called](#My-middleware-is-not-called)
 
 If you encounter a problem and later learn how to fix it, and think others might also encounter that problem, please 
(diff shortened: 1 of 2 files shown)


docs/troubleshooting.md
```diff
@@ -0,0 +1,172 @@
...
+### Solution
+
+The solution for this is rather easy, simply add `return next()` to the end of your helper middleware.
+
+```js
+router.use(function (ctx, next) {
+  if (ctx.query.id) {
+    ctx.state.id = parseInt(ctx.query.id);
+  }
+  return next();
```
tj: does the `return` actually have significance? I forget what we did with Koa 2.x now haha..

fl0w: @tj yes, it needs to return the promise :)

jonathanong:
awesome

PlasmaPower:
Should we remove the wiki page then? I don't want the pages diverging.

jonathanong:
https://github.com/koajs/koa/wiki? sure

PR #922, opened 2017-02-27
Better expose documentation in overview readme

JeffRMoore:
Highlighting the existing documentation from the overview readme.

When I started evaluating 2.x this weekend, It took quite a while before I stumbled on the api docs, which I found helpful.  Hopefully, this will help improve visibility.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -121,6 +121,41 @@ app.use(convert(function *(next) {
 }));
 ```
 
+## Context, Request and Response
+
+Each middleware receives a Koa Context object that encapsulates node's request and response objects
+and provides many helpful methods for writing web applications and APIs.
+
+Koa provides a Request object as the `request` property of the Context.  Koa's Request object
+provides a clean abstraction that delegates to, rather than extends node's request object.  This
+helps provide a cleaner interface and reduces conflicts between middleware and with node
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -121,6 +121,41 @@ app.use(convert(function *(next) {
 }));
 ```
 
+## Context, Request and Response
```
jonathanong: maybe a quick code example would be helpful here

JeffRMoore: Now with more code.

Readme.md
```diff
@@ -121,6 +121,76 @@ app.use(convert(function *(next) {
...
+Koa provides a `Request` object as the `request` property of the `Context`.  
+Koa's `Request` object provides helpful methods for working with
+http requests which delegate to an [IncomingMessage](https://nodejs.org/api/http.html#http_class_http_incomingmessage)
+from the node `http` module.
+
+Here is an example of checking that a requesting client supports xml.
+
+```js
+app.use(async (ctx, next) => {
+  if (!ctx.request.accepts('xml')) ctx.throw(406);
```
jonathanong: this would be better as:

```js
ctx.assert(ctx.request.accepts('xml'), 406)
```

Issue #903, opened 2017-02-18
this.res._headers has changed in node master


ilkkao:
Just compiled Node from the master branch and tried to use koa. It failed because the property values in `this.res._headers` are now arrays instead of strings.

```
$ node
> http = require('http');
> resp = new http.ServerResponse({});
> resp.setHeader('Content-Type', 'foo');
> result = resp._headers
```

In Node 7.x, `result` is `{ 'content-type': 'foo' }`

In Node master (pre-8.0), `result` is `{ 'content-type': [ 'Content-Type', 'foo' ] }`

The change was made here: https://github.com/nodejs/node/commit/c00e4adbb4c6dd5bfbf22a3efcecb63c56958339#diff-286202fdbdd74ede6f5f5334b6176b5cR406 (Node PR: https://github.com/nodejs/node/pull/10558)

@mscdex noticed that it broke express and provided a fix https://github.com/jshttp/fresh/pull/20/files Something similar would be needed for koa.

I'll open this issue to make sure this issue is solved for koa in some way.

mscdex:
I would probably wait until https://github.com/nodejs/node/pull/10805 lands, which will provide more public API functions around accessing HTTP headers (there is a separate PR to deprecate `_headers`).

In some cases though using functions such as the existing `getHeader()` is enough for some use cases.

ilkkao:
I can prepare a PR when `response.getHeaders` from nodejs/node#10805 has landed

mscdex:
Just a heads up, the PR has landed now in https://github.com/nodejs/node/commit/3e6f1032a4fdb8ca7fba02c7d2103fba68c0ee1f.

ilkkao:
Koa PR open: https://github.com/koajs/koa/pull/907

ilkkao:
Code in the master branch still uses `this.res._headers` directly. Shouldn't the fix go to master too?

jonathanong:
@ilkkao yes, that would be great!

ilkkao:
I'll send a PR in few days

JasonBoy:
@ilkkao `getHeaders` and `getHeaderNames` are added in node-v7.7, the comment `Node < 8` in #907 could be updated to `Node < 7.7`?

jonathanong:
released in 2.1.0 and 1.3.0 thank you @ilkkao !!

PR #925, opened 2017-03-01
Don't use http.ServerResponse._header when accessors exist

ilkkao:
Node HTTP module response._headers is considered internal to node itself. Its value will
change in a backwards incompatible way in the future node releases.

Use the documented accessor functions instead when they are available.

This is an ES6 version of the commit that went to 1.x branch: https://github.com/koajs/koa/commit/91c403a541ba91e3ba10afdfa8b28adeae74fbbe

Should close https://github.com/koajs/koa/issues/903

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -121,8 +121,14 @@ const proto = module.exports = {
       return;
     }
 
-    // unset all headers, and set those specified
-    this.res._headers = {};
+    // first unset all headers
+    if (this.res.getHeaderNames) {
+      this.res.getHeaderNames().forEach(name => this.res.removeHeader(name));
+    } else {
+      this.res._headers = {}; // Node < 7.7
+    }
+
+    // then set those specified
     this.set(err.headers);
 
     // force text/plain
--- a/test/context/onerror.js
+++ b/test/context/onerror.js
@@ -1,8 +1,10 @@
 
 'use strict';
 
+const assert = require('assert');
 const request = require('supertest');
 const Koa = require('../..');
+const context = require('../helpers/context');
 
 describe('ctx.onerror(err)', () => {
   it('should respond', done => {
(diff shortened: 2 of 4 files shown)


jonathanong:
👍

test/response/header.js
```diff
@@ -16,5 +16,12 @@ describe('res.header', () => {
       res.res._headers = null;
       res.header.should.eql({});
     });
+
+    it('should use res.getHeaders() accessor when available', () => {
```
jonathanong: we should have a test `ctx.response.header` w/o mocking like this because we are testing node v8 in travis. the test should be backported to v1.x as well

lib/context.js
```diff
@@ -121,8 +121,14 @@ const proto = module.exports = {
       return;
     }
 
-    // unset all headers, and set those specified
-    this.res._headers = {};
+    // first unset all headers
+    if (this.res.getHeaderNames) {
+      this.res.getHeaderNames().forEach(name => this.res.removeHeader(name));
+    } else {
+      this.res._headers = {}; // Node < 8
```
fengmk2: These apis has been backport to node 7.7.0 https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V7.md#7.7.0

ilkkao: Updated.

test/context/onerror.js
```diff
@@ -169,5 +171,23 @@ describe('ctx.onerror(err)', () => {
         .expect('Content-Type', 'text/plain; charset=utf-8')
         .expect('Internal Server Error', done);
     });
+
+    it('should use res.getHeaderNames() accessor when available', () => {
```
ilkkao: @fengmk2 something like this? I avoided using context helper to get the real `http.ServerResponse`

test/response/header.js
```diff
@@ -10,6 +12,30 @@ describe('res.header', () => {
...
   });
 
+  it('should use res.getHeaders() accessor when available', () => {
+    const res = response();
+    res.res._headers = null;
+    res.res.getHeaders = () => ({ 'x-foo': 'baz' });
+    res.header.should.eql({ 'x-foo': 'baz' });
+  });
+
+  it('should return the response header object when no mocks are in use', done => {
```
ilkkao: @fengmk2 something like this? I avoided using context helper to get the real `http.ServerResponse`

fengmk2: +1

fengmk2:
Before merge, you should rebase master branch and run test again.

ilkkao:
@fengmk2 nice. Rebased now.

PR #848, opened 2016-10-28
Return middleware chain promise from `callback()`

PlasmaPower:
The v2.x version of #847, see there for more details.

PR #926, opened 2017-03-02
Use jest and node-fetch for tests

### LINKED ISSUE #703: Switch to a new testing system

PlasmaPower:
Resolves #703. One test (currently set to skip) is tied to #848. I'll make the change to unskip it on the PR last to get merged.


PlasmaPower:
Hmm... the travis error is very confusing to me. Can anyone replicate it locally?

The denoted file+line is just `});`. Also, there's no combination of `request` and `done` in any test, as the error would suggest. There's also never a circumstance where `request(...).get` would be accessed.

PlasmaPower:
Oh, that's because the tests were preformed on a rebase, and new tests have been added. I'll fix that.

PlasmaPower:
Rebased. Not sure why Travis is taking some time to run, but it should work when it does.

fengmk2:
I'm -1.

No, we use `supertest` for a long time and it work great for me.
I don't see any good reason change to use `node-fetch`.

PlasmaPower:
@fengmk2 Supertest is:

- **no longer maintained**
- based on callbacks not promises
- without a good way to test for the absence of headers

I think that we need to switch at some point in time without a doubt. I'd much prefer to switch now than later.

PlasmaPower:
Also, I can't see *any* advantage of supertest over node-fetch. I've already done all the work to switch, and in some cases the code clearly gets simpler, so why not?

PlasmaPower:
Travis has passed.

dougwilson:
R.I.P. Travis

magicdawn:
supertest is extend from superagent. and superagent has faux Promise support since v1.3.0
https://github.com/visionmedia/superagent/blob/v1.3.0/lib/node/index.js#L1027

-1 on node-fetch. `fetch` always reminds that's `GET` 😂

test/application/use.js
```diff
@@ -70,60 +65,50 @@ describe('app.use(fn)', () => {
...
   });
 
-  it('should output deprecation message for generator functions', done => {
-    process.once('deprecation', message => {
-      assert(/Support for generators will be removed/.test(message));
-      done();
-    });
-
+  it('should output deprecation message for generator functions', () => {
+    const warningPromise = eventToPromise(process, 'deprecation')
```
dead-horse: use promise here is strange, I'd like to use the original syntax

PlasmaPower: With the original syntax, the asserts for the result of the request were run after the test has finished. I don't know if they would be ignored or result in a strange error, but I do know that it wouldn't work properly. I can return it to its original syntax and remove the request asserts though.

PlasmaPower:
Ah, I see. I wasn't aware that superagent was updated. However, I still don't like using supertest, as once again it is unmaintained.

If we'd like to explicitly specify GET, then I can add that to the wrapper API. However, I still support node-fetch as the base.

test/context/cookies.js
```diff
@@ -1,35 +1,26 @@
...
-      .get('/')
-      .expect(204)
-      .end((err, res) => {
-        if (err) return done(err);
-
-        res.headers['set-cookie'].some(cookie => /^name=/.test(cookie)).should.be.ok;
-
-        done();
-      });
+    const res = await request(app, '/');
```
dead-horse: use supertest@3 with promise support is much better than node-fetch

PlasmaPower: Looking into it, it does have a new maintainer, so my point there isn't valid. However, I still prefer the syntax of node-fetch. May I ask why you prefer supertest?

PlasmaPower:
Sorry for switching to `node-fetch` without asking first, I thought the consensus was once we had async/await we should switch away from supertest.

test/context/cookies.js
PlasmaPower: Also, I'd like to point out that for the specific test you've commented on using node-fetch is significantly cleaner (12 lines to 3 lines).

PlasmaPower:
One other thing about my node-fetch wrapper: it automatically closes the HTTP server, so the Jest process will exit after finishing the tests (not the case with our current supertest usage).

fengmk2:
> without a good way to test for the absence of headers

We can send a pull request to supertest then make it happen.

test/application/response.js
```diff
@@ -1,34 +1,29 @@
...
     app1.use((ctx, next) => {
-      assert.equal(ctx.response.msg, 'hello');
+      expect(ctx.response.msg).toBe('hello');
       ctx.status = 204;
     });
 
-    request(app1.listen())
-      .get('/')
-      .expect(204, done);
+    return request(app1, '/').then(res => expect(res.status).toBe(204));
```
fengmk2: this equal to supertest

```js
return request(app1)
  .get('/')
  .end(res => assert(res.status === 204));
```

test/context/cookies.js
fengmk2: Using supertest way

```js
return request(app)
  .get('/')
  .expect(204)
  .expect(res => {
    assert(res.headers['set-cookie'].join(';').match(/^name=/));
  });
```

test/application/response.js
PlasmaPower: I understand that now that promise support has been added. Also, I'd think we'd do this instead:

```js
return request(app1)
  .get('/')
  .expect(204)
  .end();
```

Also, I'm not sure if Jest would recognize it as a Promise, given that it isn't true Promise, it just has `.then` and `.catch`.

One last thing I noticed - your code would seem to trigger this condition: https://github.com/visionmedia/superagent/blob/master/lib/request-base.js#L191

test/context/cookies.js
PlasmaPower: I'm assuming even if we switch back to supertest we'd keep expect?

test/application/response.js
PlasmaPower: Looks like jest does recognize it as a promise.

PlasmaPower:
The one thing I realized supertest has going for it is that it's error messages are more relevant to the context of the assertion.

At this point, I think supertest is acceptable. I'm still not sure it's better though.

test/application/use.js
PlasmaPower: I was thinking of another use of eventToPromise, I can switch this back.

.travis.yml
```diff
@@ -17,9 +17,12 @@ before_script:
   - "[ ! -f wrk/bin/wrk ] && rm -rf wrk && git clone https://github.com/wg/wrk.git && make -C wrk && mkdir wrk/bin && mv wrk/wrk wrk/bin || true"
   - export PATH=$PATH:$PWD/wrk/bin/
 script:
-  - make lint
-  - make test-travis
-  - make bench
+  - npm run lint
+  - jest --coverage
```
gibfahn: Might it be worth having an npm script for coverage (so you could do`npm run coverage`)?

package.json
```diff
@@ -4,7 +4,8 @@
   "description": "Koa web app framework",
   "main": "lib/application.js",
   "scripts": {
-    "test": "make test"
+    "test": "jest",
+    "lint": "./node_modules/.bin/eslint benchmarks lib test"
```
gibfahn: I think you can just do `    "lint": "eslint benchmarks lib test"`

PlasmaPower: I don't think so, we also don't want to use a global eslint even if it exists (plugins).

.travis.yml
PlasmaPower: I'm not sure that would be very helpful, I think `npm test --coverage` is good enough.

package.json
gibfahn: `npm run` prepends `./node_modules/.bin` to the `$PATH`, so it will use the local eslint even if it's also installed globally.

You can test it by adding `"tmp": "echo $PATH",` to the `package.json`.

[npm run docs](https://docs.npmjs.com/cli/run-script)

PlasmaPower:
@gibfahn Good points. I've added those, except I chose `npm run test-cov` instead of `npm run coverage` as that's what we previously had.

package.json
```diff
@@ -4,7 +4,9 @@
   "description": "Koa web app framework",
   "main": "lib/application.js",
   "scripts": {
-    "test": "make test"
+    "test": "jest",
+    "test-cov": "jest --coverage",
+    "lint": "./node_modules/.bin/eslint benchmarks lib test"
```
gibfahn: Did you miss committing the lint change? I see it in the commit message 😄

PlasmaPower: Yeah, oops! Fixed.

PlasmaPower:
Hmm, that Travis failure is very odd. Retrying.

PlasmaPower:
That time it passed. Given that it was a syntax error I'm not sure it was caused by jest. I'll investigate further if it occurs again.

PlasmaPower:
The test relying on #848 is proving to be a bit more difficult than thought. I'll look into it later today.

PlasmaPower:
Okay I've updated the [socket.writable test](https://github.com/PlasmaPower/koa/blob/jest-tests/test/application/index.js#L26) (depends on app.callback returning a promise). It's maybe a bit too complicated, but it should be fairly secure against future changes.

test/application/response.js
jonathanong: anything with a `.then()` method is a true promise ;)

test/application/use.js
```diff
@@ -70,60 +65,50 @@ describe('app.use(fn)', () => {
...
-    request(server)
-      .get('/')
-      .expect(404)
-      .end(err => {
-        if (err) return done(err);
-        calls.should.eql([1, 2, 3, 4, 5, 6]);
-        done();
-      });
+    const res = await request(server, '/');
+    expect(res.status).toBe(404);
```
jonathanong: not sure I like using `expect`. anyone else prefer we use node's `assert`?

test/context/cookies.js
jonathanong: yeah, I think we should stick with `supertest@3`. fewer changes and it works fine

test/application/use.js
PlasmaPower: It comes with jest, so I figured that's what we'd use.

test/context/cookies.js
```diff
@@ -40,13 +31,12 @@ describe('ctx.cookies.set()', () => {
           }
         });
 
-        request(app.listen())
-          .get('/')
-          .expect('.keys required for signed cookies', done);
+        const text = await (await request(app, '/')).text();
```
jonathanong: lol. I always try to only do 1 async action per line. gets crazy like this otherwise

PlasmaPower: Yeah I probably should've split that up.

jonathanong:
my opinions:

- stick to `supertest@3` instead of `node-fetch`, which seems to be the consensus
- use `assert` instead of `expect`, which I don't really care about

test/application/use.js
dead-horse: assert +1, don't like expect either, maybe we can try [power-assert](https://github.com/power-assert-js/power-assert) later, it has the same API with node's assert.

PlasmaPower:
OK, I'll open a new PR with those changes later today hopefully.

jonathanong:
@PlasmaPower thanks for all your work!!!

test/application/use.js
PlasmaPower: Hmm since we've managed to avoid transformations thus far, I'm not sure we'd want to add one. However, I agree that it'd be very useful.

PlasmaPower:
I'm leaning towards a custom supertest wrapper, as we also need to shutdown the server. Is that okay?

Also, for now, I think I'll implement header not existing tests through that wrapper, then maybe make a PR to supertest in the future.

PlasmaPower:
I haven't had much time to work on this, if anyone wants to pick it up feel free.

PR #941, opened 2017-03-15
use ES6 actually

tinovyatkin:
Refactor to use actual ES6 features permitted by node 7.x requirement

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -84,7 +84,7 @@ module.exports = {
   get href() {
     // support: `GET http://example.com/foo`
     if (/^https?:\/\//i.test(this.originalUrl)) return this.originalUrl;
-    return `${this.origin}${this.originalUrl}`;
+    return this.origin + this.originalUrl;
   },
 
   /**
@@ -257,13 +257,13 @@ module.exports = {
    */
 
   get fresh() {
-    const { method, ctx: { status: s, response }, header } = this;
+    const { method, ctx: { status, response }, header } = this;
 
     // GET or HEAD for weak freshness validation only
-    if (!['GET', 'HEAD'].includes(method)) return false;
+    if ('GET' != method && 'HEAD' != method) return false;
 
     // 2xx or 304 as per rfc2616 14.26
-    if ((s >= 200 && s < 300) || 304 == s) {
+    if ((status >= 200 && status < 300) || 304 == status) {
       return fresh(header, response.header);
     }
 
--- a/lib/request.js
+++ b/lib/request.js
@@ -257,7 +257,8 @@ module.exports = {
    */
 
   get fresh() {
-    const { method, ctx: { status, response }, header } = this;
+    const { method, response, header, ctx } = this;
+    const { status } = ctx;
 
     // GET or HEAD for weak freshness validation only
     if ('GET' != method && 'HEAD' != method) return false;
--- a/lib/request.js
+++ b/lib/request.js
@@ -84,7 +84,7 @@ module.exports = {
   get href() {
     // support: `GET http://example.com/foo`
     if (/^https?:\/\//i.test(this.originalUrl)) return this.originalUrl;
-    return this.origin + this.originalUrl;
+    return `${this.origin}${this.originalUrl}`;
   },
 
   /**
@@ -225,7 +225,7 @@ module.exports = {
    */
 
   get host() {
-    const proxy = this.app.proxy;
+    const { proxy } = this.app;
     let host = proxy && this.get('X-Forwarded-Host');
     host = host || this.get('Host');
     if (!host) return '';
@@ -242,7 +242,7 @@ module.exports = {
    */
 
   get hostname() {
-    const host = this.host;
+    const { host } = this;
     if (!host) return '';
     return host.split(':')[0];
   },
@@ -257,15 +257,14 @@ module.exports = {
    */
 
   get fresh() {
-    const method = this.method;
-    const s = this.ctx.status;
+    const { method, ctx: { status: s, response }, header } = this;
 
     // GET or HEAD for weak freshness validation only
-    if ('GET' != method && 'HEAD' != method) return false;
+    if (!['GET', 'HEAD'].includes(method)) return false;
 
     // 2xx or 304 as per rfc2616 14.26
     if ((s >= 200 && s < 300) || 304 == s) {
-      return fresh(this.header, this.ctx.response.header);
+      return fresh(header, response.header);
     }
 
     return false;
@@ -292,8 +291,7 @@ module.exports = {
    */
 
   get idempotent() {
-    const methods = ['GET', 'HEAD', 'PUT', 'DELETE', 'OPTIONS', 'TRACE'];
-    return !!~methods.indexOf(this.method);
+    return ['GET', 'HEAD', 'PUT', 'DELETE', 'OPTIONS', 'TRACE'].includes(this.method);
   },
 
   /**
@@ -460,8 +458,8 @@ module.exports = {
    * @api public
    */
 
-  accepts() {
-    return this.accept.types.apply(this.accept, arguments);
+  accepts(...types) {
+    return this.accept.types(...types);
   },
 
   /**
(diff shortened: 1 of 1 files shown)


tj:
the rest LGTM

lib/request.js
```diff
@@ -257,15 +257,14 @@ module.exports = {
    */
 
   get fresh() {
-    const method = this.method;
-    const s = this.ctx.status;
+    const { method, ctx: { status: s, response }, header } = this;
 
     // GET or HEAD for weak freshness validation only
-    if ('GET' != method && 'HEAD' != method) return false;
+    if (!['GET', 'HEAD'].includes(method)) return false;
```
JacksonTian: -1

tinovyatkin: Why?

dotnil: creates new array I suppose?

tinovyatkin: And? It's will be whole native code, easily extensible in case of... Way more readable. Any benchmarks?

PlasmaPower: I don't appreciate the drop in readability here. Maybe switch it over if we add more methods.

lib/request.js
```diff
@@ -257,15 +257,14 @@ module.exports = {
    */
 
   get fresh() {
-    const method = this.method;
-    const s = this.ctx.status;
+    const { method, ctx: { status: s, response }, header } = this;
```
PlasmaPower: I'm :-1: on this type of change (also in get host and hostname). In my opinion it only decreases readability.

lib/request.js
```diff
@@ -84,7 +84,7 @@ module.exports = {
   get href() {
     // support: `GET http://example.com/foo`
     if (/^https?:\/\//i.test(this.originalUrl)) return this.originalUrl;
-    return this.origin + this.originalUrl;
+    return `${this.origin}${this.originalUrl}`;
```
PlasmaPower: I much prefer the original. Formatting strings are meant for when you have a variable as part of a larger surrounding string, not for concatenation.

lib/request.js
tinovyatkin: In original code yoy have 3 logical operators and must be sure in prorities of it. In refactored code yoy have one logical function. I can understand if you will call it 'I don't want to learn new Array methods', but 'drop of readability'?

lib/request.js
tinovyatkin: Destruction is not a very readable thing if you are doing Javascript for last 15 years. Nor Lisp is readable for me. But for somebody who is starting to develop or contribute to Koa in 2017 and agreed with node 7.6+ requirement, destruction is perfectly readable and acceptable... It's not about comfort level, that's native language features now.

lib/request.js
PlasmaPower: Knowing the array methods is not the problem. The name `includes` speaks for itself. However, at first glance, the first way conveys its meaning much better: "method is not GET and method is not HEAD", whereas the second one "GET and HEAD does not include method" takes a bit of consideration.

This is important when skimming through code while refactoring or adding a feature.

lib/request.js
PlasmaPower: Again, understanding what the syntax does is a much smaller part of it. The previous way clearly lines up variables assigned on the left and variables used on the right. Here, it's a lot more complicated, with renaming across multiple levels of destructuring. Sure, after taking a close look it becomes obvious what's assigned to what, but an initial glance might confuse a newly created variable with one destructured from `this`.

lib/request.js
tinovyatkin: Reverted this back.

lib/request.js
tinovyatkin: Removed destructions renaming (it was to keep original code below), and it looks pretty neat now

lib/request.js
tinovyatkin: Reverted to original code here.

lib/request.js
```diff
@@ -257,15 +257,14 @@ module.exports = {
    */
 
   get fresh() {
-    const method = this.method;
-    const s = this.ctx.status;
+    const { method, ctx: { status, response }, header } = this;
```
jonathanong: I'm not a big fan of crazy destructuring like this lol

lib/request.js
```diff
@@ -292,8 +291,7 @@ module.exports = {
    */
 
   get idempotent() {
-    const methods = ['GET', 'HEAD', 'PUT', 'DELETE', 'OPTIONS', 'TRACE'];
-    return !!~methods.indexOf(this.method);
+    return ['GET', 'HEAD', 'PUT', 'DELETE', 'OPTIONS', 'TRACE'].includes(this.method);
```
jonathanong: imo we should define this array somewhere else, like at the top as `IDEMPOTENT_METHODS`

lib/request.js
PlasmaPower: Do you like the smaller one-item destructuring vs assignment?

jonathanong: I guess I would write this as:

```js
const { method, ctx } = this
const { status, response } = ctx
```

or maybe just have one assignment per line:

```js
const {
  method: {
    ctx: {
      status,
      method
    }
  }
} = this
```

not a fan of complex single-line stuff

jonathanong: also, you can just access `this.response`

lib/request.js
tinovyatkin: It will drop readability, but improve (some ms) performance... How frequently this method is called?

lib/request.js
tinovyatkin: Guys, `crazy` destruction is something like
```js
({ [{ method: { [dynoprope]: aVarName } }] } = anobject);
```
This `just a destruction`, developers that coming now to Javascript will see other way to write code as a bug. Don't be so conservative creating Koa for Node 7.6!

lib/request.js
PlasmaPower: Why would it drop performance? I don't see it making any difference in that regard. Since it's a global it might actually allow *more* optimizations, IDK.

tinovyatkin: JIT will anyway compile creation of static array of literals into native code early, so, I don't think over-optimization will improve anything here...

tinovyatkin: @PlasmaPower I wrote that it will improve performance, but drop readability. Giving in account our conversation regarding importance of readability earlier, I'm afraid to do such changes in a small method of external API that doesn't affect Koa performance in any way. But I'm quite OK to move static arrays into constants if needed.

lib/request.js
tinovyatkin: Removed _crazy_ nested destruction, so, it's not look so scary 😄

lib/request.js
```diff
@@ -225,7 +225,7 @@ module.exports = {
    */
 
   get host() {
-    const proxy = this.app.proxy;
+    const { proxy } = this.app;
```
fengmk2: I still think `const proxy = this.app.proxy;` is better on there is only one property read from app.

lib/request.js
```diff
@@ -242,7 +242,7 @@ module.exports = {
    */
 
   get hostname() {
-    const host = this.host;
+    const { host } = this;
```
sdd: Having `const { host = '' } = this;` would remove the need for the line below it

fl0w: If top comment is enforced (preference `const proxy = this.app.proxy`) I would assume to find `const host = this.host` for both same reason as above and consistency throughout.

maticrivo: @sdd I'm guessing doing the `if (!host) return ''` is more performant then doing `return (this.host || '').split(':')[0];` but please do correct me if I'm wrong

robjtede: @maticrivo wrapping the `this.host || ''` in parens definitely reduces readability.

I'd be in favour of the de-structuring approach in both cases on this PR. Readability is maintained and consistency can be kept if another method required multiple property reads.

lib/request.js
iyuq: What about `return this.method == 'GET' || this.method == 'HEAD' || this.method == 'PUT' || this.method == 'DELETE' || this.method == 'OPTIONS' || this.method == 'TRACE';`

tinovyatkin: @iyuq You can even write something more, like `return /^[GHPDOT][ACEILNTPOUR]{1,5}[TDES]$/.test(this.method)`. Not sure if looks nicer hovewer...

PR #946, opened 2017-03-20
use Buffer.from instead

fundon:
See:

* https://nodejs.org/dist/latest-v6.x/docs/api/buffer.html
* https://nodejs.org/dist/latest-v7.x/docs/api/buffer.html


dead-horse:
LGTM

PR #950, opened 2017-03-27
+ expose the Application::handleRequest method to extend the context

snowyu:
Recently, I wanna use the zone.js to persist the session across async task execution.
See [zone-koa](https://github.com/snowyu/zone-koa.js).
But I have to extract the code snippet from koa project. It's ugly and hard to maintain.
I do not like this way. The better is that exposing the Application::handleRequest method 
to extend the context.


jonathanong:
@koajs thoughts?

iyuq:
I don't think this is necessary, if someone wants to override the handleRequest or use the handleRequest method, he can extends the Koa class and override the callback function to achieve that. There is no other usage of the expose of the handleRequest method.

snowyu:
@iyuq Yes, you can override the `callback` function and copy the most original code to the new `callback`.
But how to maintain these codes when the `callback` function changed in KOA? Copy again?

I DO NOT think copy and modified the original `callback` function source code in KOA is a correct way. 

Give a way If you think it is not necessary.

dead-horse:
I'm ok with this

jonathanong:
imo it doesn't hurt and it makes the code a little easier to understand.

PR #955, opened 2017-03-29
add option to force async functions, fixes #954

### LINKED ISSUE #954: Errors not caught when using "common" sync functions - consider forcing async
The docs currently state that [Koa supports "common" functions as well as the new async/await](https://github.com/koajs/koa#middleware). While this works with happy path code, it falls apart with error handling. Here's an example where my error doesn't get caught:

```js
const Koa = require('koa');
const app = new Koa();

app.use(({response}, next) => {
	try {
		return next();
	} catch (ex) {
		console.log('Gotcha', ex.stack);
		response.body = ex.stack;
	}
});
app.use((ctx, next) => {
	console.log('About to throw');
	throw new Error('Catch me if you can.');
});
```

Updating middleware to be all async/await fixes the issue:

```js
const Koa = require('koa');
const app = new Koa();

app.use(async ({response}, next) => {
	try {
		return await next();
	} catch (ex) {
		console.log('Gotcha', ex.stack);
		response.body = ex.stack;
	}
});
	console.log('About to throw');
	throw new Error('Catch me if you can.');
});
```

If there is no easy solution to this problem, then might I suggest that you advise all middleware to be async? The koa-composer is [wrapping all middleware with a Promise](https://github.com/koajs/compose/blob/master/index.js#L44) already (which is likely the reason why the "sync" error is not caught). Just a personal opinion, but I think it might simplify all things if you advise all middleware to be async/await (easier for newcomers, fewer issues, easier docs, etc).

fl0w:
Just in case Koa devs wants to go this direction.
Related to #954

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -40,6 +40,7 @@ module.exports = class Application extends Emitter {
     super();
 
     this.proxy = false;
+    this.forceAsync = false;
     this.middleware = [];
     this.subdomainOffset = 2;
     this.env = process.env.NODE_ENV || 'development';
@@ -102,6 +103,9 @@ module.exports = class Application extends Emitter {
    */
 
   use(fn) {
+    if (this.forceAsync && fn.constructor.name !== 'AsyncFunction') {
+      throw new TypeError('middleware must be an AsyncFunction!');
+    }
     if (typeof fn !== 'function') throw new TypeError('middleware must be a function!');
     if (isGeneratorFunction(fn)) {
       deprecate('Support for generators will be removed in v3. ' +
--- a/test/application/use.js
+++ b/test/application/use.js
@@ -126,4 +126,16 @@ describe('app.use(fn)', () => {
 
     (() => app.use('not a function')).should.throw('middleware must be a function!');
   });
+
+  it('should only allow async function when forceAsync is true', () => {
+    const app = new Koa();
+    app.forceAsync = true;
+    (() => app.use(() => {})).should.throw();
+  });
+
+  it('should allow common and async function when forceAsync is false', () => {
+    const app = new Koa();
+    app.forceAsync = false;
+    [async () => {}, () => {}].forEach(v => (() => app.use(v)).should.not.throw());
+  });
 });


DesignByOnyx:
I wanted to leave a note here as this only covers half of the problem IMO. There is more detail in the original issue referenced in this PR - but in order to allow error handling with `try/catch`, all middleware must be asynchronous. If a single middleware is not async, then try/catch error handling breaks. Middleware can be asynchronous in one of two ways:

- middleware must be an 'AsyncFunction' (async/await)
- or middleware must return a Promise

jonathanong:
@koajs thoughts?

dead-horse:
look at the original issue #954, the reason of why common sync function can't catch the errors is because you **can't catch async error with try catch in common sync functions just like the old callback style**.

```js
app.use(({response}, next) => {
  try {
    return next();
  } catch (ex) {
    // you can't catch async error thrown by next in this common sync function
    console.log('Gotcha', ex.stack);
    response.body = ex.stack;
  }
});
app.use((ctx, next) => {
  console.log('About to throw');
  throw new Error('Catch me if you can.');
});
```

this is just like what we do with callback style: 

```js
try {
  fs.readFile(__filename, () => {
    throw new Error('oops');
  });
} catch (err) {
  // you can't catch async error by this try catch
}
```

force koa accepts async functions cannot fix this issue, because some other modules like `koa-router` is also use koa-compose and can use common sync function.

fl0w:
This doesn't fix anything really, we're still dealing with a user error.
If anything it's just a feature which is flirting with feature creep - and as far as I read it, has no real benefit to it.

lib/application.js
```diff
@@ -102,6 +103,9 @@ module.exports = class Application extends Emitter {
    */
 
   use(fn) {
+    if (this.forceAsync && fn.constructor.name !== 'AsyncFunction') {
+      throw new TypeError('middleware must be an AsyncFunction!');
+    }
```
nickb1080: If this does land, seems like this check should go after the following one (`if (typeof fn !== 'function') ...`) so that accidentally passing a non-function gives a more helpful error.

fl0w: I'll respectfully disagree. If `forceAsync` is true getting `must be an AsyncFunction` thrown seems pretty spot on no matter if you're passing a string or a regular function IMO.

But in any case, I'll just hold of changes and conflict resolution until "dust has settled".

bttmly: Fair enough, but maybe add a check so you don't get `cannot read property "constructor" of undefined`

fl0w: Ah, you're right. For that reason I agree that it's better to reorder the checks to avoid duplicating errors/checks. :) But again, holding off a bit until discussion settles.

Equim-chan:
A middleware doesn't have to be an async functions to use promises. "The middleware is a 'AsyncFunction'" isn't a sufficient condition to "the middleware returns a promise", so such check may not help much IMO.

fl0w:
I think this is a useless feature so I'm going to close this, unless koajs/owners objects - in which case you can reopen and I'll honour maintaining the PR.

PR #964, opened 2017-04-19
Added note about arrays

franciscop:
Added note about arrays being returned as JSON.

### DIFF
--- a/docs/api/response.md
+++ b/docs/api/response.md
@@ -117,7 +117,7 @@ so you can make a correction.
   - `string` written
   - `Buffer` written
   - `Stream` piped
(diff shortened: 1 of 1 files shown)


fl0w:
This occurs because of the resent header parser change in core if I'm not mistaken, correct?

franciscop:
@fl0w I am sorry I got some trouble understanding the English in that sentence, could you rephrase it please? What header parser change in core? (is that something that happened in Koa? or something I don't understand?)

If it makes it any more clear, I wrote an issue here that explains why it happens: https://github.com/koajs/koa/issues/965

PR #968, opened 2017-04-22
remove isJSON

joehecn:
Repeated judgment
``` javascript
// lib/response.js get length() {}
...
if (null == len) {
    if (!body) return;
    if ('string' == typeof body) return Buffer.byteLength(body);
    if (Buffer.isBuffer(body)) return body.length;
    if (isJSON(body)) return Buffer.byteLength(JSON.stringify(body));
    return;
}
```
``` javascript
// koa-is-json
function isJSON(body) {
  if (!body) return false;
  if ('string' == typeof body) return false;
  if ('function' == typeof body.pipe) return false;
  if (Buffer.isBuffer(body)) return false;
  return true;
}
```


jonathanong:
you're not wrong, but i'm wondering what the edge cases are here. i'm wondering if we should add more checks and understand the logic better https://github.com/joehecn/koa/blob/7672fa30faa36980b816142f8d2a018364435b7e/lib/application.js#L211 and https://github.com/joehecn/koa/blob/7672fa30faa36980b816142f8d2a018364435b7e/lib/application.js#L233

PR #978, opened 2017-05-04
docs: A middleware is always responsible for calling downstream

designgrill:
Irrespective of whether it is called by an upstream middleware or not.


jonathanong:
thanks. i'm not sure what the original intention was!

PR #980, opened 2017-05-05
Remove no-op code

ziyunfei:


PlasmaPower:
:+1: - https://github.com/jshttp/mime-types/blob/master/index.js#L88

PR #981, opened 2017-05-08
Use Jest

gillesdemey:
Inspired by https://github.com/koajs/koa/pull/926, fixes #703 and partially addresses #559

As I understand the consensus seems to be:

* Use `Jest`
* Prefer `assert`
* Keep using `supertest@3`

I used Promises / async where it made most sense and where it simplified the tests.

I'm using `--forceExit` because some tests do not properly close `app.listen`, I don't think it's that big of a deal and the current test suite has the same behaviour.

Haven't focused on making the benchmarks work cross-platform, might send a PR in the future to use [autocannon](https://github.com/mcollina/autocannon) instead of wrk (although `wrk` is fine for me, it seems some folks are concerned about the Windows platform).

PR #982, opened 2017-05-10
Use supertest's promise api where applicable

marvinhagemeister:
This saves a few lines here and there inside our test files.

```js
// New
it("Some test",() => {
  return request(server)
    .get('/')
    .expect(200)
    .then(res => {
       res.headers.should.have.property('something');
     });
});
```
versus:

```js
// Current
it("Some test", done => {
  request(server)
    .get('/')
    .expect(200)
    .end((err, res) => {
      if (err) return done(err);
       res.headers.should.have.property('something');
       done();
     });
});
```
There are a few places where this change is more or less cosmetic though.

### DIFF
--- a/test/application/context.js
+++ b/test/application/context.js
@@ -10,25 +10,25 @@ describe('app.context', () => {
   app1.context.msg = 'hello';
   const app2 = new Koa();
 
-  it('should merge properties', done => {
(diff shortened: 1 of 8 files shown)


dead-horse:
this is fixed by #981

PR #983, opened 2017-05-11
Add note to cookies usage.

liuyanghejerry:
Putting a note to cookies usage before others using cookies like this way:

```js
ctx.cookies.set('foo', 'bar');
ctx.cookies.get('foo') !== 'bar'
```


dead-horse:
+0

This may seem redundant.

PR #990, opened 2017-05-23
change node engines to 7.6.0 or higher.

sirius1024:
change node engines to 7.6.0 or higher.


dead-horse:
lots of people are still using babel and node 6, so we will keep it as-is for now.

PR #992, opened 2017-06-01
test: run ci on node 7,8

dead-horse:


.travis.yml
```diff
@@ -1,15 +1,8 @@
...
 node_js:
-  - stable
-  - 7.6
-matrix:
-  include:
-    - node_js: "8.0"
-      env: "NVM_NODEJS_ORG_MIRROR=https://nodejs.org/download/nightly"
-  allow_failures:
-    # Allow the nightly installs to fail
-    - env: "NVM_NODEJS_ORG_MIRROR=https://nodejs.org/download/nightly"
```
iyuq: I think node 6.x should be tested.

fl0w: Koa 2 is communicated as node >7.6 only.

iyuq: @fl0w Should‘t remove v6 until 8 become LTS?

fl0w: There's no point in testing and actively developing against node 6 if 7.6> is the target. There's however nothing working against a user to themselves transpile Koa 2 down to node 6.

jonathanong: i'm okay with testing node 6 as long as it doesn't require work (i.e. babelfying everything)

PR #1003, opened 2017-06-17
chore: update the node version to 7.6.0

wxnet2013:


jonathanong:
this is what we had previously, but removed it so that people who use babel with node v6 can still use koa

Issue #1000, opened 2017-06-12
have problem to handle urlencoded GET param


raphaelsoul:
the hello world example
```
const koa  = require('koa')

const app = new koa()

app.use(ctx => {
	ctx.body = 'ok'
})

app.listen(3000)
```

the following url got result:
* http://localhost:3000/oauth/steam/return?openid.ns=http%3A%2F%2Fspecs.openid.net%2Fauth%2F2.0&openid.mode=id_res
_empty response_
* http://localhost:3000/oauth/steam/return?openid.mode=id_res _receive string 'ok'_

test environment:
* OSX 10.12.5
* node 8.0.0
* koa 2.2.0

UPDATE:
on Safari it works fine, but failed on Google Chrome. Weired.
* Chrome 59.0.3071.86 (installed by brew)
* Safari 10.1.1 (12603.2.4)

raphaelsoul:
see also:  https://github.com/alexmingoia/koa-router/issues/360

jonathanong:
@raphaelsoul is this still an issue with #1004?

raphaelsoul:
Today I test it again. It works fine on Chrome. It seems to be fixed in v2.3.0. Thanks

PR #1004, opened 2017-06-17
Use node 7+ WHATWG parser for hostname, fixes #1002

### LINKED ISSUE #1002: Ipv6 hostname not supported?
Current method for extracting hostname portion from host (`host.split(':')[0];`) obviously does not work for Ipv6 hosts like [::1]:80. Is it an intention?

fl0w:
@jonathanong
> really, the ideal solution is a separate module that supports it. if you know of any, let us know!

Well, there's the node 7+ WHATWG compliant [url.URL](https://nodejs.org/api/url.html#url_the_whatwg_url_api), which could potentially help Koa "de-depend"  parseurl and querystring as well (I think).

~~This PR could be expanded to substitute public API for host/hostname/query and maybe a few more.
There would be some minor adjustments to output like `URL.port` returning `''` when :80 or :443 is provided (special scheme ports).~~ (performance impacted to heavily)

I'm also assuming `url.URL` would come with some ~~minor~~major performance hit.

```js
const { URL } = require('url')
[
  "https://github.com",
  "http://www.github.com:80",
  "https://127.0.0.1:8080",
  "http://[::1]",
  "https://[::1]:801",
  "https://[2001:db8::1428:57ab]",
  "http://[2001:db8::1428:57ab]:805/index.js"
].forEach(url => {
  let parsed = new URL(url)
  console.log(`${url} - hostname: ${parsed.hostname}, port: ${parsed.port}`)
})
```

```sh
$ node -v
v8.1.0
$ node whatwg.js
https://github.com - hostname: github.com, port: 
http://www.github.com:80 - hostname: www.github.com, port: 
https://127.0.0.1:8080 - hostname: 127.0.0.1, port: 8080
http://[::1] - hostname: [::1], port: 
https://[::1]:801 - hostname: [::1], port: 801
https://[2001:db8::1428:57ab] - hostname: [2001:db8::1428:57ab], port: 
http://[2001:db8::1428:57ab]:805/index.js - hostname: [2001:db8::1428:57ab], port: 805
```
*notice that all port 80 returns empty string, as http and https are [special scheme](https://url.spec.whatwg.org/#url-miscellaneous), thus parsed implicitly*


lib/request.js
```diff
@@ -5,6 +5,7 @@
  * Module dependencies.
  */
 
+const URL = require('url').URL;
 const net = require('net');
 const contentType = require('content-type');
 const stringify = require('url').format;
```
fl0w: Thought about destructing here (`const { stringify, URL } = require('url')`),
though I'm not seeing it elsewhere so assumed it was taboo in koajs/koa.

fl0w:
Btw, should the use of WHATWG parse expand, #1000 could be fixed as well.

```js
'use strict'
const { URL } = require('url')
const url = 'http://localhost:3000/oauth/steam/return?openid.ns=http%3A%2F%2Fspecs.openid.net%2Fauth%2F2.0&openid.mode=id_res'
const parsed = new URL(url)
console.log(parsed)
```
```sh
$ node -v
v8.1.0
$ node 1000-issue.js
URL {
  href: 'http://localhost:3000/oauth/steam/return?openid.ns=http%3A%2F%2Fspecs.openid.net%2Fauth%2F2.0&openid.mode=id_res',
  origin: 'http://localhost:3000',
  protocol: 'http:',
  username: '',
  password: '',
  host: 'localhost:3000',
  hostname: 'localhost',
  port: '3000',
  pathname: '/oauth/steam/return',
  search: '?openid.ns=http%3A%2F%2Fspecs.openid.net%2Fauth%2F2.0&openid.mode=id_res',
  searchParams: URLSearchParams {
     'openid.ns' => 'http://specs.openid.net/auth/2.0',
     'openid.mode' => 'id_res'
  },
  hash: ''
}
```

lib/request.js
jonathanong: @fl0w i think we didn't add it for node v4 support previously

fl0w:
@abouthiroppy that will mess with users who transpile Koa down to e.g. node 6.

lib/request.js
```diff
@@ -242,9 +243,12 @@ module.exports = {
    */
 
   get hostname() {
```
fl0w: This needs some adjusting - the code path is weird and for some reason I'm not using `host` whilst also assigning `const host = this.host` (relic from the past). I'll clean this up later today.

lib/request.js
```diff
@@ -242,9 +243,12 @@ module.exports = {
...
-    const host = this.host;
-    if (!host) return '';
-    return host.split(':')[0];
+    if (!this.parsedUrl) {
+      const host = this.host;
+      if (!host) return '';
+      this.parsedUrl = new URL(`${this.protocol}://${this.host}`);
```
fengmk2: How about the performance? Slow down or faster?

fengmk2: Use WHATWG parser for hostname when `host` start with `[`, otherwise keep the old way to get high performance.

fengmk2: image

fl0w: Min sharing your bench src?

fengmk2: @fl0w https://gist.github.com/fengmk2/5b25066b3bdedbfc7223f5a59902c725

jonathanong: seems like a potential module to me :)

fl0w:
@fengmk2 @jonathanong 
I exposed parsed URL and made it complete with `originalUrl` *(maybe used for other stuff now or in the future, like Punycode where `request.host` will return empty)*
The naming might be conflicting with `url`?
Kept memoization.

lib/request.js
fl0w: @jonathanong the host validation or bench src? :)

lib/request.js
```diff
@@ -244,10 +245,29 @@ module.exports = {
...
+   * @return {URL}
+   * @api public
+   */
+
+  get URL() {
+    if (!this.memoizedURL) {
+      const protocol = this.protocol;
+      const host = this.host;
+      const originalUrl = this.originalUrl || ''; // avoid undefined in template string
+      this.memoizedURL = new URL(`${protocol}://${host}${originalUrl}`);
```
fl0w: URL will throw TypeError if something's off.
Koa is generally has `value || ''` type of solutions in most cases.

Maybe wrap this in a try-catch and memoize empty object so that `this.URL.hostname` when malformed does not throw/crash the entire app?

fl0w: A try-catch shouldn't deopt since node 7.x (https://github.com/v8/v8/commit/9aac80f)

lib/request.js
```diff
@@ -244,10 +245,29 @@ module.exports = {
   get hostname() {
     const host = this.host;
     if (!host) return '';
+    if ('[' == host[0]) return this.URL.hostname; // IPv6
```
fengmk2: 👍

lib/request.js
dead-horse: > Maybe wrap this in a try-catch and memoize empty object so that this.URL.hostname when malformed does not throw/crash the entire app?

+1

lib/request.js
dead-horse: `this.URL` can be an empty object, `return this.URL.hostname || ''` here

urugator:
I am not sure if this is a fortunate solution.
I think it brings an inconsitency into how `request` object state relates to the underlying `req` object depending on the hostname format.
Until now the `request` was always up to date with the `req` object. So any changes of `req` were accordingly projected to the `request`.
Currently due to a memoization this is no longer true - if I do `request.headers['host'] = 'rewrittenHost'` the output of `request.hostname` will depend on the hostname format and whether it has been already accessed or not. Even wierdly the `request.host` can mismatch the `request.hostname`.   
I don't know to which extend the modifications of `req` inside middlewares are supported, but I think the behavior should not depend on hostname format as it can be a source of confusion and unpredictability.
Also note that returned `URL` object can be modified outside of `request` object, possibly changing the outcome of `request.hostname` again based on the hostname format...
Maybe you could make everything (protocol, hostname, url...) dependent on that URL object by default, keeping it as a source of truth instead of the wrapped `req`. It woudn't be ideal, but at least quite consistent.

fl0w:
Yes, this is a side effect I've been pondering about. As a quick solution, you can simply clear memoization by setting `request.memoizedURL = undefined`, this will force a new parse.

This was done because the parse felt quite costly (as noted by fengmk2).

fl0w:
@urugator though correct me if I'm wrong - this should only be of concern when hostname is called with an IPv6 host?

As far as inconsistencies, memoization could be removed but it would still require one to fetch a new URL object every time `request.req.[headers|host|*]` is changed.

urugator:
@fl0w The concern is that code behaves differently based on the hostname format.
In one request `hostname` depends on `req.headers[host]` and in another on `memoizedUrl` (where both are exposed to user as different things).
If I have a middleware for changing headers like `req.headers[host] = newHostname`, it will work for some requests, but not for anothers.
Similary middleware which modifies URL object like `request.URL.hostname = newHostname;` sometimes also modifies `request.hostname` and sometimes not.
The same goes for modifying `request.URL.host`, which sometimes also modifies `request.hostname`, but never `request.host`.
The behavior also depends on to which format the `headers["host"]` was possibly updated, whether it happend after or before URL was acessed (memoized), etc.   
I think it should be clear how things depends on one another and how changing one thing affects another and this behavior should stay always the same and shouldn't depend on host format.

Actually I would say that simply stating that Ipv6 hostname is not supported is a more acceptable solution, because seems easier to reason about.

However I still wonder whether the  WHATWG parser is really neccessary for this.

> require one to fetch a new URL

It has to be kept in sync manually, but not re-constructed:
```javascript
const newHost = "myLittleHost"
request.headers["host"] = newHost;
request.URL.host = newHost;
```

PR #1011, opened 2017-06-25
application: removed koa-convert (v3)

fl0w:
~~Added a separate commit for added test and keeping the check with a throw instead.
If d89e72f is an excessive guard, I'll just revert/squash it pre-merge.~~

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -81,9 +81,8 @@ app.use((ctx, next) => {
 
 ### Koa v1.x Middleware Signature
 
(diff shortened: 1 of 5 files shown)


dead-horse:
IMO we can hold this change until node 8 into LTS, then we can drop support of koa 1.x and start koa 3.x, so we don't need to maintain three branches at the same time.

fl0w:
@dead-horse I agree. I'll honour maintaining the PR unless you want me to close it for now?

3imed-jaberi (approved):

PR #1012, opened 2017-06-25
Fix context.inspect when called on the prototype

### LINKED ISSUE #837: console.log(require.cache) fails when koa is present
``` js
koa = require('koa')
console.log(require.cache)
```

will fail because of the custom `inspect` implementation in [`context.js`](https://github.com/koajs/koa/blob/master/lib/context.js#L27), which uses the custom `toJSON`, which relies on a properly set `this`, which is not the case.

not very important. But noteworthy - I guess
At least is costed me some time to figure it out.

CodingPower472:
Fixes #837

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -25,6 +25,7 @@ const proto = module.exports = {
    */
 
   inspect() {
+    if (this === proto) return this;
     return this.toJSON();
   },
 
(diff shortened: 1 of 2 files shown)


dead-horse:
can you add a test case for this change?

CodingPower472:
@dead-horse Done

PR #1018, opened 2017-07-06
use this.req insteadof this.ctx.req

hkc452:
use this.req directly

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -226,7 +226,8 @@ function respond(ctx) {
 
   // responses
   if (Buffer.isBuffer(body)) return res.end(body);
-  // include string and json
   if ('string' == typeof body) return res.end(body);
   if (body instanceof Stream) return body.pipe(res);
+
+  res.end(body);
 }


lib/response.js
```diff
@@ -177,8 +177,10 @@ module.exports = {
     }
 
     // json
-    this.remove('Content-Length');
+    val = JSON.stringify(val);
```
dead-horse: don't move these code to here. it will do stringify every time set `res.body`

lib/application.js
```diff
@@ -229,10 +229,5 @@ function respond(ctx) {
   if ('string' == typeof body) return res.end(body);
   if (body instanceof Stream) return body.pipe(res);
 
-  // body: json
-  body = JSON.stringify(body);
-  if (!res.headersSent) {
-    ctx.length = Buffer.byteLength(body);
-  }
   res.end(body);
```
gabel0287: What about this? 

```
if (body instance of Stream) return body.pipe(res);

return res.end(body);
```

jonathanong:
what's the purpose of this?

fl0w:
I'm going to close this PR due to inactivity.

PR #1021, opened 2017-07-07
modified some sample code of doc with wrong keyword (yield -> await)

purepennons:
Update some keyword of doc in official website to match Koa2.

* yield -> await


jonathanong:
thanks

PR #1023, opened 2017-07-13
docs: fix typo

haoxins:


fl0w:
-1, get is a [delegated method](https://github.com/koajs/koa/blob/master/lib/context.js#L185)

haoxins:
@fl0w here is `response`

fl0w:
Response and request do blend don't they :)  My bad - you're right.

PR #1025, opened 2017-07-14
Update Node Version requirement

uniibu:
This is in regards to the high severity vulnerability on all node versions.
https://nodejs.org/en/blog/vulnerability/july-2017-security-releases/

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -16,7 +16,10 @@
 
 ## Installation
 
-Koa requires __node v7.10.1__(or higher) or __node v8.1.4__(or higher) for ES2015 and async function support.
+Koa recommends the following node versions(for ES2015 and async function support):
+ - __Node v4.8.4__ (via babel)
+ - __Node v7.10.1__ (or higher)
+ - __Node v8.1.4__ (or higher)
 
 ```
 $ npm install koa
--- a/Readme.md
+++ b/Readme.md
@@ -16,7 +16,7 @@
 
 ## Installation
 
-Koa requires __node v7.6.0__ or higher for ES2015 and async function support.
+Koa requires __node v7.10.1__(or higher) or __node v8.1.4__(or higher) for ES2015 and async function support.
 
 ```
 $ npm install koa


uniibu:
@iyuq Hi, in case you have misread, this is not about which node version supports async/await. This is about the latest 'vulnerability' found on Node versions that are lower than the latest. This vulnerability affects all Node's release lines prior to the latest. So i think it is Koa team's duty to only allow Node versions that are not vulnerable. You can check the Node's announcement on the link given. 
Cheers!

BruceHem:
U are mistaken.

uniibu:
@BruceHem Not sure which part I have mistaken? Anyway, this commit is merely a 'doc fix', it doesn't change the `package.json` or any other files that forces user's to update to the latest versions. Therefor users can still install what ever they want. However, I still think it is Koa team's duty to inform the public and to not recommend vulnerable versions whether via documentation or other type of communication.

fl0w:
Which version Koa can or should run on are separate questions and unrelated to each other. By this logic the docs should just grep latest git tag for all major node.js releases 🤣 

-1

uniibu:
As stated on https://github.com/koajs/koa/pull/1025/commits/0e639210c63a0539e07aedf65a2582f62c72bf93 The purpose of this is merely a doc fix and matter of morality and conduct.

Would you `require Node v7.6.0 or higher` because it supports async/await but has vulnerabilities prior to the latest and users might end up getting affected in the future, or would you `recommend Node v7.10.1 or v 8.1.4` which also supports async/await and has fixed previous vulnerabilities? 

The doc says it "requires" 7.6.0 or higher, and yet further below `If you're not using node v7.6+, we recommend setting up babel with babel-preset-env`. The word 'require' is something that a user 'must' do or have and is not optional, and yet Koa can run below v7.6 using babel. I know many still uses v4 with babel, So i think it is best to list the versions in which Koa 'recommends', with security in mind.

iyuq:
if this commit is accepted, the version number should change each node version realeased to stay latest, there is no necessary to do that.
-1 too.

uniibu:
The commit does not necessarily say that it recommends the **latest** it just states the **minimum** recommended version prior to the "**Constant Hashtable Seeds vulnerability**" found on all Node release lines. So if Node releases v7.10.2 or 8.1.5, the version number on this doc does **NOT** need to be updated. 
I appreciate the opinion. 
Cheers!

iyuq:
I think this duty should belongs to node.js doc, not koa framework.

uniibu:
That is certainly true, it is why they have released a fix for it, tho i think it is Koa's duty to not recommend vulnerable versions.

BruceHem:
-1

jonathanong:
the versions are fine, but:

1. I wouldn't put anything babel related in the readme
1. The readme should really only recommend node `>= 8` now
2. The versioning should be done in package.json

We can have this in package.json's `.engines.node = "^4.8.4 || ^6.10.1 || 7.10.1 || ^8.1.4"` (or whatever the versions are.

PR #1027, opened 2017-07-16
allow bypassing compose by exposing fn (minor)

fl0w:
Am I the only one getting the feeling that koa-compose is unnecessary at this point?

For now expose `Application.fn`, but in a distant future perhaps remove koa-compose altogether?
By removing koa-compose Koa would:
  * Almost get its original signature back (`async ctx => {}`), `next()` would be obsolete.
  * Save a few closures
  * Simplify stack trace
  * Keep backwards compatibility (developers can compose themselves until ecosystem is up to par)
  * New USP: "parallel middleware" :)

Koa#next example:
```js
const compose = require('koa-compose') // batteries not included

// (ctx, next) signature mw
const legacyMw = compose([
  legacyMw1,
  legacyMw2
])

app.fn = ctx => {
  await mw1(ctx)
  await legacyMw(ctx)
  await Promise.all([
    mw2(ctx), // maybe redis session?
    mw3(ctx) // maybe microservice logger?
  ])
}
```

Obviously this can all be done today. Though that just raises the question again - is koa-compose really necessary? Maybe this time around a signature update is just too much though :)

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -39,6 +39,7 @@ module.exports = class Application extends Emitter {
   constructor() {
     super();
 
+    this.fn = undefined;
     this.proxy = false;
     this.middleware = [];
     this.subdomainOffset = 2;
@@ -123,7 +124,8 @@ module.exports = class Application extends Emitter {
    */
 
   callback() {
-    const fn = compose(this.middleware);
+    // Optional koa-compose bypass
+    const fn = this.fn || compose(this.middleware);
 
     if (!this.listeners('error').length) this.on('error', this.onerror);
 
--- a/test/application/fn.js
+++ b/test/application/fn.js
@@ -1,0 +1,34 @@
+
+'use strict';
+
+const request = require('supertest');
+const assert = require('assert');
+const Koa = require('../..');
+
+describe('app.fn = fn', () => {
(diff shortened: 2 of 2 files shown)


iyuq:
I don't think this is necessary, you can simply do this by:
```
const compose = require('koa-compose') // batteries not included

// (ctx, next) signature mw
const legacyMw = compose([
  legacyMw1,
  legacyMw2
])

const mw = ctx => {
  await mw1(ctx)
  await legacyMw(ctx)
  await Promise.all([
    mw2(ctx), // maybe redis session?
    mw3(ctx) // maybe microservice logger?
  ])
}

app.use(mw)
```
koa is a framework based on middleware, this PR will make it be able to not based on middleware.

fl0w:
> koa is a framework based on middleware, this PR will make it be able to not based on middleware.
The point was to raise the question, why do Koa need compose anymore anyway. The purpose was to purpose Koa middleware composition untouched, e.g. language feature. Not a library daisy-chain.

Buut, I think I jumped the gun. It's not as backwards compatible as I initially thought. I'm messing with the Promise unwinding. The "legacy"-stack would need to wrap everything else anyway.

iyuq:
compose can improve some performance.

PR #1031, opened 2017-07-22
Activate servers in docs.

tolmasky:
This pull request goes with https://github.com/koajs/koajs.com/pull/38, to turn on the two examples.

Reviewed by @tolmasky.

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -48,6 +48,7 @@ You can also use the [env preset](http://babeljs.io/docs/plugins/preset-env/) wi
 
   The obligatory hello world application:
 
+<!-- runkit:endpoint -->
 ```js
 const Koa = require('koa');
 const app = new Koa();
@@ -74,6 +75,7 @@ app.listen(3000);
   middleware to execute downstream, the stack will unwind and each middleware is resumed to perform
   its upstream behaviour.
 
+<!-- runkit:endpoint -->
 ```js
 const Koa = require('koa');
 const app = new Koa();


jonathanong:
anyone other than @tolmasky have an opinion on this?

fl0w:
I wouldn't mind these if the examples actually did a curl request or something to present output (without convoluting the example). As it is today, there's no benefit to being able to run these IMO.

tolmasky:
HI @fl0w -- not sure if you've had a chance to try the live examples yet (they're actually running on the site currently), but you can actually see a fair amount of output in this current version. When you click the link to the running site, the page itself opens in another tab, and the *server's output* shows up on the current page. So for example, in the second snippet, the user can actually inspect the logs placed in the various handlers:

screen shot 2018-08-01 at 8 33 49 am

The main idea here is that, especially for someone that is fairly new to these technologies, the friction from reading about a framework and to actually trying it is quite high -- installing node if they're coming from another language, npm installing the package itself, etc. In my experience the difference between thinking you understand something and hitting a wall hours later when you actually try it, vs being able to poke at something immediately is huge. With the running examples on the site, the user is allowed to immediately experiment with the async/await nature and start getting a feel for the framework without having to do any significant context switch:

27849775-76c14778-6104-11e7-858d-47673dbc58b9

Anyways, perhaps you were already fully aware of the functionality as it stands, in which case I apologize if any of this was redundant, but wanted to share some of the base ideas.

PR #1049, opened 2017-08-29
HTTP/2 has no status message

BernzSed:
This change allows Koa to be used with Node's http2 module, introduced in Node 8.4.0, which you can use if you launch node with `--expose-http2`


jonathanong:
I believe there was an issue saying that node fixed this. anyone else know?

dougwilson:
Yes, fixed in Node.js and in 8.5

BernzSed:
Looks like it's fixed, but it emits a warning. So this PR isn't needed anymore, but it will prevent that warning, if you care about that.

PR #1040, opened 2017-08-12
deps: update min engines

fl0w:
In light of [Jonathan's comment](https://github.com/koajs/koa/pull/1025#issuecomment-315571111) in PR #1025.

PR #1050, opened 2017-09-04
Update Node engine to 7.6.0 or higher

nickmccurdy:
The engine versions, Travis versions, and recommended version from the readme should be consistent.

From the readme:

> Koa requires **node v7.6.0** or higher for ES2015 and async function support.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "supertest": "^3.0.0"
   },
   "engines": {
-    "node": ">= 6.0.0"
+    "node": ">= 7.6.0"
   },
   "files": [
     "lib"


fl0w:
https://github.com/koajs/koa/pull/1040

nickmccurdy:
I disagree with allowing Node 4, it's best practice to not require your users to use a preprocessor for any supported engine versions. The other engine constraints (`^6.10.1 || ^7.10.1 || >= 8.1.4`) are just to avoid vulnerabilities, right?

fl0w:
So you think Koa should break all users who do transpile down to node 4?

nickmccurdy:
As far as I'm aware, npm now allows installations with invalid Node engine versions by default (it used to break them). It should display a warning, which I think could be a helpful reminder that those users should try a transpiler. Does this sound fair?

JounQin:
@nickmccurdy As I know, `yarn` will break on invalid Node engine versions.

https://yarnpkg.com/lang/en/docs/package-json/#toc-engines

nickmccurdy:
Thanks, I didn't know that it was intentional to have Node 6 supported in engines. I'll change this PR to make the documentation and test matrix more consistent instead, leaving engines alone.

nickmccurdy:
Nevermind, a discussion on this first would be better. Opened #1051.

PR #1053, opened 2017-09-05
Set request protocol to https when "X-AppEngine-HTTPS" header is "on"

caseyWebb:
Supports deployment on Google Cloud Platform, specifically Google Cloud Functions, which does *not* set the `X-Forwarded-Proto` header.

Note, this does not require `app.proxy` to be true.

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -390,7 +390,8 @@ module.exports = {
    * is enabled the "X-Forwarded-Proto" header
    * field will be trusted. If you're running behind
    * a reverse proxy that supplies https for you this
-   * may be enabled.
+   * may be enabled. "X-AppEngine-HTTPS" is respected
+   * for deployment on Google Cloud Platform.
    *
    * @return {String}
    * @api public
@@ -399,6 +400,7 @@ module.exports = {
   get protocol() {
     const proxy = this.app.proxy;
     if (this.socket.encrypted) return 'https';
+    if (this.get('X-AppEngine-HTTPS') === 'on') return 'https';
     if (!proxy) return 'http';
     const proto = this.get('X-Forwarded-Proto') || 'http';
     return proto.split(/\s*,\s*/)[0];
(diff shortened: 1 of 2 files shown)


fl0w:
This is necessary, though is there a more generic way to accomplish this? My reasoning is, as the number of PaaS providers grows, should Koa really be adding `if`-statements for each case?

lib/request.js
```diff
@@ -399,6 +400,7 @@ module.exports = {
   get protocol() {
     const proxy = this.app.proxy;
     if (this.socket.encrypted) return 'https';
+    if (this.get('X-AppEngine-HTTPS') === 'on') return 'https';
```
fengmk2: It should after `proxy` detect. Otherwise any one can send a request with the fake `X-AppEngine-HTTPS: on` header.

caseyWebb:
@fl0w, I agree. I was really hesitant to open this PR, but I needed it in order to deploy.

Maybe an `app.trustedHeaders`? Something like...

```javascript
// default
app.trustedProxyHeaders = {
     'X-Forwarded-Proto': 'https'
}

// in user code
app.trustedProxyHeaders['X-AppEngine-HTTPS'] = 'on'
```

jonathanong:
@caseyWebb I feel like this belongs in a separate module. 

how are you sure that `X-AppEngine-HTTPS` was set by a trusted proxy and the client didn't just set that header?

caseyWebb:
@jonathanong doesn't that same concern apply to `X-Forwarded-Proto`?

The main issue in my case is that `ctx.cookies` throws up when I attempt to set a cookie with `secure: true`, making it impossible to use secure cookies on GCP without a reverse proxy.

jonathanong:
> doesn't that same concern apply to X-Forwarded-Proto?

yes

@dougwilson what do you think?

caseyWebb:
Gentle ping. Is there anything I can do to get this merged or work towards an ultimate solution?

jonathanong:
i'm thinking this would be better as a method on `app`. something like `app.trustProtocol` or something...

miwnwski:
I really don't like this approach. if anything, the approach suggested by Jonathan seems more relevant IMO.

PR #1055, opened 2017-09-08
refactor: remove duplicate assignment in respond test

clarkdo:
Signed-off-by: Clark Du <clark.duxin@gmail.com>

### DIFF
--- a/test/application/respond.js
+++ b/test/application/respond.js
@@ -544,7 +544,6 @@ describe('app.respond', () => {
           ctx.length = fs.readFileSync('package.json').length;
           const stream = fs.createReadStream('package.json');
           ctx.body = stream;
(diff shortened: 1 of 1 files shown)


test/application/respond.js
```diff
@@ -544,7 +544,6 @@ describe('app.respond', () => {
           ctx.length = fs.readFileSync('package.json').length;
           const stream = fs.createReadStream('package.json');
           ctx.body = stream;
-          ctx.body = stream;
```
dead-horse: from the description of this test case(should keep content-length if overwritten with the same stream), I think we should not delete this line here.

clarkdo: right, sorry I missed the case description

Issue #998, opened 2017-06-09
Empty response if result is null


damianobarbati:
Why the following returns an empty response? `null` is a valid json value.

```
const config = require('./package.json');

const http = require('http');
const koa = require('koa');
const cors = require('kcors');
const compress = require('koa-compress')
const noTrailingSlash = require('koa-no-trailing-slash');
const json = require('koa-json');
const body = require('koa-body');
const send = require('koa-send');
const router = require('koa-router')();

const app = new koa();

app.use(cors());
app.use(compress());
app.use(noTrailingSlash());
app.use(json({ pretty: true, spaces: 4 }));
app.use(body({ formLimit: '5mb', jsonLimit: '5mb', strict: false, multipart: true }));

app.use(async (ctx, next) => {
    try {
        await next();
    }
    catch(error) {
        ctx.status = 400;
        ctx.body = error.message || error;
    }
});

router.all('/ciao', async ctx => {
    ctx.body = null;
});

app.use(router.routes());
http.createServer(app.callback()).listen(8080);
```

Reponse is empty:
```
macbook:react-app damiano$ curl -v localhost:8080/ciao
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8080 (#0)
> GET /ciao HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.51.0
> Accept: */*
> 
< HTTP/1.1 204 No Content
< Vary: Origin, Accept-Encoding
< Date: Fri, 09 Jun 2017 14:17:57 GMT
< Connection: keep-alive
< 
* Curl_http_done: called premature == 0
* Connection #0 to host localhost left intact
macbook:react-app damiano$ 
```

damianobarbati:
Nobody experiencing this?! :|

fengmk2:
`null` body meaning no content on koa https://github.com/koajs/koa/blob/master/lib/response.js#L139

damianobarbati:
@fengmk2 that's not good: what about an API endpoint returning "null" as data?
Shouldn't koa allow some kind of config about it?
`null` is a valid json value but returning a 204 will trigger problems front-end side.
We cannot assume everybody is enclosing result in an object.

tj:
Just my opinion but personally I most people expect an object or array with json these days, apis returning primitives is a little weird, even if it's technically fine

damianobarbati:
@tj I agree, but in some cases it may be needed and returning a 204 could be unexpected.
[...]

PR #1059, opened 2017-09-14
retrun json "null" if ctx.type is set to json and ctx.body is set to …

likegun:
issue [998](https://github.com/koajs/koa/issues/998)


lib/response.js
```diff
@@ -137,7 +137,10 @@ module.exports = {
 
     // no content
     if (null == val) {
-      if (!statuses.empty[this.status]) this.status = 204;
+      if(!statuses.empty[this.status]) {
```
fl0w: Nit: space between if and parenthesis.

@jonathanong the reason this passes tests is because --fix is automatically run on travis, and so it'll fix it but the commit remains "invalid" according to eslintrc.yml.

lib/response.js
```diff
@@ -137,7 +137,10 @@ module.exports = {
...
+        if(this.type == 'application/json') return this._body = 'null';
```
fl0w: Same as above.

likegun:
@fl0w done

dead-horse:
IMO it is a breaking change, and we have to hold this PR until we plan to develop koa@3.x.

damianobarbati:
@dead-horse do you have any approximate schedule for the 3.0 to ship this within?

jonathanong:
@damianobarbati there's no schedule. if we get another PRs to justify a v3, we should push.

Issue #1044, opened 2017-08-18
don't set header when `respond=false` and header sent


dead-horse:
many people using `next.js` or migrate there app from express will set `respond=false`, but lots of koa's middleware will set header like:

```js
async (ctx, next) => {
    await next();
    ctx.set('foo', 'bar');
}
```

it will cause `Can't set headers after they are sent` error if we already sent response body before. Maybe we can make `ctx.set()` invalid when `respond=false` and header sent. (We can log something when NODE_ENV is develop to hint developer this unexpected behavior)

related issues: 

- https://github.com/koajs/session/pull/95
- https://github.com/koajs/session/issues/89

Any thoughts?

nordsimon:
I just saw your reference @dead-horse. From my point of view, respond.false is very useful and sometimes neccessary in specific implementations. I would go for logging a warning in develop (or via debug) module mode but still allow and fix the behaviour. Maybe add some more info what side effects respond false has ?

nordsimon:
And yes, I see that this should actually live in koa and not in koa-session, I will close my PR in order to find a solution here

dead-horse:
this is already fixed in https://github.com/koajs/koa/pull/1137

PR #1060, opened 2017-09-15
fix: this.body setter does not respect this.respond

jasonxia23:
It's easy to fix the problem, but pretty hard to write a test case, since error is thrown after the response sent, and would not close the connection.
Would be appreciate if someone can help.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -131,6 +131,9 @@ module.exports = {
     var original = this._body;
     this._body = val;
 
+    // falsy respond
+    if (this.ctx.respond === false) return;
+
     // no content
     if (null == val) {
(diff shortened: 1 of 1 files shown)


likegun:
I feel confused, is there a situation people want to handle response by themself and also want to set context.body in the meanwhile?

dead-horse:
related issue: https://github.com/koajs/koa/issues/1044

jasonxia23:
The PR is based on this issue.

jonathanong:
can you make a test case where this is necessary? why not just check `.headerSent` instead?

jonathanong:
closing in favor of https://github.com/koajs/koa/pull/1078

PR #1062, opened 2017-09-17
tools: remove --fix to catch lint errors

fl0w:
Since `--fix` always runs, lint on travis does not non-zero exit unless it's "unfixable".


fl0w:
I'm not sure why the test failed?

Issue #477, opened 2015-09-07
HTTP/2 support


pensierinmusica:
Hi,

Is there any plan to support [HTTP/2](https://en.wikipedia.org/wiki/HTTP/2) in Koa?

Thanks!

jonathanong:
you can use https://github.com/indutny/node-spdy, which AFAIK supports HTTP 2

koa is server agnostic - just create wrap the app in whichever server implementation you'd like. 

if you were looking for a http push helper, you can try https://github.com/jshttp/spdy-push

pensierinmusica:
Hi @jonathanong, thanks for clarifying that Koa is server agnostic, I will look into the server implementation that best suits my use case!

Apart from this, in [Koala's docs](https://github.com/koajs/koala/blob/master/docs/spdy.md) there's a section that states:
>Once SPDY becomes an HTTP standard and is baked into node.js, some of these features will be placed in Koa.

In my understanding, HTTP/2 has replaced and extended the SPDY protocol (which [will not be supported from Google](https://blog.chromium.org/2015/02/hello-http2-goodbye-spdy-http-is_9.html) from early 2016 onwards).

As for the differences between SPDY adn HTTP/2, some information can be found in the Wiki link that I posted in the first comment: https://en.wikipedia.org/wiki/HTTP/2

Putting the two pieces of information together, I wondered if there was any plan to integrate HTTP/2 as the standard `app.listen()` in future Koa versions.

Cheers!

jonathanong:
`app.listen()` will continue use `require('http')` by default. we just gotta wait until node supports http2 out of the box.

rally25rs:
I'm curious if anyone actually got this to work. I was trying to use [node-http2](https://github.com/molnarg/node-http2) and this very simple example:

```
var http = require('http2');
var koa = require('koa');
var app = koa();

var options = {
  key: fs.readFileSync('./localhost.key'),
  cert: fs.readFileSync('./localhost.crt')
};

var handleRequest = function*(next) {
  this.body = 'Hello World';
  yield next;
  console.log('done');
};

app.use(handleRequest);
http.createServer(options, app.callback()).listen(8080);
```

never completes the request. From the client perspective, the request is 'pending' forever, and there are no errors in node. The line `console.log('done');` is hit and writes to the console.

If I swap the `require` from `'http2'` to `'http'` then it works fine.

tomchentw:
[...]

PR #1063, opened 2017-09-17
http: allow passing server object, for future node http2

fl0w:
Just trying to get this ball rolling because there's some activity in closed issue #477.

This is what I'm purposing to allow user space to experiment with http2 without breaking backwards compatibility.

Purposed API would become:

```js
const http2 = require('http2')
const Koa = require('koa')
const app = new Koa({ server: http2 })
```

I could move the require into the construction if args.server is falsey to avoid loading the library at all if http2 is chosen - not sure about the implications of doing so. I'd still have to be done during initialisation of a Koa application so my two cents would be to do it. Not sure how Koa team feels about a require in a constructor thought :)
```js
constructor() {
  super()
  ...
  this.server = args.server || require('http');
  ...
}
```

**PS** I am embarrassingly ignorant about http2, and the node implementation.

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -19,10 +19,19 @@ const accepts = require('accepts');
 const Emitter = require('events');
 const assert = require('assert');
 const Stream = require('stream');
-const http = require('http');
 const only = require('only');
 const convert = require('koa-convert');
 const deprecate = require('depd')('koa');
+
+function requireServer(server = 'http') {
+  if ('http' == server) {
+    return require('http');
+  } else if ('http2' == server) {
+    return require('http2');
+  } else {
+    return server;
+  }
+}
 
 /**
  * Expose `Application` class.
@@ -36,12 +45,13 @@ module.exports = class Application extends Emitter {
    * @api public
    */
 
-  constructor() {
+  constructor(args = {}) {
     super();
 
     this.proxy = false;
     this.middleware = [];
     this.subdomainOffset = 2;
+    this.server = requireServer(args.server);
     this.env = process.env.NODE_ENV || 'development';
     this.context = Object.create(context);
     this.request = Object.create(request);
@@ -60,7 +70,7 @@ module.exports = class Application extends Emitter {
 
   listen(...args) {
     debug('listen');
-    const server = http.createServer(this.callback());
+    const server = this.server.createServer(this.callback());
     return server.listen(...args);
   }
 


likegun:
Would it be better with:

```javascript
const Koa = require('koa')
const app = new Koa({ server: 'http2' })
```

```javascript
this.server = require(args.server == 'http2' ? 'http2' : 'http');
```
Otherwise if people pass wrong server value to constructor,they will get error like "this.servercreateServer is not a function" and feel confused.

iyuq:
I don't think this is necessary. We can simply use http2 in the following way.
```
const Koa = require('koa');
const app = new Koa();

const http2= require('http2');
http2.createServer(app.callback()).listen(...);
```

fl0w:
@iyuq I agree with you, but that sort of opens the can for me :) In that case Koa should be completely agnostic and even remove the `koa#Application.listen` sugar!

dead-horse:
I think `const app = new Koa({ server: 'http2' })` is better, `koa#Application.listen` sugar no need to require http or http2 manually. so we can create a server by:

```
new Koa({ server: 'http' }).listen();
new Koa({ server: 'http2' }).listen();
require('http').createServer(new Koa().callback()).listen();
require('http2').createServer(new Koa().callback()).listen();
```

fl0w:
@dead-horse is the update better? This would allow either `{ server: 'http' }` or `{ server: 'http2' }`, or a custom http/http2 compliant implementation `{ server: require('custom-spdy-server') }`.

I'll add tests and docs if this approach is acceptable.

dead-horse:
LGTM

jonathanong:
i'd rather keep `.listen()` minimal. people should be understanding that `app.callback()` just returns a function that you can use in any http server

fl0w:
Closing because of bloat.

Rukeith:
@fl0w What is the finally solution?
Use `const app = new Koa({ server: 'http2' })` ?
I didn't find anything about this in document

andjosh:
@Rukeith I believe this was closed in favor of using `app.callback()`, like in https://github.com/koajs/koa/pull/1063#issuecomment-330112106

felixsanz:
But... what if you want to use both http2 and http1? You need to create 2 separate instances of Koa?

fl0w:
@felixsanz yes - that would be the case regardless of this PR

PR #1065, opened 2017-09-22
docs: command is wrong for running tests

xueqingxiao:


jasonxia23:
Seems to be a legacy from v1.x.

Issue #1064, opened 2017-09-18
Don't use jest --forceExit


nickmccurdy:
This is an antipattern since it can make running the test watcher less performant and the flag is meant to be a temporary solution. The best way to fix this is by passing `app.callback()` to supertest instead of `app.listen()`. The former has a server automatically created and closed by supertest, while the latter creates an unclosed server.

likegun:
Jest will invoke `.createServer()` and `server.listen(0)` if you pass `app.callback()` to the constructor, there is no difference whether you pass `app.callback()` or `app.listen()`.

nickmccurdy:
There is no difference in how the server is requested, but the server is automatically created *and closed* when you use `app.callback()` . `app.listen()` does not automatically close the server, it stays open. We could set up after each statements in every test suite to manually close the servers, but I find it easier to just use `app.callback()`. See https://github.com/koajs/koala/pull/55/commits/31b6f42609f6c0ec20713a84a6f2ea626e9a49ae.

likegun:
You are right, supertest only close server that is self created.
[supertest close server only if server(this._server) is set](https://github.com/visionmedia/supertest/blob/199506d8dbfe0bb1434fc07c38cdcd1ab4c7c926/lib/test.js#L126)
[supertest only set this._server when it's self created server](https://github.com/visionmedia/supertest/blob/199506d8dbfe0bb1434fc07c38cdcd1ab4c7c926/lib/test.js#L59))

fl0w:
> Jest will invoke .createServer() and server.listen(0) if you pass app.callback() to the constructor.

What constructor? The test?

likegun:
@fl0w Oh sorry, the supertest function not constructor. The problem is that if we use supertest this way`request(app.listen())`, supertest will not close the koa server automatically when `request.end()` is invoked.See this [issue](https://github.com/koajs/koa/issues/1061#issuecomment-330421733)

likegun:
@nickmccurdy close it?

nickmccurdy:
Thanks! Looks like we forgot to remove the flag though, see #1071.

Issue #1061, opened 2017-09-15
Tests: koa server not closed


entropitor:
This is not really a bug report but I'm wondering why you don't close the server in your tests (e.g. you open one here: https://github.com/koajs/koa/blob/master/test/application/index.js#L22) but you never close it, as far as I can find. How can the test process terminate?

fl0w:
supertest closes the server once it's done. https://github.com/visionmedia/supertest/blob/master/lib/test.js#L126

entropitor:
Thank you. I'll check with supertest then https://github.com/visionmedia/supertest/issues/436

likegun:
Supertest only close server when server(this._server) is set
```javascript
Test.prototype.end = function(fn) {
  var self = this;
  var server = this._server;
  var end = Request.prototype.end;

  end.call(this, function(err, res) {
    if (server && server._handle) return server.close(assert);
    ...
  });

  return this;
};
```
But this._server is set only when app.address() do not exist, which means the app is created by supertest itself.
```javascript
Test.prototype.serverAddress = function(app, path) {
  var addr = app.address();
  var port;
  var protocol;

  if (!addr) this._server = app.listen(0);
  ...
};
```
So if we pass `app.listen()` to supertest, supertest will not close it.

khotey:
@likegun so If I want to close server on done I need pass app without listen() ?

fl0w:
try `supertest.agent(app.callback())`

khotey:
@fl0w works, but server don't close

fl0w:
Im not close to a computer now, but is it possible your app opens other persistent connections? Such as perhaps a database connection pool?

khotey:
@fl0w koa sever connected to mongoose server, if I add ` --forceExit` flag server close and mongo connection close to. I thought supreteset closes server connection and after server closes mongoose connection automatically without ` --forceExit` , dont it ?

fl0w:
No, you'll have to manage a graceful shutdown of mongoose yourself in app. Supertest only closes the http server (if correctly used), it has no knowledge of other lingering connections.

khotey:
@fl0w thanks!) Can you tell how it do with koa, because when supertest up server he up with mongoose and I can get mongo instance. I monitor connections in mongo by `db.serverStatus().conecctions` `--forceExit` kill node process and it automatically kill connection to mongo.

fl0w:
I do it by hooking on a teardown function on https://nodejs.org/api/http.html#http_event_close

PR #1066, opened 2017-09-22
replace request(app.listen()) with request(app.callback())

likegun:
If we use request(app.listen()), supertest won't close the server automatically. The jest exit beacause of `--forceExit` option even if there are servers opened.  [1066](https://github.com/koajs/koa/issues/1064) [1061](https://github.com/koajs/koa/issues/1061)


fl0w:
LGTM

PR #1071, opened 2017-09-26
Remove --forceExit flag for Jest

### LINKED ISSUE #1064: Don't use jest --forceExit

nickmccurdy:
Fixes #1064

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "Koa web app framework",
   "main": "lib/application.js",
   "scripts": {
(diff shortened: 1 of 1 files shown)


nickmccurdy:
Tests pass! Please review when possible, it's just a one line change.

PR #1073, opened 2017-09-29
fix: upgrade fresh to the latest version

popomore:
Relate security issue https://snyk.io/vuln/npm:fresh:20170908


popomore:
@dead-horse

PR #1075, opened 2017-09-29
feat: invalidateMiddleware

z-vr:
To be able to change middleware at run-time.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -4,6 +4,7 @@ Adam Lau <skyros@gmail.com>
 Aesop Wolf <aesopwolf@users.noreply.github.com>
 AlexeyKhristov <AlexeyKhristov@users.noreply.github.com>
 Amit Portnoy <amit.portnoy@gmail.com>
+Anton <anton@sobes.io>
 Arjun <arjun453@gmail.com>
 Avindra Goolcharan <aavindraa@gmail.com>
 Bartol Karuza <bartol.k@gmail.com>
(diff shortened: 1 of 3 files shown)


fl0w:
I'm -1 on this.

popomore:
You can modify this.middleware before run callback.

z-vr:
@popomore you can but I need to modify it from middleware. I want to update middleware dynamically from a route of my app, currently it's not possible, since the initial function to call before handling requests is always the same.

fl0w:
@z-vr how about just initialising a new app and hot swap referenced callback used by http server instead?

z-vr:
@fl0w what is your concern with this PR

dotnil:
Isn't skipping middlewares already possible by arranging `ctx.state` flags? I don't see the point of baking this in the framework itself.

fl0w:
Note that I am not part of Koa core team, I am merely expressing my opinion as a contributor. Don’t feel discouraged if you disagree with me.

I feel the PR adds unnecessary bloat to its current simplicity.

The concept of mutating middleware chain during runtime can have many side effects, in particular by introducing state, for instance if you scale your app horizontally. It should be avoided and definitely does not fit as a feature in Koa (IMO). If you want to introduce this yourself - it’s easy enough to dynamically compose a chain (or simply a new Koa app) at runtime and hotswap it - in which case Koa did its job, which is to delegate application logic to the author.

I’m inject no opinion on the actual quality of your PR, that’s not my concern at this point. It is specifically that it introduces an inelegant flow.

z-vr:
@fl0w I liked your idea regarding swapping a listener -- it's very good, thank you! Not sure what to do with the PR, probably will just close it (can someone from the Koa team confirm that it won't be useful please?)

AUTHORS
```diff
@@ -4,6 +4,7 @@ Adam Lau <skyros@gmail.com>
 Aesop Wolf <aesopwolf@users.noreply.github.com>
 AlexeyKhristov <AlexeyKhristov@users.noreply.github.com>
 Amit Portnoy <amit.portnoy@gmail.com>
+Anton <anton@sobes.io>
```
jonathanong: this should be in a separate PR

jonathanong:
yes we intentionally do not support this feature. if you really want this, what you should do is create a middleware that supports a mutate-able middleware array:

```js
const compose = require('koa-compose')

const middleware = []

app.use((ctx, next) => compose(middleware)(ctx, next)
```

you'll be able to optimize this function as much as you want

PR #1079, opened 2017-10-10
replace === with ==

rdmclin2:
1. always use `===`, since `==` may cause unexpected type coercion.
2. replace `body == null` and `len == null` with `!body` and `!len` since body and len can be undefined.

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -215,7 +215,7 @@ function respond(ctx) {
   }
 
   // status body
-  if (null === body) {
+  if (!body) {
     body = ctx.message || String(code);
     if (!res.headersSent) {
       ctx.type = 'text';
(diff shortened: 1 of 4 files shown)


Runrioter:
> always use ===, since == may cause unexpected type coercion.
Maybe it's unnecessary, if it run ok.

> 2. replace body == null and len == null with !body and !len since body and len can be undefined.

It's wrong.
1. When `var` is `undefined`,  `var == null` is the same as `!var`
2. But when `var` is `''`, your change is broken.

rdmclin2:
@Runrioter I think it's always a best practice to use === instead of ==。For `body == null` and `len == null` , I agree with you.

fl0w:
Strict comparison (`===`) may be common practice which is driven as a best practice because abstract comparison (`==`) is "complicated". However, when used properly there's benefits to abstract comparison as javascript is not statically typed.

I'm -1.

PR #1082, opened 2017-10-16
feat: Add close method to close the http server.

axetroy:
Some time you need to close connection on manual

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -46,6 +46,7 @@ module.exports = class Application extends Emitter {
     this.context = Object.create(context);
     this.request = Object.create(request);
     this.response = Object.create(response);
+    this.server = null;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


axetroy:
在一些进程退出等，与关闭数据库连接一起，关闭http端口的监听

Runrioter:
1. Coverage down.
2. All `callback &&` should be deleted.
3. Do it allow multiple call `close`?
4. Buggy, if err != null, this.server also is assigned to null. Because 
> callback will be called with an Error as its only argument if the server was not open when it was closed.

axetroy:
@Runrioter 
1. My bad code...
2. Why should be deleted?

how about refactor to this:

```javascript
  /**
   * close the http listener if http server is listening
   * @param callback
   * @return {void}
   * @api public
   */
  close(callback) {
    callback = callback || function() {};
    if (typeof callback !== 'function') {
      throw new Error(`close callback must be a function!`);
    }
    if (!this.server) {
      setImmediate(callback, null);
    } else {
      this.server.close(err => {
        if (!err) {
          this.server = null;
        }
        callback(err || null);
      });
    }
  }
```

3. Yes. allow multiple call close. It does not matter. because this.server have been assigned to ``null``, it will not apply this.server.close twice

4. if err != null, this.server also is assigned to null

```javascript
      this.server.close(err => {
        // if close server fail or throw and error. 
        // server should not be closed successfully and this.server should not assigned to 'null'
        if (!err) {
          this.server = null;
        }
        callback(err || null);
      });
```

lib/application.js
```diff
@@ -60,11 +61,30 @@ module.exports = class Application extends Emitter {
 
   listen(...args) {
     debug('listen');
-    const server = http.createServer(this.callback());
+    const server = this.server = http.createServer(this.callback());
     return server.listen(...args);
```
dead-horse: I think we don't need to add a close method because it rarely used, and we can close http server like this:

```js
const app = new Koa();
const server = app.listen();
server.close();
```

axetroy: Yep. You are right! I will close this PR later

PR #1084, opened 2017-10-18
feat: replace npm with yarn 

clarkdo:


dead-horse:
we don't use yarn, and lock versions is difficult to find out the problems in this repo.

Issue #1085, opened 2017-10-18
Update to fresh 0.5.2


dlebedynskyi:
[fresh](https://github.com/jshttp/fresh) has [Vulnerable Regular Expression issue](https://github.com/jshttp/fresh/issues/24).
https://snyk.io/vuln/npm:fresh:20170908
Current version listed in package.json is 0.5
it is advised to up update to 0.5.2 for fix.

jonathanong:
18e4fafb024366112b22084e2d0311dd5f4f8922

PR #1086, opened 2017-10-23
Update fresh to ^0.5.2 to close vulnerability

hcz:
@dlebedynskyi reports in #1085:
> fresh has Vulnerable Regular Expression issue.
https://snyk.io/vuln/npm:fresh:20170908
Current version listed in package.json is 0.5
it is advised to up update to 0.5.2 for fix.

This can be fixed by updating package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "destroy": "^1.0.3",
     "error-inject": "~1.0.0",
     "escape-html": "~1.0.1",
-    "fresh": "^0.5.0",
+    "fresh": "^0.5.2",
     "http-assert": "^1.1.0",
     "http-errors": "^1.2.8",
     "is-generator-function": "^1.0.3",


popomore:
koa use `^0.5.0` that will install 0.5.2, besides you lock the version with package-lock.json or yarn.

iyuq:
if you install before and installed `0.5.0`, then rerun `npm i`, will the `0.5.2` be installed? If not, I think it's better to update `0.5.0` to `0.5.2`. @popomore

hcz:
@popomore @iyuq 
`npm i` won't update previously installed fresh 0.5.0 to 0.5.2 (both with and without package-lock.json)

popomore:
@hcz do you remove node_modules before install?

dlebedynskyi:
this does not really impacts new koa installations without lockfiles. 

However 0.5.0 satisfies requirements. I belive that for case when lockfiles (or already installed node_modules) have 0.5.0 it will remain 0.5.0 and won't update to 0.5.2. 

One way or another appreciate looking into this guys. Update with 0.5.2 is much appreciated. At minimal it will silence `snyk dogs`

Issue #1089, opened 2017-10-27
How to handle ECONNRESET and EPIPE on streaming?


stepankuzmin:
Consider this example

```js
const fs = require('fs');
const Koa = require('koa');

const app = new Koa();

app.use(async (ctx) => {
  ctx.body = fs.createReadStream('really-large-file');
});

app.listen(4000);
```

If user aborts request I'm getting either

```shell
  Error: read ECONNRESET
      at _errnoException (util.js:1024:11)
      at TCP.onread (net.js:618:25)
```

or

```shell
  Error: write EPIPE
      at _errnoException (util.js:1024:11)
      at WriteWrap.afterWrite [as oncomplete] (net.js:870:14)
```

What is the proper way to handle this type of errors?

**P.S.** I have no errors after request aborts with `express`

```js
const fs = require('fs');
const express = require('express');

const app = express();

app.get('/', (req, res) => {
  fs.createReadStream('really-large-file').pipe(res);
});

app.listen(4000);
```

**P.P.S.** I've tried

```js
app.use(async (ctx) => {
  fs.createReadStream('really-large-file').pipe(ctx.res);
  ctx.respond = false;
});
```

But it had no effect.

dead-horse:
try [koa-onerror](https://github.com/koajs/onerror) to catch this error and handle it manually?

jonathanong:
depends on how you want to handle the error. once the headers and the stream is already sent, there's not much you can do.

GuillaumeCisco:
Ok just found out resolving this problem in my case is impossible as `event-stream` do not communicate data to the server. So the server will never know when an user refresh its tab. We have to use a websocket communication to handle this.

I discovered webpack-koa with https://github.com/webpack-contrib/webpack-hot-client which implements a websocket approch and should be better than using koa-error. I will try that ;)

shellscape:
@GuillaumeCisco just thought I would mention: I wrote webpack-hot-client, and was researching the same ECONNRESET error in a separate project. your comment reminded me that I've already solved this problem in the past. Cheers.

rvanmil:
I know this is an old issue but in case someone finds it and is unable to get things to work properly (koa-onerror did not work for me), here's how I got this to work (koa 2.7.0).
[...]

PR #1093, opened 2017-11-05
added onFinished() handler

RoflanDaniil:
issue https://github.com/koajs/koa/issues/1089
for example
``` javascript
const Koa = require('koa');
const fs = require('fs');

const app = new Koa()

app.onFinished = (err, res) => {
  console.log(`err = ${err}`);
}

app.use(ctx => {
  ctx.body = fs.createReadStream('./large');
}).listen(3000)
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ test.js
 coverage
 npm-debug.log
 package-lock.json
+yarn.lock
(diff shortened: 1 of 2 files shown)


fl0w:
Why is this better than simply just using `ctx.onerror`? Sorry.

jonathanong:
yes, what's the use case for this? this feature should add an additional "onFinished" handler, not replace the default, but i don't even know a use case for that

dead-horse:
close for now, feel free to reopen it if necessary

PR #1100, opened 2017-11-22
Update mgol's name in AUTHORS, add .mailmap

mgol:
In this way mgol's past contribution is mapped correctly.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,1 @@
+Michał Gołębiowski-Owczarek <m.goleb@gmail.com>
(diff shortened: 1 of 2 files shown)


mgol:
Test failures don't seem related to this PR.

dead-horse:
I think this AUTHORS list is generated from git log (https://github.com/koajs/koa/blob/master/package.json#L11)

fl0w:
This list was generated using `git log ...`. So even if this gets merged as requested it'll be a recurring issue whenever AUTHORS is updated.

https://github.com/koajs/koa/blob/841844ee2fd0cf91478b5ed8d2f401e8b97eb4a2/package.json#L11

mgol:
@fl0w The command you pasted respects `.mailmap` per https://git-scm.com/docs/pretty-formats:
> %aN': author name (respecting .mailmap, see git-shortlog[1] or git-blame[1])

So this shouldn't be an issue (and I see it taken into account locally).

fl0w:
Oh, today I learned. Then I'm +1 :)

PR #1102, opened 2017-11-28
Add app.start method - a listen promise

ipotuzhnov:
Hi, I've been interested in the project for quite some time. I've started using it on my new project and I had an issue with catching server errors on boot up, because `app.listen()` returns an `http.Server` that is in fact event emitter. Here is an example:
```js
const Koa = require('koa')
async function start () {
  const app = new Koa()
  // when there is a server listening on port 3000 throws unhandled 'error' event
  await new Promise((rs, rj) => app.listen(3000, err => err ? rj(err) : rs()))
}
start().catch(err => console.error(err))
```
My solution was to use `app.callback()`
```js
const http = require('http')
const Koa = require('koa')
async function start () {
  const app = new Koa()
  const server = http.createServer(app.callback())
  await Promise.race([
    new Promise((rs, rj) => server.once('error', rj)),
    new Promise((rs, rj) => server.listen(3000, err => err ? rj(err) : rs()))
  ])
}
start().catch(err => console.error(err))
```
So I thought it'd be nice to have a method on Koa application that guarantees to return a listening server.
```js
const server = await app.start(3000)
```

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -72,11 +72,18 @@ module.exports = class Application extends Emitter {
    * @api public
    */
   start(...args) {
-    return new Promise(resolve => {
+    return new Promise((resolve, reject) => {
       const server = http.createServer(this.callback());
-      server.listen(...args);
 
-      server.once('listening', () => resolve(server));
+      function onerror(err) { reject(err); }
+
+      server.once('error', onerror);
+      server.once('listening', () => {
+        server.removeListener('error', onerror);
+        resolve(server);
+      });
+
+      server.listen(...args);
     });
   }
 
--- a/lib/application.js
+++ b/lib/application.js
@@ -65,6 +65,22 @@ module.exports = class Application extends Emitter {
   }
 
   /**
+   * Eventually return a listening http server.
+   *
+   * @param {Mixed} ...
+   * @return {Promise<Server>}
+   * @api public
+   */
+  start(...args) {
+    return new Promise(resolve => {
+      const server = http.createServer(this.callback());
+      server.listen(...args);
+
+      server.once('listening', () => resolve(server));
+    });
+  }
+
+  /**
    * Return JSON representation.
    * We only bother showing settings.
    *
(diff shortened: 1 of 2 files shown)


fl0w:
This feels sort of opinionated, and it's a high level abstraction on top of `.listen` which has previously been denied expansion (argument was; keep `.listen` minimal, if you need more power use `.callback`).

What if I wanted to http2 here?

I'm (subjectively) -1.
Maybe as a separate package?

ipotuzhnov:
@fl0w I see your point. Didn't know that you've already had this discussion, could you provide a link please?
My reasoning was that it's not that trivial to promisify an `EventEmitter` and I thought it'd be nice to have out of the box.

In case of http2, I guess you'd want something like
```js
// index.js - usage example
const http2 = require('http2')
const Koa = require('koa')
const app = new Koa({ createServer: http2.createServer })
const server = app.listen(/* http2.createServer arguments */)
// or
const server = await app.start(/* http2.createServer arguments */)

// lib/application.js - koa
class Application extends Emitter {
  constructor (options) {
    this.createServer = options.createServer || http.createServer
  }
  listen (...args) {
    debug('listen');
    const server = this.createServer(this.callback());
    return server.listen(...args);
  }
} 
```
It should work fine since `(http|https|http2).createServer` and `http2.createSecureServer` aren't methods on a class and more like static methods.
Bonus point: it will make unit-testing `listen` and `start` functions easier.

coreyfarrell:
I agree with keeping koa itself to a minimum, it's hard to get rid of a feature once it's added.

@ipotuzhnov, you've given me the motivation to finally create a generic module to get a promise from an emitter.  I published v0.1.0 to npm, will probably bump to v1.0.0 once I'm sure the API is what it should be.

https://github.com/cfware/promisify-emitter

fl0w:
@ipotuzhnov sure, https://github.com/koajs/koa/pull/1063#issuecomment-333333559

danopia:
I know this ticket is old, just adding for future visitors that a rendition of EventEmitter promises is in nodejs nowadays, as of v11.13.0:

```js
const { once } = require('events');
```

```js
    const promise = once(http, 'listening');
    http.listen(...args);
    await promise;
```

PR #1103, opened 2017-12-06
node version consistent for readme

JasonBoy:
since `package.json` require node `^7.10.1`, the `Readme.md` may also need to sync with that.


nickmccurdy:
Thanks! Now the docs are more consistent as I suggested. However, shouldn't be also change the `"node": "^4.8.4 || ^6.10.1 || ^7.10.1 || >= 8.1.4"` engines to `"node": "^7.10.1 || >= 8.1.4"` if we're going to do this?

JasonBoy:
@nickmccurdy updated in package.json, at least they are consistent for latest version🤣

fl0w:
This PR would require a semver major bump unless we want to break all node 4 and 6 users.

node.js 4 is under maintenance until April 2018 (if I recall correctly), there's no benefit to omit those users unless restrictions actively prohibits Koa development, in which case a semver major bump is a reasonable path for further development.

I'm -1.

nickmccurdy:
You're right that Node 4 is still in active LTS and this should be a breaking change. 

If all I dad was `require('koa')` (ignoring async/await and other syntax features), would this work on Node 4 and 6? If not I would still be in favor of merging this to provide more accurate node engines for users on older versions of Node, otherwise maybe we should change the docs instead of the engines.

iyuq:
aysnc/await will not work on node.js 4 or 6, but user can still use normal function in koa2. Since koa itself needs node.js >= 6.x, I agree with remove 4.x, but I think 6.x must remain.

fl0w:
There's no reason to deprecate an entire LTS version right now and to force a Koa semver major bump. I also think that it is reasonable to drop 6 as well and just support node 8+ for Koa 3+. That would also legitimise a quality of life refactor using all the shiny syntax as long as it doesn't affect performance negatively.

The package.json change feels premature.

nickmccurdy:
I think our two best options are to either add full support for Node 4 and 6 (which could be as simple as adding Babel to a `prepublish` or `prepare` script) or promptly release a breaking change that removes Node < 6 engines. I do not find this premature, I honestly believe the engines are innacurate and allowing silent installation failures is worse than releasing a breaking change and major version.

JasonBoy:
On the koa running requirements perspective, the engines may still better to stay with the min runnable version of every major node version

nickmccurdy:
@fl0w Maybe I misinterpreted, were you suggesting to alternatively pause development on Koa 2 and wait until April to make the engines more restrictive after Node 4 is out of maintenance mode?

jonathanong:
yeah, we can't merge the `package.json` change without bumping major version. i can merge this if you remove it

JasonBoy:
bad PR, close

PR #1105, opened 2017-12-11
Use eslint-config-koa

nickmccurdy:
https://github.com/koajs/eslint-config-koa


nickmccurdy:
It works well with the current versions of eslint plugins, but the latest versions of `eslint-plugin-standard` and its `peerDependencies` are also supported by `eslint-config-koa`. Do we want to merge later and shelve the package updates for later or would you like me to do it in this pull request?

PR #1108, opened 2017-12-18
Avoid generating package locks instead of ignoring them

nickmccurdy:
Fixes https://github.com/npm/npm/issues/17168. Package locks should not be ignored, they should be disabled, otherwise `npm version` and other commands may fail.


nickmccurdy:
See #1109

.npmrc
```diff
@@ -0,0 +1 @@
+package-lock=false
```
fl0w: Nit: Add a newline

fl0w:
LGTM

.npmrc
nickmccurdy: Whoops, I meant to do this before. Done.

PR #1113, opened 2017-12-22
Error handling: try to stringify non-error throw if it is an object

Alexsey:
In case when object were thrown error message is really non helpful - you just see `non-error thrown: [object Object]` and stack trace is leads to Koa source. If someone throws an object - it is his fault that no stack trace would be provided, but seen the object itself is already a good start. A specially when it was not your code but some of the modules that you used

With this PR thrown objects would be converted to JSON

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -5,6 +5,7 @@
  * Module dependencies.
  */
 
+const util = require('util');
 const createError = require('http-errors');
 const httpAssert = require('http-assert');
 const delegate = require('delegates');
@@ -105,7 +106,7 @@ const proto = module.exports = {
     // to node-style callbacks.
     if (null == err) return;
 
-    if (!(err instanceof Error)) err = new Error(`non-error thrown: ${err}`);
+    if (!(err instanceof Error)) err = new Error(util.format('non-error thrown: %j', err));
 
     let headerSent = false;
     if (this.headerSent || !this.writable) {
--- a/test/context/onerror.js
+++ b/test/context/onerror.js
@@ -185,5 +185,23 @@ describe('ctx.onerror(err)', () => {
 
       assert.equal(removed, 2);
     });
+
+    it('should stringify error if it is an object', done => {
+      const app = new Koa();
+
+      app.on('error', err => {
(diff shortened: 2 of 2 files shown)


lib/context.js
```diff
@@ -105,7 +105,14 @@ const proto = module.exports = {
     // to node-style callbacks.
     if (null == err) return;
 
-    if (!(err instanceof Error)) err = new Error(`non-error thrown: ${err}`);
+    if (!(err instanceof Error)) {
+      try {
+        if (Object.prototype.toString.call(err) === '[object Object]') {
+          err = JSON.stringify(err);
```
dead-horse: err must convert to Error instance here, BTW can we use `new Error(util.format('non-error thrown: %j',  err)` here?

Alexsey: What is the reason to convert it to an error if on the next line it would be converted to a string with `${err}`? And generally there is no sense in the conversion because stack trace is already lost - new Error would have a stack trace to Koa internals which is not helpful :( Although usage of `util.format` may be a good idea - I'll try to play with it to make a code nicer

dead-horse: it will emit in `app` [here](https://github.com/koajs/koa/pull/1113/files/7a2889efccde3df6035390f7d1886c09bad21720#diff-d911385a221c0b7502eabe1fdac9d209R123), we must keep `err` to be an instance of Error so it won't break anyone's code.

Alexsey: Hm. But it would already always become an error [two lines lower](https://github.com/koajs/koa/pull/1113/files/7a2889efccde3df6035390f7d1886c09bad21720#diff-d911385a221c0b7502eabe1fdac9d209R114)

Alexsey: You where right about `util.format` - it really covers all this cases without `if`s and additional code! Consider a new version now

Alexsey:
There is still one case that `util.format` doesn't handle well - `Symbol`. It would convert it to `undefined` and this is frustrating. I think I'll make a special treatment for `Symbol` case later today

dead-horse:
this is enough, throw a literal is an edge case, and throw a symbol is meaningless.

PR #1115, opened 2017-12-27
response.redirect: extract special-cased back, fixes #904

### LINKED ISSUE #904: Drop special 'back' case from response.redirect?
https://github.com/koajs/koa/blob/18dbf5f59c9f3e487e2b49c6dac483d1bed88974/lib/response.js#L254

expressjs/express#2692 is stuck. I thought, Koa is the future, so you may be more open to breaking changes. Sorry if I'm wrong.

fl0w:
The simple solution to drop special-cased `'back'` in `response.redirect`.
I can't label but should be labeled as version-major.

This PR **does not** make use of Symbol as purposed in #904.

**Edit**
If this solution is acceptable, a deprecation should be added to Koa 2 on `'back'` use.


fl0w:
I'm assuming this is a travis-cache thingie that fails, can't reproduce it locally.

```bash
$ node -v
v9.3.0
$ npm -v
5.6.0
$ npm view koa version
2.4.1
$ npm t

# abbreviated jest output

Test Suites: 66 passed, 66 total
Tests:       298 passed, 298 total
Snapshots:   0 total
Time:        2.265s
Ran all test suites.
```

jonathanong:
👍 

let's add this as a feature + deprecation message, then a separate PR to remove it in the next major version

fl0w:
@jonathanong updated, and should be safe as a semver-minor merge. Just take a quick look at the deprecation message before doing so.

lib/response.js
```diff
@@ -258,7 +259,11 @@ module.exports = {
 
   redirect(url, alt) {
     // location
-    if ('back' == url) url = this.ctx.get('Referrer') || alt || '/';
+    if ('back' == url) {
+      deprecate('Special-cased string "back" through redirect will be removed in v3,' +
+                'consider migrating usage to ctx.back() instead.');
```
ricordisamoa: Missing space between `v3,` and `consider`?

kevinpeno:
@jonathanong should be gtg. Codecov is bitching, but the related missed line is not related to the changes made in this PR. It is likely the methods added to this PR pushed the file over its limits. I think we should ignore it for this PR.

iwanofski:
I couldn't see this update because this was done om my "stolen" account :) Great to see this merged.

PR #1119, opened 2018-01-05
Add support for custom statuses

### LINKED ISSUE #1042: Allow non-standard http status codes to be used
Koa throws when a status code not defined in the `statuses` package is used in response:

https://github.com/koajs/koa/blob/78832ff6c63fdd93ad9a49f31ba2edd8cbe8adee/lib/response.js#L84

However, the HTTP RFC does not prevent custom status codes to be used:
http://greenbytes.de/tech/webdav/draft-ietf-httpbis-p2-semantics-25.html#rfc.section.6.p.2

I can understand why status code validation might be useful for people just starting with http protocols, but for people who actually need to use custom codes, this is very limiting.

### Proposal

There are several options that the Koa team could choose to remediate the problem:

- Do not throw an error and just emit a warning
- Keep throwing but add a configuration flag to disable the check (ie. `koa.validateStatus = false`)
- Allow adding custom http status codes to Koa so that it accepts them (ie. `koa.AddStatus(440, 'Invalid Chicken')`)

I personally like the second and third approach. Currently the third approach is possible, albeit in a hacky way:

```js
import statuses from 'statuses'

// Add a hypothetical "invalid chicken" response code
statuses[440] = 'Invalid chicken'
statuses.codes.push(440)
statuses['Invalid chicken'] = 440
```

CodingPower472:
Fixes #1042


lib/application.js
```diff
@@ -46,6 +46,18 @@ module.exports = class Application extends Emitter {
...
+
+  /**
+   * Add custom status code
+   * @param {Number} code
+   * @param {String} description
+   * @api public
+   */
+
+  addStatus(code, description) {
+    this.customStatuses[code] = description;
```
jonathanong: only issue is naming. i'd prefer this to be called `customStatusCodes`. what does everyone else think?

jonathanong: other than that, i'm good with this PR

lib/application.js
```diff
@@ -46,6 +46,18 @@ module.exports = class Application extends Emitter {
     this.context = Object.create(context);
     this.request = Object.create(request);
     this.response = Object.create(response);
+    this.customStatuses = {};
```
fl0w: Nit: `Object.create(null)`

lib/application.js
```diff
@@ -46,6 +46,18 @@ module.exports = class Application extends Emitter {
...
+  }
+
+  /**
+   * Add custom status code
```
MarkHerhold: All other comments have a period after the description text.

lib/application.js
MarkHerhold: Not that my opinion matters, but personally I'd prefer `addCustomStatusCode`, erring on the side of a more descriptive name

matttowerssonos:
Any updates on merging this fix?  We've run in to the same problem and trying to monkey patch has introduced other issues.

fl0w:
**Highly subjective and not a reason for this not being merged** 
I dislike this solution. I would much rather see [statuses](https://github.com/jshttp/statuses) to provide this and Koa simply delegating. Not sure this is of interest to jshttp though.

lib/response.js
```diff
@@ -81,11 +81,11 @@ module.exports = {
 
   set status(code) {
     assert('number' == typeof code, 'status code must be a number');
-    assert(statuses[code], `invalid status code: ${code}`);
+    assert(statuses[code] || this.customStatuses[code], `invalid status code: ${code}`);
     assert(!this.res.headersSent, 'headers have already been sent');
     this._explicitStatus = true;
     this.res.statusCode = code;
-    if (this.req.httpVersionMajor < 2) this.res.statusMessage = statuses[code];
+    if (this.req.httpVersionMajor < 2) this.res.statusMessage = statuses[code] || this.customStatuses[code];
```
chittolina: If I get you correctly, you meant to use the custom status code instead of the default status code. Right? So, actually the logic would be: 
```
this.customStatuses[code] || statuses[code];
```
Then, if you have no custom status codes the default status will be used instead.

m-a-r-c-e-l-i-n-o:
@fl0w Hey man, I don't understand why this has not been addressed yet. The way I see it, KOA has to either provide a mechanism or go with a different status module because as one of the maintainers of `statuses` says:
screen shot 2018-11-05 at 4 52 28 pm

They are leaning towards the former (being a list of known status codes), and it sounds like (and is apparent by the code) that KOA is leaning towards the latter. So if KOA is leaning towards the latter, it should provide a mechanism instead of imposing on the `statuses` library's preferences.

Issue #1114, opened 2017-12-24
koa 3.0 roadmap


dead-horse:
There are lots of PRs are blocked due to breaking change, and maybe it's time to make a plan for koa 3.0 now, if anyone has other suggestion, please post in this issue. :)

- https://github.com/koajs/koa/issues/1044
- https://github.com/koajs/koa/issues/1008
- https://github.com/koajs/koa/issues/998
- https://github.com/koajs/koa/issues/904
- https://github.com/koajs/koa/pull/1011

fl0w:
Is it to early to create a v3 branch? I have some spare time coming up and wouldn’t mind doing some work.

dead-horse:
All these pending PRs are not so emergency to me, so if anyone has any thoughts or some feature / refactor required, post in the issue to help us decide when to start create v3 branch. :)

fl0w:
How about officially dropping support for < node.js 8?

iyuq:
`koa-compose` 可以用 `async/await` 重写下

iyuq:
totally rewrite using `Typescript`

fl0w:
I'm personally against introducing TypeScript, adding external typings should be easy enough for those in need.

marswong:
> How about officially dropping support for < node.js 8?

@fl0w IMO, engines support should always follow the [Node.js release schedule](https://github.com/nodejs/Release#release-schedule) :)

marswong:
> totally rewrite using TypeScript

like what @fl0w said above, just one more thing:
`koa` is quite vary from `egg`、`express` and any other **framework**. It's just a pure http request middleware close to `Node.js` as you can see. So we can just feel free to write in pure JavaScript, follow the traditional convention and trust `Node.js` :)

jonathanong:
if you have any breaking changes you'd like to introduce, please state them now :)

fl0w:
Let's announce EOL for Koa v1.

zhangxiang958:
How about support HTTP/2 ? https://nodejs.org/api/http2.html

hiroqn:
How about publish `koa` and official middleware module under `@koajs/` ?

likegun:
I think [#1044](https://github.com/koajs/koa/issues/1044) is already solved by this merged [commit](https://github.com/koajs/koa/commit/3c23aa5b74d19352e0a78d135df5ed6979d4e242)

tiendq:
[...]

PR #1131, opened 2018-02-02
do not override already set content-type, fixes #1120

### LINKED ISSUE #1120: Koa always overwrite `Content-Type` for JSON responses
```js
const Koa = require('koa');
const app = new Koa();
app.use(ctx => {
  ctx.set('Content-Type', 'application/vnd.myapi.v1+json');
  ctx.body = {message: 'hello'};
});
app.listen(3000);
```

When running this code I aways get `Content-Type: application/json; charset=utf-8`.

I can make it work if I write the header after setting the body. It only affects `json` responses. I'm not sure if that's a bug, but I couldn't find any information about that behavior in documentation.

fl0w:
semver major, #1114

This could be done as semver minor - if an application setting is addd, e.g. `app.responsOverride = true` or something.

### DIFF
--- a/test/response/body.js
+++ b/test/response/body.js
@@ -15,14 +15,14 @@ describe('res.body=', () => {
     });
 
     describe('when body is an object', () => {
-      it('should override as json', () => {
+      it('should not override as json', () => {
         const res = response();
 
         res.body = '<em>hey</em>';
         assert.equal('text/html; charset=utf-8', res.header['content-type']);
 
         res.body = { foo: 'bar' };
-        assert.equal('application/json; charset=utf-8', res.header['content-type']);
+        assert.equal('text/html; charset=utf-8', res.header['content-type']);
       });
     });
 
--- a/lib/response.js
+++ b/lib/response.js
@@ -178,7 +178,7 @@ module.exports = {
 
     // json
     this.remove('Content-Length');
-    this.type = 'json';
+    if (setType) this.type = 'json';
   },
 
   /**


jonathanong:
this might break apps if someone is relying on this behavior, so i would wait for semver major.

jonathanong:
unless we only set it to json if the type is not json-like

fl0w:
Yes, definitely semver-major.
Though with this PR, #1137, #1115 and a few more a v3 is already justified (IMO).

It would be nice to drop a v3 alpha, call out EOL for v1 (soon-ish?) and break the stigma of having to rewrite Koa to justify a major bump and add proper deprecations to v2 and v1.

niftylettuce:
I agree @fl0w

jkomyno:
Any news about this PR? @jonathanong

3imed-jaberi (approved):

jkomyno:
Hi, it's been two years, any update on this PR?
@jonathanong @3imed-jaberi

test/response/body.js
```diff
@@ -15,14 +15,14 @@ describe('res.body=', () => {
...
-      it('should override as json', () => {
+      it('should not override as json', () => {
         const res = response();
 
         res.body = '<em>hey</em>';
         assert.equal('text/html; charset=utf-8', res.header['content-type']);
 
         res.body = { foo: 'bar' };
-        assert.equal('application/json; charset=utf-8', res.header['content-type']);
+        assert.equal('text/html; charset=utf-8', res.header['content-type']);
```
jonathanong, line 25: this test case is already updated, so this feature must already be live

PR #1134, opened 2018-02-05
[Fix] TypeError: ctx.onerror is not a function

und3fined:
fuk js scope

`const onerror = err => ctx.onerror(err);`
-> `ctx.onerror` is `undefined` WTF????


jonathanong:
this actually breaks everything because `ctx` is no longer `this` of `onerror`

PR #1136, opened 2018-02-06
feat: add support for breakMw

killagu:
move _respond from application to response.

add support for break middleware chain.


killagu:
@dead-horse

maticrivo:
@killagu whats a good use case for this?

killagu:
@maticrivo  proxy middleware will transfer the request in some case.

the response should immediately end,and other middlewares should not continue.

PR #1078, opened 2017-10-09
response body setter checks headersSent

likegun:
related issue: [843](https://github.com/koajs/koa/issues/843)
related pr: [1060](https://github.com/koajs/koa/pull/1060)

PR #1137, opened 2018-02-07
feat: ignore set/remove header/status when header sent

dead-horse:
from https://github.com/koajs/koa/pull/1078

we should not throw headers sent error to make `ctx.responed` and `ctx.flushHeaders` more safety.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -80,9 +80,10 @@ module.exports = {
    */
 
   set status(code) {
+    if (this.headerSent) return;
+
     assert('number' == typeof code, 'status code must be a number');
     assert(statuses[code], `invalid status code: ${code}`);
-    assert(!this.res.headersSent, 'headers have already been sent');
     this._explicitStatus = true;
     this.res.statusCode = code;
     if (this.req.httpVersionMajor < 2) this.res.statusMessage = statuses[code];
@@ -132,8 +133,6 @@ module.exports = {
   set body(val) {
     const original = this._body;
     this._body = val;
-
-    if (this.res.headersSent) return;
 
     // no content
     if (null == val) {
@@ -233,6 +232,8 @@ module.exports = {
    */
 
   vary(field) {
+    if (this.headerSent) return;
+
     vary(this.res, field);
   },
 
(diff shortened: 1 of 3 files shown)


lib/response.js
```diff
@@ -133,8 +134,6 @@ module.exports = {
     const original = this._body;
     this._body = val;
 
-    if (this.res.headersSent) return;
```
dead-horse: we shouldn't return here because there are some specific logic for streams below.

fengmk2:
Should patch this feature into koa v1

dead-horse:
I'd like to consider it as a minor version change because it only avoid throw exception when set header after header sent, which bothered developers using 1.x and 2.x. And it is harmless if you don't break koa's response handling by `ctx.flushHeader` or `ctx.responed`.

lib/response.js
```diff
@@ -80,9 +80,10 @@ module.exports = {
    */
 
   set status(code) {
+    if (this.headerSent) return;
+
     assert('number' == typeof code, 'status code must be a number');
     assert(statuses[code], `invalid status code: ${code}`);
-    assert(!this.res.headersSent, 'headers have already been sent');
```
fl0w: I understand my opinion wasn't asked for, but I'm considering if this could be a breaking change.
Is there any reason a user would rely on something like:

```js
try {
  ctx.status = 201
} catch (err) {
  if (err.message.match(/headers have already been sent/) {
    // ...
  }
}
```

dead-horse: yes, it is hard to make this decision, but even in this situation, you can do nothing but log this error because the http header has already sent.

dead-horse:
Any suggestions? Or I'll merge as a minor change and port it to 1.x.

noinkling:
This error can be a very useful indicator that there's a flaw in my app's logic somewhere, therefore I'm a bit wary of this change because now it will be harder to detect those sorts of bugs. Is there any way we could have the best of both worlds?

dead-horse:
@noinkling if you don't use `ctx.responed` or `ctx.flushHeader`, there is nothing to worry about.

if you do use these methods, you should take care of them and avoid to get into this situation.

noinkling:
In my case it's libraries that are doing it (e.g. wrapped Express middlewares), where I shouldn't be trying to mess with the response later in the chain. But I see your point that it shouldn't be a problem under normal usage, so I withdraw my comment.

PR #1141, opened 2018-02-09
run coverage using --runInBand, fixes #1109

### LINKED ISSUE #1109: Intermittent build failures on Node 8
There's a daily cron job that's been failing for a while with a syntax error on Node 8 that appears to be intermittent. Here's an example of two cron builds using the same commit, only one fails. I'm guessing it's a bug in Jest's transpilation and/or cache.

https://travis-ci.org/koajs/koa/builds/317786352
https://travis-ci.org/koajs/koa/builds/318262778

```
> jest "--coverage"

 FAIL  test/context/cookies.js
  ● Test suite failed to run

    /home/travis/build/koajs/koa/lib/response.js:7
    }});
      ^
   
    SyntaxError: Unexpected token )
     
      at ScriptTransformer._transformAndBuildScript (node_modules/jest-runtime/build/ScriptTransformer.js:289:17)
      at Object.<anonymous> (lib/application.js:11:68)
      at Object.<anonymous> (test/context/cookies.js:6:13)
```

fl0w:
Ref https://github.com/facebook/jest/issues/1874
Fixes https://github.com/koajs/koa/issues/1109

The gist: Jest seems to be writing to cache from two different child processes. It's slower, but doesn't really affect Koa.

I'm expecting this to solve flaky travis tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "lib/application.js",
   "scripts": {
     "test": "jest",
-    "test-cov": "npm run test -- --coverage",
+    "test-cov": "jest --coverage --runInBand --forceExit",
     "lint": "eslint benchmarks lib test",
     "bench": "make -C benchmarks",
     "authors": "git log --format='%aN <%aE>' | sort -u > AUTHORS"


nickmccurdy:
Nice! It seems like the cron job has already been passing, does this fix some existing problem I'm not noticing or is this is a proactive change?

fl0w:
It’s a race condition apparently, and so until that issue is closed I’m thinking it’s safe to assume to use the fallback for now.

PR #1148, opened 2018-02-15
chore: update istanbul to 0.4.5

abouthiroppy:
Currently, istanbul is 0.4.0 and has minimatch@0.3.0.
minimatch@0.3.0 should be updated.

as you can see
```
npm WARN deprecated minimatch@0.3.0: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue
```


fl0w:
Actually, I think istanbul can be removed since Koa depends on jest which has coverage built-in.

PR #1153, opened 2018-02-27
Allow to app.use multiple middlewares at once

rocketspacer:
Usage
```
app.use(foo, bar, baz)
app.use([foo], [bar, baz])
```

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -96,21 +96,25 @@ module.exports = class Application extends Emitter {
    *
    * Old-style middleware will be converted.
    *
-   * @param {Function} fn
+   * @param {Array} funcs
    * @return {Application} self
    * @api public
    */
 
-  use(fn) {
-    if (typeof fn !== 'function') throw new TypeError('middleware must be a function!');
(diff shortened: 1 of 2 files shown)


jonathanong:
not opposed to this, but it needs tests

rocketspacer:
@jonathanong added

rocketspacer:
@jonathanong any news?

fl0w:
I think this is better dealt by koa-compose. In my opinion it’s just better to let compose handle the complexity of flattening (and other resolves?), keep `Application.use` simple. I’m in favor of just removing the guards and pass whatever’s in `middleware` to compose with a flattening PR merged.

jonathanong:
@fl0w is right, we should be doing this in `koa-compose`

3imed-jaberi (commented):
I agree with @miwnwski  (@fl0w) ✔️ +1

niftylettuce:
v4.2.0 released of koa-compose @ https://github.com/koajs/compose/releases/tag/v4.2.0 per https://github.com/koajs/compose/pull/96

PR #1160, opened 2018-03-16
Record the time spent on each middleware

zhoulinxiang:
Use this data for performance optimization


fl0w:
I'm afraid this flirts with feature bloat. It would serve better as a separate package.

jonathanong:
yes, this itself is middleware or a decorator ;)

PR #1161, opened 2018-03-20
use app.context.state in koa as app.locals in express

bkisser:
Once set, the value of app.context.state properties persist throughout the life of the application, in contrast with ctx.state properties that are valid only for the lifetime of the request.


fl0w:
This PR would break allot of applications. However, I am against this because it doesn't really solve an issue. By [extending context prototype](http://koajs.com/#app-context) via app you can achieve expected behaviour already.

```js
app.context.persistent = ... 
app.use(ctx => {
  ctx.body = ctx.persistent
})
```

Unless I've misunderstood the use case this PR only convolutes `ctx.state`.

PR #1170, opened 2018-04-10
Fix tests

BridgeAR:
This came up in [CITGM](https://github.com/nodejs/citgm) while improving `assert.throws` in Node.js. See https://github.com/nodejs/node/pull/19867 for more information.

The current tests never checked for the error message and instead, the error message would have been returned, just the same as if there would be no second argument at all.

This makes sure the error message is actually tested for.

Since I just looked at the code, I also added another commit to remove all `assert.doesNotThrow` calls. Those are always unnecessary. They do not provide any benefit since in case of an error it will just catch the error and rethrow.


jonathanong:
👍 thanks

PR #1195, opened 2018-05-24
better demonstrate middleware flow

jasonmacgowan:
The way this example is written makes it seem like middleware can't act on the context changes of other middleware.  This updates the logger middleware to use the calculated response time value from the response time middleware instead of calculating it itself.

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -78,6 +78,14 @@ app.listen(3000);
 const Koa = require('koa');
 const app = new Koa();
 
+// logger
+
+app.use(async (ctx, next) => {
+  await next();
+  const rt = ctx.response.get('X-Response-Time');
+  console.log(`${ctx.method} ${ctx.url} - ${rt}`);
+});
+
 // x-response-time
 
 app.use(async (ctx, next) => {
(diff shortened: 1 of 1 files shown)


fl0w:
I'm +/-0 on this.

doug-wade:
I get where you're headed on this, and I dig it, but wouldn't this make the `X-Response-Time` ever so slightly too short?

docs/api/index.md
```diff
@@ -78,22 +78,21 @@ app.listen(3000);
...
-// x-response-time
+// logger
 
 app.use(async (ctx, next) => {
-  const start = Date.now();
   await next();
-  const ms = Date.now() - start;
-  ctx.set('X-Response-Time', `${ms}ms`);
+  const rt = ctx.response.get('X-Response-Time');
+  console.log(`${ctx.method} ${ctx.url} - ${rt}`);
```
dead-horse: it's not good here because `rt` is a string

jasonmacgowan: @dead-horse I'm not sure what you mean; can you elaborate please?  Thanks!

dead-horse: never mind, my misunderstanding.

PR #1198, opened 2018-05-27
fix: use non deprecated custom inspect

BridgeAR:
Custom inspection with the `inspect` property is deprecated and will
not work in Node.js 11 anymore. This fixes it by using the custom
inspect symbol where existent and falls back to the old style in case
it does not exist.

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -46,6 +46,9 @@ module.exports = class Application extends Emitter {
     this.context = Object.create(context);
     this.request = Object.create(request);
     this.response = Object.create(response);
+    if (util.inspect.custom) {
+      this[util.inspect.custom] = this.inspect;
+    }
   }
 
   /**
--- a/lib/response.js
+++ b/lib/response.js
@@ -18,6 +18,7 @@ const assert = require('assert');
 const extname = require('path').extname;
 const vary = require('vary');
 const only = require('only');
+const util = require('util');
 
 /**
  * Prototype.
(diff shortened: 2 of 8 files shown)


lib/response.js
```diff
@@ -545,3 +546,18 @@ module.exports = {
...
+ * Custom inspection implementation for newer Node.js versions.
+ *
+ * @return {Object}
+ * @api public
+ */
+if (util.inspect.custom) {
+  module.exports[util.inspect.custom] = function() {
+    if (!this.res) return;
+    const o = this.toJSON();
+    o.body = this.body;
```
dead-horse: can we move this line to `this.toJSON()` ?

BridgeAR:
I just pushed another commit to simplify it. That way it is always aligned, no matter if there are any changes to the actual inspection function.

lib/application.js
```diff
@@ -23,6 +23,7 @@ const http = require('http');
 const only = require('only');
 const convert = require('koa-convert');
 const deprecate = require('depd')('koa');
+const util = require('util');
```
dead-horse:
```
    /home/travis/build/koajs/koa/lib/application.js: Duplicate declaration "util"
        24 | const convert = require('koa-convert');
        25 | const deprecate = require('depd')('koa');
      > 26 | const util = require('util');
```

please fix this and good to merge. 👍

BridgeAR:
Rebased and force pushed.

BridgeAR:
Ping @dead-horse

fl0w:
Should we backport this to 1.x as well perhaps?

PR #1199, opened 2018-05-30
Throw a TypeError instead of a AssertionError

initial-wu:
Cuz AssertionError is not reading-friendly, you could not directly get where the error is emitted.
image
As TypeError seems better.
image


doug-wade:
Seems like an `instanceof` check should throw a `TypeError` to me.

initial-wu:
@dead-horse the test case has been fixed

PR #1204, opened 2018-06-08
Throw clearer errors.

initial-wu:
For the same reason with [#1199](https://github.com/koajs/koa/pull/1199), AssertionError is not reading-friendly.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -14,7 +14,7 @@ const escape = require('escape-html');
 const typeis = require('type-is').is;
 const statuses = require('statuses');
 const destroy = require('destroy');
-const assert = require('assert');
+const util = require('util');
 const extname = require('path').extname;
 const vary = require('vary');
 const only = require('only');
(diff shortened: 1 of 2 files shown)


fl0w:
Similarly to your [previous PR](#1199) to I'd like to raise awareness to the fact that changing error messages and types could easily be considered as breaking change.

initial-wu:
Agree

lib/response.js
```diff
@@ -82,8 +82,8 @@ module.exports = {
   set status(code) {
     if (this.headerSent) return;
 
-    assert('number' == typeof code, 'status code must be a number');
-    assert(statuses[code], `invalid status code: ${code}`);
+    if ('number' !== typeof code) throw new TypeError(util.format('non-numeric status code: %j', code));
```
jedahan: why util.format instead of string interpolation?

initial-wu: There is no official reason to reject string interpolation. The results of the followed forms are same.
```javascript
new TypeError(util.format('non-numeric status code: %j', code));
new TypeError(`non-numeric status code: ${JSON.stringify(code)}`);
```
But I'd like to consider `util.format` as a pattern cuz it has been used in many places: 
https://github.com/koajs/koa/blob/77a4cfb8297bed0bc577425a6f42a0397547ca32/lib/context.js#L109
https://github.com/koajs/koa/blob/77a4cfb8297bed0bc577425a6f42a0397547ca32/lib/application.js#L188

jedahan: Cool, I can make another pull request to change that in a different PR some day in the future.

dead-horse:
close this for now, it not a big deal to change the error message and it may break somebody's code (Although it may be small)

PR #1205, opened 2018-06-14
fix: parse port in x-forwarded-for (#827)

### LINKED ISSUE #827: request.ip via x-forwarded-for should strip out ports
on azure (yes, i know), ports are also being sent as well as the ip in `x-forwarded-for`:

```
'x-forwarded-for': '23.243.1.1:38242'
```

`request.ip` should not return the port section. aye or nay?

jasonmacgowan:
Parses out IPs even if a port is sent in x-forwarded-for for IPv4.  Probably fails if port is sent with an IPv6 address, but works otherwise.

fixes #827

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -433,21 +433,28 @@ module.exports = {
   get ips() {
     const proxy = this.app.proxy;
     const val = this.get(this.app.proxyIpHeader);
-    let ips = proxy && val
-      ? val.split(/\s*,\s*/)
-          .map(host => {
-            let normalizedHost = host;
-            if (net.isIPv6(host)) {
-              normalizedHost = `[${host}]`;
-            }
+    let ips = [];
 
-            return parse(`http://${normalizedHost}`).hostname;
-          })
-          .filter(ip => !!ip)
-      : [];
+    if (proxy && val) {
+      ips = val.split(/\s*,\s*/)
+        .map(host => {
+          let normalizedHost = host;
+
+          if (net.isIPv6(host)) {
+            normalizedHost = `[${host}]`;
+          }
+
+          const hostname = new URL(`http://${normalizedHost}`).hostname;
+
+          return hostname.replace(/(^\[|\]$)/g, '');
+        })
+        .filter(ip => !!ip);
+    }
+
     if (this.app.maxIpsCount > 0) {
       ips = ips.slice(-this.app.maxIpsCount);
     }
+
     return ips;
   },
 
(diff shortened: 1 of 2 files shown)


3imed-jaberi (commented):
@jasonmacgowan, thank you for your PR 🙏🏻.
Plz, can you resolve the conflicts then use `parseurl` module over the `url` module.

lib/request.js
```diff
@@ -433,12 +433,28 @@ module.exports = {
...
+        .map(host => {
+          let normalizedHost = host;
+
+          if (net.isIPv6(host)) {
+            normalizedHost = `[${host}]`;
+          }
+
+          const hostname = new URL(`http://${normalizedHost}`).hostname;
+
+          return hostname.replace(/(^\[|\]$)/g, '');
```
jasonmacgowan, line 449: @3imed-jaberi I don't think I can use `parseurl` module here as it expects a `http.IncomingMessage` object and here we're working on a header.  I see `url.parse` is now deprecated so I swapped to using the `URL` constructor.  

Problem is this doesn't strip the `[]` chars from IPv6 hosts so now we gotta do it manually.  Regex to the rescue

jasonmacgowan:
Looks like @3imed-jaberi is going to be unavailable for a while based on his profile.

Anyone else around to take a look at this?

/cc @jonathanong @dougwilson

test/request/ips.js
```diff
@@ -23,6 +23,24 @@ describe('req.ips', () => {
...
+    describe('and contains IPv4', () => {
+      it('should not return port', () => {
+        const req = request();
+        req.app.proxy = true;
+        req.header['x-forwarded-for'] = '127.0.0.1:80,127.0.0.2';
+        assert.deepEqual(req.ips, ['127.0.0.1', '127.0.0.2']);
+      });
+    });
+
+    describe('and contains IPv6', () => {
```
dougwilson, line 36: I would recommend adding a test for a IPv6 address with a port (i.e `[::1]:80`) as well as a dns name with a port (i.e. `localhost:80`).

lib/request.js
```diff
@@ -433,12 +433,28 @@ module.exports = {
...
-      : [];
+    let ips = [];
+
+    if (proxy && val) {
+      ips = val.split(/\s*,\s*/)
+        .map(host => {
+          let normalizedHost = host;
+
+          if (net.isIPv6(host)) {
+            normalizedHost = `[${host}]`;
```
dougwilson, line 444: Because since this already validates it is the IP only, I would suggest simply shortcutting the rest of the processing for this case:
```suggestion
            return host;
```

PR #1209, opened 2018-06-25
Lazily initialize `request.accept` 

initial-wu:
To save memory consumption.


lib/request.js
```diff
@@ -462,6 +463,17 @@ module.exports = {
...
   },
 
+  /**
+   * Get accept object.
+   * Lazily memoized.
+   *
+   * @return {Object}
+   * @api public
+   */
+  get accept() {
```
dead-horse: maybe you should add `set accept()` to avoid breaking change.

initial-wu: All right.

initial-wu: I've pushed another commit to this PR.

lib/request.js
```diff
@@ -462,6 +463,27 @@ module.exports = {
...
+  },
+
+  /**
+   * Set accept object.
+   *
+   * @param {Object}
+   * @api public
+   */
+  set accept(obj) {
+    return this._accept = obj;
```
dead-horse: need a test case

lib/context.js
```diff
@@ -194,6 +194,7 @@ delegate(proto, 'request')
   .access('query')
   .access('path')
   .access('url')
+  .access('accpet')
```
fl0w: Misspelled

initial-wu: thx for reminding, i should have run a test

initial-wu:
Added some test cases and reorganized the commits. @fl0w @dead-horse

PR #1214, opened 2018-07-02
comment & cov: Fix comment & approve cov

xtx1130:
- fix comment for api
- ignore useless else cov branch

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -642,7 +642,7 @@ module.exports = {
    *     // => "text/plain"
    *
    *     this.get('Something');
-   *     // => undefined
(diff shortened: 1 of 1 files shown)


lib/request.js
```diff
@@ -280,6 +280,7 @@ module.exports = {
    */
 
   get URL() {
+    /* istanbul ignore else */
```
dead-horse: i think it is useless here?

xtx1130: Istanbul makes `else` as a default branch, even though you don't write `else`. So coverage test will cover not-exist `else` as coverage

PR #1220, opened 2018-07-11
perf: avoid stringify when set header

dead-horse:


lib/response.js
```diff
@@ -439,8 +439,8 @@ module.exports = {
     if (this.headerSent) return;
 
     if (2 == arguments.length) {
-      if (Array.isArray(val)) val = val.map(String);
-      else val = String(val);
+      // maybe we can drop support of undefined in the next major version
+      if (val === undefined) val = 'undefined';
```
dead-horse: other type of values will be stringify in `res.setHeader`

dead-horse:
about ~3% performance improve.

-before 
image

- after

image

lib/response.js
fengmk2: res.getHeader will broken change after this change.

fengmk2: > Sets a single header value for headers object. If this header already exists in the to-be-sent headers, its value will be replaced. Use an array of strings here to send multiple headers with the same name. Non-string values will be stored without modification. Therefore, request.getHeader() may return non-string values. However, the non-string values will be converted to strings for network transmission.

dead-horse: fixed

PR #1221, opened 2018-07-13
pref: use typeof instead of instanceof in stream send mode

xtx1130:
- use `typeof` instead of `instanceof`, `typeof` has higher performance than `instanceof`
- benchmark: [typof-vs-instanceof](https://jsperf.com/typof-vs-instanceof/1) 
image


dead-horse:
It can't improve too much here and `body instanceof Stream` is more semantic

PR #1225, opened 2018-07-25
Always use strict equality

medanat:


medanat:
I think the reason for all these `===` PRs is 2 fold:

1. Consistency: Koa already uses `===` in several places and `==` in others, no distinction as to why one or the other is used
2. Clear intentions: Given the type coercion behavior of `==`, it's best to use `===` to signify that no type coercion was necessary in the comparison

There was a time when `===` was more performant, that is no longer the case, not to measurable significance at least. However it's one less thing to consider when maintaining the codebase.

medanat:
@3imed-jaberi done

lib/response.js
```diff
@@ -166,12 +166,12 @@ module.exports = {
...
 
     // stream
-    if ('function' == typeof val.pipe) {
+    if ('function' === typeof val.pipe) {
       onFinish(this.res, destroy.bind(null, val));
       ensureErrorHandler(val, err => this.ctx.onerror(err));
 
       // overwriting
-      if (null != original && original != val) this.remove('Content-Length');
+      if (null != original && original !== val) this.remove('Content-Length');
```
3imed-jaberi, line 174:
```suggestion
      if (null !== original && original !== val) this.remove('Content-Length');
```

medanat: Given the nature of type coercion with nulls, I'm not sure this will be a safe move.

`null == undefined
true
null === undefined
false`

PR #1232, opened 2018-08-10
fix: redirect failed when not match rfc7230

XadillaX:
When I write:

```js
ctx.redirect('你好世界');
```

Node.js will throw an error due to https://github.com/nodejs/node/blob/862389b0aacfd41770c3fa6f692aed11c2a4b3ed/lib/_http_common.js#L245.

This PR is to encode the uri to match [rfc7230](https://tools.ietf.org/html/rfc7230#section-3.2.6).

> Needs backward.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -260,7 +260,17 @@ module.exports = {
   redirect(url, alt) {
     // location
     if ('back' == url) url = this.ctx.get('Referrer') || alt || '/';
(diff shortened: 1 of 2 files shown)


dead-horse:
you should encode uri before call `ctx.redirect` or `ctx.set` methods. we shouldn't change people's input here because it may break other people's code.

PR #1237, opened 2018-08-31
Adding a headers setter to match the request api in the response api

MYDIH:
Hi,

In a matter of consistency, I wish to be able to use a headers setter on the response, like I would on a request.

Thanks

PS: Koa is amazing, thank you guys

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -63,6 +63,21 @@ module.exports = {
   },
 
   /**
+   * Set an object of header fields.
+   *
+   * Examples:
+   *
+   *    this.headers({ Accept: 'text/plain', 'X-API-Key': 'tobi' });
+   *
+   * @param {Object|Array} headers
+   * @api public
+   */
+
+  set headers(headers) {
+    return this.set(headers);
+  },
+
+  /**
    * Get response status code.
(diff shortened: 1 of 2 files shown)


dead-horse:
-0 for me, I'd not like to add a duplicate API, and most of users are already used to set header with `response.set(headers)`

fl0w:
I'm +/-0 also.

slaskis:
I can see how `res.headers =` is expected usage. 

If anything maybe we can freeze the `res` object from properties that are not in use by koa to avoid confusion? Probably a very breaking change found at runtime sadly.

MYDIH:
I know that adding a duplicate API seems useless, but there is already one in the Request (headers is just calling header) ... I personnaly think that consistency should be preserved between the two APIs (adding set/get methods to Request is fine too). Having this would make things more obvious ...

3imed-jaberi (commented):
I'm +/-0 also. But I can say this help the newcomers.

jonathanong:
the API being a `=` is definitely confusing as you would expect the entire header object to be overwritten - you would not expect a merge / overwrite.

PR #1240, opened 2018-09-09
response.attachment append a parameter: options from contentDisposition

zzzs:
response.attachment require content-disposition, and content-disposition supports option parameter to make some settings, so i hope attachment can also support it.

thanks !


dead-horse:
can you add a test case for this to avoid later changes modified this feature.

zzzs:
@dead-horse thanks, already done!

zzzs:
Ok, I will do it two days later.

zzzs:
@fl0w already done!

PR #1242, opened 2018-09-11
test: make ci run faster

fengmk2:


test/response/status.js
```diff
@@ -28,7 +28,7 @@ describe('res.status=', function(){
       it('should throw', function(){
         assert.throws(function() {
           response().status = 999;
-        }, 'invalid status code: 999');
+        }, /invalid status code: 999/);
```
fengmk2: node 10 only support regexp

Issue #1243, opened 2018-09-11
Update debug package to version 4.0.0


ryanelian:
I am getting warnings in my CI due to updated debug package:

- https://github.com/visionmedia/debug/issues/603#issuecomment-420237335

- https://github.com/visionmedia/debug/releases/tag/4.0.0

```
warning webpack-serve > koa > debug@3.2.4: `debug` versions 3.2.0-3.2.2 break Node 4. Users of node LTS and onward (modern Node versions): please upgrade to `debug@4` or above as soon as possible. Node 4 users: please pin to `debug@>=3.2.3 <4` as it will be the last version range that supports Node@4.
```

The `debug` project recommends a package update to version 4.0.0 to mitigate vulnerabilities and performance improvement.

fengmk2:
Already fixed on https://github.com/koajs/koa/pull/1241, please reinstall and it will be fixed now.

fengmk2:
https://github.com/koajs/koa/blob/master/package.json#L30 koa master need to update

ryanelian:
> v1.x is already fixed on #1241, please reinstall and it will be fixed now.

https://github.com/koajs/koa/pull/1241/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R28

3.2.4 is what causing the deprecation warning. Perhaps pin the package to exact 3.1.0?

fengmk2:
@ryanelian I'm going to pin debug@2 on koa@1 and debug@~3.1.0 on koa@2.

PR #1245, opened 2018-09-11
fix: pin debug@~3.1.0 avoid deprecated warnning

fengmk2:
closes https://github.com/koajs/koa/issues/1243


fengmk2:
@dead-horse https://github.com/koajs/koa/blob/master/package.json#L60 need to support node@4 unless release a major version.

PR #1247, opened 2018-09-12
fix: Stream body & ENOENT support & a test case

initial-wu:
- The oringinal test `end(done)` can not expose the error as expected, so you can not find the problem before. i replaced it with a returning promise.

- call `end()` without a callback will cause a error in about 2 minutes after the test end only if there are some failing tests.

image

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -109,6 +109,9 @@ const proto = module.exports = {
     // to node-style callbacks.
     if (null == err) return;
 
+    // ENOENT support
+    if ('ENOENT' == err.code) err = createError(404);
+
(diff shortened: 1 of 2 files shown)


lib/context.js
```diff
@@ -109,6 +109,9 @@ const proto = module.exports = {
     // to node-style callbacks.
     if (null == err) return;
 
+    // ENOENT support
+    if ('ENOENT' == err.code) err = createError(404);
```
dead-horse: why move this to front? Also we should not change the original `err` here

initial-wu: An ENOENT error is NOT a instance of Error, so it will be considered a non-error.

image

And server will reply a 500 instead of 404.

image

PR #1249, opened 2018-09-14
fix Readme.md shortcuts error

iamswf:
Readme.md has a shortcuts error...
 `ctx.type`  is a shortcut for `ctx.response.type`.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -135,7 +135,7 @@ app.use(async (ctx, next) => {
 ```
 
 The `Context` object also provides shortcuts for methods on its `request` and `response`.  In the prior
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -135,7 +135,7 @@ app.use(async (ctx, next) => {
 ```
 
 The `Context` object also provides shortcuts for methods on its `request` and `response`.  In the prior
-examples,  `ctx.type` can be used instead of `ctx.request.type` and `ctx.accepts` can be used
+examples,  `ctx.type` can be used instead of `ctx.response.type` and `ctx.accepts` can be used
```
dead-horse: `ctx.type` is delegate to `ctx.request.type` and `ctx.type=` is delegate to `ctx.response.type=`

PR #1251, opened 2018-09-20
refactor: use absolute equal for string comparison

ulivz:
## Summary

It's recommended to use `===` instead of `==` for value comparison. in koa, now `null` and `undefined` are considered the same, that makes sense, but not for strings.

### DIFF
--- a/docs/api/request.md
+++ b/docs/api/request.md
@@ -181,7 +181,7 @@ ctx.body = await db.find('something');
 
 ### request.secure
 
-  Shorthand for `ctx.protocol == "https"` to check if a request was
+  Shorthand for `ctx.protocol === "https"` to check if a request was
   issued via TLS.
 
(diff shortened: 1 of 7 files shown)


medanat:
> I don't understand why there's endless need to change functional code that makes use of defined, well documented, coercion 😆.
> 
> I'm -1 on this as I see this equivalent to stylistic PRs which are denied by convention previously.

https://github.com/koajs/koa/pull/1225#issuecomment-424697950

anlexN (approved):

3imed-jaberi (commented):
Thank you @ulivz..
I don't like to be this person but I think should close this PR because it's outdated.
Without talking about documentation, the rest of the code was updated as this PR do in the current Koa version.

@dead-horse @niftylettuce @miwnwski  @jonathanong  ...

PR #1256, opened 2018-10-04
fix: switched JSON.stringify to fast-safe-stringify

niftylettuce:
ref: https://github.com/davidmarkclements/fast-safe-stringify

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -21,6 +21,7 @@ const http = require('http');
 const only = require('only');
 const convert = require('koa-convert');
 const deprecate = require('depd')('koa');
+const safeStringify = require('fast-safe-stringify');
 
 /**
  * Expose `Application` class.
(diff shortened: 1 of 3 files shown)


crobinson42:
My initial thoughts are hesitant to add this as a dependency. 

- what problem does it solve
- are there user issues raised about this as a concern or _issue_
- is this the best library to use to replace JSON.stringify

dead-horse:
actually it is easy to define your own response method using `fast-safe-stringify` to serialize JSON body without change the default JSON.stringify

niftylettuce:
@dead-horse is it possible for you to give a clean and easy example?

dead-horse:
```js
ctx.body = safeJSON.stringify(obj);
ctx.type = 'json';
```

or you can make it as a function in `context` with a middleware?

niftylettuce:
Yeah I meant without having to do `ctx.body = foobar` everytime.  Is there a simple way to do this?  We should possibly allow a custom stringify function to get passed?  It would also open up the door for people to "blacklist" certain properties they don't want to get sent along.

PR #1262, opened 2018-10-19
fix: use :authority header of http2 requests as host

### LINKED ISSUE #1174: ctx.request.origin is wrong
Ciao!

How are you?

Weird output:
```js
    app.use(ctx => {
        ctx.body = {
            hello: 'world',
            origin: ctx.request.origin,
        }
    })
```

image

image

mgjm:
The now stable [http2 api](https://nodejs.org/api/http2.html#http2_headers_object) does not use the host header. The host can be found in the http2 [`:authority` header](https://http2.github.io/http2-spec/#rfc.section.8.1.2.3.p.1). Therefore koa should use this header instead.

This fixes the `host` property and all properties that depend on it for http2 requests. These dependent properties include the `hostname`, the `origin` and the `URL`. This will also fix #1174.

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -252,6 +252,7 @@ module.exports = {
   get host() {
     const proxy = this.app.proxy;
     let host = proxy && this.get('X-Forwarded-Host');
+    if (this.req.httpVersionMajor >= 2) host = this.get(':authority');
     host = host || this.get('Host');
     if (!host) return '';
     return host.split(/\s*,\s*/)[0];
--- a/test/request/host.js
+++ b/test/request/host.js
@@ -18,6 +18,39 @@ describe('req.host', () => {
     });
   });
 
+  describe('when less then HTTP/2', () => {
+    it('should not use :authority header', () => {
+      const req = request({
+        'httpVersionMajor': 1,
+        'httpVersion': '1.1'
+      });
+      req.header[':authority'] = 'foo.com:3000';
+      req.header.host = 'bar.com:8000';
+      assert.equal(req.host, 'bar.com:8000');
+    });
+  });
+
+  describe('when HTTP/2', () => {
+    it('should use :authority header', () => {
+      const req = request({
+        'httpVersionMajor': 2,
+        'httpVersion': '2.0'
+      });
+      req.header[':authority'] = 'foo.com:3000';
+      req.header.host = 'bar.com:8000';
+      assert.equal(req.host, 'foo.com:3000');
+    });
+
+    it('should use host header as fallback', () => {
+      const req = request({
+        'httpVersionMajor': 2,
+        'httpVersion': '2.0'
+      });
+      req.header.host = 'bar.com:8000';
+      assert.equal(req.host, 'bar.com:8000');
+    });
+  });
+
   describe('when X-Forwarded-Host is present', () => {
     describe('and proxy is not trusted', () => {
       it('should be ignored', () => {


fengmk2:
cc @dead-horse

lib/request.js
```diff
@@ -252,6 +252,7 @@ module.exports = {
   get host() {
     const proxy = this.app.proxy;
     let host = proxy && this.get('X-Forwarded-Host');
+    if (this.req.httpVersionMajor >= 2) host = this.get(':authority');
```
atian25: `X-Forwarded-Host` will force overwrite at http2?

fengmk2: https://github.com/nodejs/node/blob/e7f710c1d4729890806610e5450d7a0edfb4ac11/lib/internal/http2/compat.js#L358

use `this.req. authority` will be better.

mgjm: This seems to be an undocumented property. Is it save to use it and should koa use undocumented props?

mgjm: The way I understand the definition of `:authority` is that it can be viewed as a replacement of the `Host` header.
Therefore a Proxy would set the `:authority` header to the backend hostname und set the `X-Forwarded-Host` to the original host.
So the current order seems to be correct.

panva: I think we wanna keep the forwarded header, don’t we?

dead-horse: If `app.proxy` was set, I'd like to keep the `X-Forwarded-Host` set by reverse proxy(nginx).

 And in my practice, when client request nginx with http 1/2 and nginx proxy to node server, it always use `X-Forwarded-Host`, it is a de-facto standard for reverse proxy to set host when visit upstream.

dead-horse: If `app.proxy` was set, I'd like to keep the `X-Forwarded-Host` set by reverse proxy(nginx).

 And in my practice, when client request nginx with http 1/2 and nginx proxy to node server, it always use `X-Forwarded-Host`, it is a de-facto standard for reverse proxy to set host when visit upstream.

dead-horse: If `app.proxy` was set, I'd like to keep the `X-Forwarded-Host` set by reverse proxy(nginx).

 And in my practice, when client request nginx with http 1/2 and nginx proxy to node server, it always use `X-Forwarded-Host`, it is a de-facto standard for reverse proxy to set host when visit upstream.

dead-horse: If `app.proxy` was set, I'd like to keep the `X-Forwarded-Host` set by reverse proxy(nginx).

 And in my practice, when client request nginx with http 1/2 and nginx proxy to node server, it always use `X-Forwarded-Host`, it is a de-facto standard for reverse proxy to set host when visit upstream.

dead-horse: If `app.proxy` was set, I'd like to keep the `X-Forwarded-Host` set by reverse proxy(nginx).

 And in my practice, when client request nginx with http 1/2 and nginx proxy to node server, it always use `X-Forwarded-Host`, it is a de-facto standard for reverse proxy to set host when visit upstream.

mgjm: Ok, so no changes are required.

fl0w: It's unsafe either way, at least `this.req.authority` is a specific getter with a sane naming that conforms with the RFC headerfield.

mgjm: `this.req.headers` is the documented way to access http headers and `this.get` is using this way. And `:authority` is the official header name from the [RFC 7540](https://tools.ietf.org/html/rfc7540#section-8.1.2.3). So this way should not be unsafe.

Whereas `this.req.authority` is not documented and therefore could change in a minor update and is considered an internal API (as described in the [node COLLABORATOR_GUIDE](https://github.com/nodejs/node/blob/master/COLLABORATOR_GUIDE.md#internal-vs-public-api)).

But I agree that `this.req.authority` should be part of the public API and I just created an Issue in the node repo (nodejs/node#23825).

panva:
The requested change to keep x-forwarded-for values as host is not in?

mgjm:
@panva The current order is...
1. If `app.proxy` the `X-Forwarded-Host` header
2. If HTTP/2 the `:authority` header
3. The `Host` header

panva:
This is not what the review by @dead-horse [suggested](https://github.com/koajs/koa/pull/1262#discussion_r226896280)

> If `app.proxy` was set, I'd like to keep the `X-Forwarded-Host` set by reverse proxy(nginx).
> 
> And in my practice, when client request nginx with http 1/2 and nginx proxy to node server, it always use `X-Forwarded-Host`, it is a de-facto standard for reverse proxy to set host when visit upstream.

dead-horse:
> 1. If app.proxy the X-Forwarded-Host header
> 2. If HTTP/2 the :authority header
> 3. The Host header

I think this works for me, `X-Forwarded-Host` is set by reverse proxy, we just trust the proxy when `app.proxy` is true. Otherwise we detect HTTP 1 / HTTP 2 to choose which header we should use to get `host`.

panva:
I agree, but the implementation takes :authority whenever it's set over x-forwarded-host at the moment.

fengmk2:
@panva I will release a bugfix version soon.

panva:
@fengmk2 perfect!

fengmk2:
https://github.com/koajs/koa/pull/1263

mgjm:
@panva Good catch. @fengmk2 Thanks for the quick fix 👍

Issue #1253, opened 2018-09-21
UnsupportedWarning: Status message is not supported by HTTP/2 (RFC7540 8.1.2.4)


khflx:
Hi
I tried to use the koa 2.5.3 framework on http2, but when the client accesses a url that does not exist, it will trigger the "UnsupportedWarning: Status message is not supported by HTTP/2 (RFC7540 8.1.2.4)" warning.
This seems to be because koa is set or gets [response.statusMessage](https://nodejs.org/dist/latest-v10.x/docs/api/http2.html#http2_response_statusmessage)[Related source code](https://github.com/nodejs/node/blob/master/lib/internal/http2/compat.js#L550)

Is there any way to solve it?

fl0w:
Yes and no. Setting the message is done currently so removing that would arguably be breaking.

One could remove [setting the message]() for HTTP/2 but that would require Koa to know what server it's running on - uncool.

We could make it flag-able, e.g. `app.statusMessageOn404 = false` *(working title)* with default being set to true. But that's something @dead-horse @jonathanong could pitch in on before I do it.

edevil:
What if our app is running using both http and http2? Would we need two separate apps with different flag values?

fl0w:
With that solution, yes. But I agree.

Sort of off topic but it feels like Koa needs a grand vision on how to support http/2 before I start putting some work into it (and I'm probably not the right person to envision that). Specifically how the flow of pushing should be implemented. I almost feel like there needs to be another level of abstraction between Koa and server to keep Koa agnostic (without going `if http1 {} else if http2 {}` all over the place).

ddtramp:
use `spdy` package should be ok
npm install spdy -SE

 const server = spdy.createServer(options, app.callback())
  server.listen(443)

ddtramp:
do not use http2 module in the nodejs core, use spdy

fl0w:
Solved by https://github.com/koajs/koa/pull/1264

PR #1264, opened 2018-10-23
Status message is not supported on HTTP/2

edevil:
Addresses #1253 

Logic example for different behaviour depending on http version taken from #1262


edevil:
@dead-horse @fl0w any feedback on this?

dead-horse:
can you add a test case for this change?

edevil:
@dead-horse done

Issue #1261, opened 2018-10-17
Deprecation Warning http-errors - createError


matttowerssonos:
`http-errors` has a deprecation warning for calls to `createError()` where the first argument is non-numeric.  This happens when `throw()` is called in [`context.js [96]`](https://github.com/koajs/koa/blob/master/lib/context.js#L75-L97), which allows for non-numeric first parameters.  The documentation for `throw()` should be updated to indicate that the first parameter should be the HTTP response code.

The deprecation warning can be found in `http-errors`, [`index.js [68]`](https://github.com/jshttp/http-errors/blob/master/index.js#L68)

PR #1268, opened 2018-10-29
doc: updated docs for throw() to pass status as first param.

WaleedAshraf:
Closes: #1261 

I have updated comments/docs to use `status` only as the first param.
Let me know if we should also remove `number` type from the second param of `throw()` function.

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -73,18 +73,19 @@ const proto = module.exports = {
   assert: httpAssert,
 
   /**
-   * Throw an error with `msg` and optional `status`
-   * defaulting to 500. Note that these are user-level
+   * Throw an error with `status` (default 500) and
+   * `msg`. Note that these are user-level
    * errors, and the message may be exposed to the client.
    *
    *    this.throw(403)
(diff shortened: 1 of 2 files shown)


WaleedAshraf:
@fl0w Ping!

dead-horse:
thanks!

lib/context.js
```diff
@@ -73,19 +73,20 @@ const proto = module.exports = {
...
    *    this.throw(400, 'name required')
    *    this.throw('something exploded')
-   *    this.throw(new Error('invalid'), 400);
-   *    this.throw(400, new Error('invalid'));
+   *    this.throw(new Error('invalid'))
+   *    this.throw(400, new Error('invalid'))
    *
    * See: https://github.com/jshttp/http-errors
    *
+   * Note: `stats` should only be passed as the first parameter.
```
fl0w: This was misspelled, fixed in https://github.com/koajs/koa/pull/1269

WaleedAshraf: Oh, thanks 🎉

PR #1274, opened 2018-11-24
Update readme to add babel 7 instructions

vikr01:
- Update mentioned `babel` dependencies to their corresponding new names.
- Update links to babel dependencies mentioned.
- `@babel/register` should not be intended for production (since it's building and then caching files during runtime). `@babel/cli` should be used to build the files.
- Since the files will be built after `babel` is run, it isn't needed at runtime and is therefore only needed as a `devDependency` (`npm install --save` => `npm install --save-dev`).

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -161,16 +161,22 @@ Learn more about the application object in the [Application API Reference](docs/
 
 ## Babel setup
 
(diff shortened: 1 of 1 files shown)


fl0w:
LGTM

PR #1282, opened 2018-12-05
[#1277]: Documentation: response.set(fields) overwrites previous header fields?

### LINKED ISSUE #1277: Documentation:
Under Response (koa response object), there's `response.set(fields)`. The text reads "Set several response header fields with an object:," with an example. However, it's not clear if setting these fields via object means that the header becomes ONLY those fields, or whether passed-in-object extends the previous header. It's also not clear what the behavior is if the fields are identical to current header values. 

It would be helpful to clarify these questions in the documentation. I still haven't figured the answer out. 

https://github.com/koajs/koa/blob/master/docs/api/response.md#responsesetfields

doug-wade:
Fixes #1277

### DIFF
--- a/docs/api/response.md
+++ b/docs/api/response.md
@@ -193,7 +193,7 @@ ctx.set({
 });
 ```
 
-This sets or updates header field by specified keys and doesn't reset the entire header.
+This delegates to [setHeader](https://nodejs.org/dist/latest/docs/api/http.html#http_request_setheader_name_value) which sets or updates headers by specified keys and doesn't reset the entire header.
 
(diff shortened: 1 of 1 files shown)


fl0w:
Personally I would somehow add that `set` "delegates to [setHeader](https://nodejs.org/dist/latest/docs/api/http.html#http_request_setheader_name_value) which only sets or updates [...]".

Otherwise LGTM

doug-wade:
Thanks for the feedback @fl0w !  I've made the requested changes.

Issue #1162, opened 2018-03-21
What does ctx.app.emit does?


bertolo1988:
I saw it here: https://github.com/koajs/koa/wiki/Error-Handling

fl0w:
A Koa application extends the internal EventEmitter.
`ctx.app.emit` simply emits an event with a type, defined by the first argument. For each event you can hook on "listeners", which is a function that is called when the event is emitted.

Take a quick look at http://koajs.com/#error-handling and the official node.js documentation on [EventEmitters](https://nodejs.org/docs/latest-v8.x/api/events.html)

bertolo1988:
Thanks for the answer @fl0w .

PR #1284, opened 2018-12-07
[#1123]: Document ctx.app.emit

### LINKED ISSUE #1123: Document ctx.app.emit
The only place I've seen any information about `ctx.app.emit` is one line in https://github.com/koajs/koa/wiki/Error-Handling, but it's not clear how or why to use it. What is the point of doing this when in a middleware you can just try/catch and if there were any errors you just set the appropriate HTTP code and response body as in that example. You can simply do any logging or other work you need to there right?

doug-wade:
Fixes #1123

### DIFF
--- a/docs/api/context.md
+++ b/docs/api/context.md
@@ -62,6 +62,10 @@ ctx.state.user = await User.find(id);
 
   Application instance reference.
 
+### ctx.app.emit
+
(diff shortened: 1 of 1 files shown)


doug-wade:
Credit to @fl0w for writing the copy in #1162 -- thanks!

PR #1285, opened 2018-12-07
[#965]: Document other body types

### LINKED ISSUE #965: Other body types
While in the documentation it states that the body will return a JSON object if the body is set to an object, I found out that arrays also work the same way:

```js
ctx.body = ['foo', 'bar'];
```

This is awesome, but it might be a bug since [in the code for that part](https://github.com/koajs/koa/blob/master/lib/response.js#L180) the *leftovers* are treated as JSON:

```js
set body(val) {
    // ...
    if ('string' == typeof val) // ...
    if (Buffer.isBuffer(val)) // ...
    if ('function' == typeof val.pipe) // ...
    
    // Everything else
    this.remove('Content-Length');
    this.type = 'json';
}
```

Notably this makes other things work:

```js
app.use(ctx => {
  ctx.body = false;
});
```

But some other things won't:

```js
app.use(ctx => {
  ctx.body = () => {
    console.log('Hello there!');
  };
});
```

**Is this an intended feature?** Everything that is not one of the above is just treated as JSON?

Notably in counter-intuitive examples, this will work and return a json `{}` with a status code `200` (I know about `throw()`, just pointing out this works not as expected and might be confusing):

```js
app.use(ctx => {
  ctx.body = new Error('I work!');
});
```

doug-wade:
Fixes #965

### DIFF
--- a/docs/api/response.md
+++ b/docs/api/response.md
@@ -124,6 +124,16 @@ so you can make a correction.
 
 If `response.status` has not been set, Koa will automatically set the status to `200` or `204`.
 
+Koa doesn't guard against everything that could put as a response body -- a function doesn't serialise meaningfully, returning a boolean may make sense based on your application, and while an error works, it may not work as intended as some properties of an error are not enumerable.  We recommend adding middleware in your app that asserts body types per app.  A sample middleware might be:
+
+```
+app.use(async (ctx, next) => {
+  await next()
+
+  ctx.assert.equal('object', typeof ctx, 500, 'some dev did something wrong')
+})
+```
+
 #### String
 
   The Content-Type is defaulted to text/html or text/plain, both with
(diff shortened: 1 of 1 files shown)


docs/api/response.md
```diff
@@ -124,6 +124,16 @@ so you can make a correction.
 
 If `response.status` has not been set, Koa will automatically set the status to `200` or `204`.
 
+Koa doesn't guard against everything that could put as a response body -- a function doesn't serialise meaningfully, returning a boolean may make sense based on your application, and while an error works, it may not work as intended as some properties of an error are not enumerable.  We recommend adding middleware in your app that asserts body types per app.  A sample middleware might be:
```
fl0w:
```suggestion
Koa doesn't guard against everything that could be put as a response body -- a function doesn't serialise meaningfully, returning a boolean may make sense based on your application, and while an error works, it may not work as intended as some properties of an error are not enumerable.  We recommend adding middleware in your app that asserts body types per app.  A sample middleware might be:
```

PR #1286, opened 2018-12-09
HTTP requests processing prioritisation

jkyberneees:
Allows Koa to optionally process HTTP requests using the "check" tasks queue, resulting in a bigger performance gain if running in multicore CPU runtimes.

- With `app.prioRequestsProcessing = true;` = **37440.67 reqs/s**  
- Without = 25711.95

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -193,6 +193,44 @@ And have your `.babelrc` setup:
 }
 ```
 
+## Performance
+
+By default, Koa is already fast enought for most common use cases.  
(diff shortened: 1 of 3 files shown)


dead-horse:
IMO, Maybe we can override `app.handleRequest` to implement this feature when really need it. It is not necessary to provide `app.prioRequestsProcessing` by koa.

fl0w:
I agree, this feels like a separate package, not a core lib feature.

Issue #1275, opened 2018-11-30
Content length response header incorrectly parsed


kerim1:
When requesting the length using `ctx.length`, Koa parses the content length response header using the double tilde operator: https://github.com/koajs/koa/blob/master/lib/response.js#L214 However, this operation also performs a 32-bit conversion, resulting in incorrect and weird results when serving large files. Isn't it therefore better to use the `Math.trunc()` function instead to parse the content length response header?

JeremiGendron:
using Math.trunc is 5% slower on 100 000 000 iterations of res.length when it's set to 164 chars.

fl0w:
Fixed in 9be858312553002841725b617050aaff3c48951d

PR #1288, opened 2018-12-20
ibu: replace ~~ with Math.trunc in res.length (option)

JeremiGendron:
#1275 is fixed by this, though this implementation probably shouldn't be merged into master it can benefit whoever was suffering from the content length being too large to be notnoted.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -211,7 +211,7 @@ module.exports = {
       return;
     }
 
-    return ~~len;
+    return Math.trunc(len);
   },
 
   /**


JeremiGendron:
test.js:
```js
const { response } = require('./context.js')

res = response()
res.body = "Testing purposes, the average payload is most likely much larger than this; therefore some extra padding was added during the creation of this magnificent sentence."

const start = Date.now()

for(let i = 0; i < 100000000; i ++) {
  let temp = res.length
}

const diff = Date.now() - start
console.log(res.length)
console.log(diff / 1000)
```

lib/response.js
```diff
@@ -211,7 +211,7 @@ module.exports = {
       return;
     }
 
-    return ~~len;
```
dead-horse: `Math.trunc(len) || 0` to avoid `NaN`

PR #1300, opened 2019-01-16
fix: app.middleware -> app.middlewares

nswbmw:
I think `middlewares` is more reasonable for an array property, but this change may bring incompatible changes.


dead-horse:
as you said, it may bring some incompatible change and it is not a big deal IMO, I'd like to keep it as-is.

PR #1302, opened 2019-01-16
Initial create .npmignore

mingchen:
Do not pack un-necessary files to npm package to reduce package size.

see https://docs.npmjs.com/misc/developers


dead-horse:
https://github.com/koajs/koa/blob/master/package.json#L62  we already specified `files` here, thanks for your contribution.

Issue #1019, opened 2017-07-06
consider providing a way to override `respond` function


bttmly:
I'd be interested in a way to override the [`respond` function](https://github.com/koajs/koa/blob/aaac09af1a6aa02161fead1422fac460fbdcce3e/lib/application.js#L193-L238) – perhaps exposing it as a method?

My particular use case at the moment has to do with how the current `respond` function interacts with `ctx.respond = false`, in the context of embedding Koa within a larger application. Generally this works fine -- you can set `ctx.respond` to false, and after Koa's middleware stack runs to completion you can check `res.headersSent` to see if Koa handled the request; if not, continue handling it elsewhere. However this fails when `ctx.body` is a stream. Because of the semantics of `.pipe()`, the first write will happen asynchronously, and when `ctx.headersSent` is checked, it is `false`. This particular case could possibly be solved by doing a `res.write("")` before `body.pipe(res)` but that fails a couple test cases, for reasons that weren't immediately obvious.

While this is definitely a niche case, exposing a way to override or customize `respond` would certainly have other benefits. During a large project we are currently undertaking to migrate an application from Express to Koa, not being able to customize and instrument the response step has required several awkward workarounds.

jonathanong:
yeah, streams kind of suck. ideally, we could fix it here. perhaps you could give us a test case and we can fix it here.

bttmly:
Turns out that the behavior I want (flush the headers when you do `body.pipe(res)`) is incompatible with handling errors in streaming response bodies. For my specific case I would rather have the headers flushed immediately at the expense of error handling; it sort of depends on the use case. I'm also curious, since I don't see a test case for it, what is the intended behavior if a stream encounters an error after its first `data` event has arrived (and thus the status code has been set and the headers flushed)?

Anyway, providing a way to override `respond` would be a nice escape hatch for advanced or uncommon use cases, since there are a lot of assumptions baked into the `respond` function.

jonathanong:
> what is the intended behavior if a stream encounters an error after its first data event has arrived (and thus the status code has been set and the headers flushed)?

you either make your stream super smart and change its data to an error page or something or you do nothing and let it finish.

bttmly:
[...]

PR #1303, opened 2019-01-17
respond as instance method

bttmly:
Per https://github.com/koajs/koa/issues/1019, this PR moves `respond` into the Application class so it can be subclassed or overridden. A couple use cases off the top of my head:
- specialized handling of additional types that may be in ctx.body
- ability to instrument serialization time for exceptionally large responses
- escape hatch for some [gripes](https://github.com/koajs/koa/issues/998) about Koa's existing responder

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -146,7 +146,7 @@ module.exports = class Application extends Emitter {
     const res = ctx.res;
     res.statusCode = 404;
     const onerror = err => ctx.onerror(err);
-    const handleResponse = () => respond(ctx);
+    const handleResponse = () => this.respond(ctx);
     onFinished(res, onerror);
     return fnMiddleware(ctx).then(handleResponse).catch(onerror);
   }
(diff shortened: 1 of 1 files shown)


dead-horse:
I don't think rewrite respond method is a good idea, because everyone who write middlewares for koa will assume koa will treat `ctx.body` in a certain way.

also I think it is easy to implement your needs using middleware, which is more fit with koa's philosophy:

```js
  await next();
  if (ctx.type === 'additional-types') {
    ctx.body = handleAdditionalType(ctx.body);
  }

  if (ctx.bodySchema) {
    ctx.body = fastJsonStringifys.get(ctx.bodySchema)(ctx.body);
  }

  if (ctx.body === null && ctx.path.startsWith('/api/')) {
    ctx.body = { error: 'empty' };
  }
});
```

bttmly:
Fair enough 👍

PR #1308, opened 2019-01-27
Change set status assert, use statuses when possible

fl0w:
Reference, #1119 #1042

This is a different approach to #1119 for allowing custom statuses. Instead of adding an API in Koa on top of statuses, just don't disallow anything unless it's a 3 digit integer.

This is a result from the fact that #1119 [got stale or never got any traction](https://github.com/koajs/koa/pull/1119#issuecomment-457650532), that statuses package can be ~~repopulated~~ added to by its design, [dead-horses comment](https://github.com/koajs/koa/issues/1042#issuecomment-455100429) about RFC and no real purpose to disallow custom statuses, and the fact that I personally dislike previous approaches.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -83,8 +83,8 @@ module.exports = {
   set status(code) {
     if (this.headerSent) return;
 
-    assert('number' == typeof code, 'status code must be a number');
-    assert(statuses[code], `invalid status code: ${code}`);
+    assert(Number.isInteger(code), 'status code must be a number');
+    assert(code >= 100 && code <= 999, `invalid status code: ${code}`);
     this._explicitStatus = true;
     this.res.statusCode = code;
     if (this.req.httpVersionMajor < 2) this.res.statusMessage = statuses[code];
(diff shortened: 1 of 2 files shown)


lib/response.js
```diff
@@ -84,7 +84,7 @@ module.exports = {
     if (this.headerSent) return;
 
     assert('number' == typeof code, 'status code must be a number');
-    assert(statuses[code], `invalid status code: ${code}`);
+    assert(/^[0-9]{3}$/.test(code), `invalid status code: ${code}`);
```
dead-horse: how about 

```js
assert(Number.isIngeter(code) && code >= 100 && code <= 999, 'message')
```

fl0w: Yea, I'll update it. I ran a quick bench just to check how bad RegExp is in comparison (even though it's pointless in a real world scenario as I can't imagine `ctx.status` being a bottleneck:

```sh
Valid RegExp x 37,061,865 ops/sec ±0.51% (88 runs sampled)
Valid expression x 728,862,990 ops/sec ±1.60% (82 runs sampled)
```

The only thing I'd change is to not check both for number and Number.isInteger. I'll move isInteger to the above assert and keep that message (less "breaking") and replace RegExp.

fl0w:
Updated, squashed and rebased.

dead-horse:
koa@2.7.0 released

PR #1309, opened 2019-02-01
use 'fs-extra' instead of 'fs-promise'

rosald:
fs-promise is now deprecated.
Use fs-extra instead.

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -196,7 +196,7 @@ app.use(async function (ctx, next) {
 
 
 ```js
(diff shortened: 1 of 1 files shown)


fl0w:
-1

fs-promise is only a wrapper while fs-extra extends. If anything, use native fs promises (`require('fs').promises`) but as it's still experimental I'm on the fence about this as well.

dead-horse:
use `mz/fs` is better here

```js
const fs = require('mz/fs'); 
```

PR #1318, opened 2019-04-07
update devDependencies and fix discovered issues

jstewmon:
- context.onerror destorys the request socket if it is called after
  headers have been sent to complete the request
- context.onerror now throws TypeError if passed a non-Error to align
  with application.onerror
- fixed lint errors
  - remove usages of deprecated assertions
  - removed space between async and opening parens
- replace calls to server.listen by passing app.callback() to supertest
  so that tests do not leave open event handlers
- replaced callback completion test with async tests in cases where the
  tests failed due to callbacks being called in unexpected order

### DIFF
--- a/benchmarks/middleware.js
+++ b/benchmarks/middleware.js
@@ -13,7 +13,7 @@ console.log(`  ${n}${useAsync ? ' async' : ''} middleware`);
 
 while (n--) {
   if (useAsync) {
-    app.use(async (ctx, next) => await next());
+    app.use(async(ctx, next) => next());
   } else {
     app.use((ctx, next) => next());
   }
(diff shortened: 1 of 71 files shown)


jstewmon:
This one is a bit of a doozie - I started out wanting to address an issue when the body is a stream (which I'll follow up with in another issue), but I was distracted by all the deprecation warnings, and noticed the potential for some issues. Updating jest uncovered some issues with the test implementations (and one with the app), which are now fixed. :-)

jstewmon:
node 7 is failing the deepStrictEqual check for the context query... I struggled to get that check passing locally w node 10 - I guess the prototype comparison fails when expected is `{}`. I found that using jest's `expect(ctx.query).toEqual({})` passed too, but I tried to stick with the builtin assert methods, since there's no other use of `expect`...

dead-horse:
can you separate it into different PRs to help us review your code ?

jstewmon:
@dead-horse I think all these changes are strictly necessary to remove depreciation warnings fix the tests after updating the deps. I suppose I could separate the depreciation updates, which is the source of most what you’d consider noise if that’s what you’re looking for?

anlexN (approved):

3imed-jaberi (commented):
Thank you @jstewmon 
I don't like to be this person but I think should close this PR because it's outdated.

PR #1321, opened 2019-04-16
Buffer() is deprecated due to security and usability issues. so use t…

tkvern:
…he Buffer.alloc() instead


tkvern:
Buffer() is deprecated due to security and usability issues. so use the Buffer.from() instead

PR #1327, opened 2019-04-25
Allow to use custom http modules(e.g unit-http)

alg3rd:
This change will allow to use application server Nginx Unit(https://unit.nginx.org/installation/#installation-nodejs-package) in koa applications without any changes to existing codebases.

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -61,7 +61,29 @@ module.exports = class Application extends Emitter {
 
   listen(...args) {
     debug('listen');
-    const server = http.createServer(this.callback());
+    let httpModule = http;
+    if (args.length > 0 && typeof args[0] === 'object') {
+      httpModule = args[0];
+      args.shift();
+    }
+    const server = httpModule.createServer(this.callback());
+    return server.listen(...args);
+  }
+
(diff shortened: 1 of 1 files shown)


lib/application.js
```diff
@@ -61,7 +61,12 @@ module.exports = class Application extends Emitter {
 
   listen(...args) {
     debug('listen');
-    const server = http.createServer(this.callback());
+    let httpModule = http;
+    if (args.length > 0 && typeof args[0] === 'object') {
+      httpModule = args[0];
```
kroleg: you can write `httpModule = args.shift();`

kroleg:
i don't think that this should be included in koa because there is already an easy way to use custom http server:
```
const app = new Koa();
const unitHttp = require('unit-http');
const server = unitHttp.createServer(app.callback());
server.listen(port, () => {
    console.log('Listening on port ' + port);
});
```

fl0w:
Aye, this is a no go from me. Use `app.callback() // e.g. spdy.listen(app.callback())`

alg3rd:
Thank you for responses.

PR #1346, opened 2019-06-24
chore: Consider more concise syntax

junyiz:


lib/request.js
```diff
@@ -343,7 +343,7 @@ module.exports = {
 
   get idempotent() {
     const methods = ['GET', 'HEAD', 'PUT', 'DELETE', 'OPTIONS', 'TRACE'];
-    return !!~methods.indexOf(this.method);
```
dead-horse: koa need support node 4 and `Array.includes` didn't support in this version.

https://github.com/koajs/koa/blob/master/package.json#L59

junyiz: Okay, THX

PR #1349, opened 2019-06-27
chore(deps): fix audit error by upgrading jest to ^24

fl0w:
Upgrade Jest due to npm audit errors, which gets me to what @jstewmon stumbled upon in #1318.

This PR **only** upgrades Jest (due to audit errors), and unlike #1318 it does not change the error type in [lib/context#116](https://github.com/koajs/koa/pull/1318/files#diff-d911385a221c0b7502eabe1fdac9d209R116), which to me is a breaking change. However the solution is the same.

I dislike having to work around Jest in lib for it to pass but the only other option is to move away from Jest.

Referens from comments to Jest issue: https://github.com/facebook/jest/issues/2549

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -180,7 +180,11 @@ module.exports = class Application extends Emitter {
    */
 
   onerror(err) {
-    if (!(err instanceof Error)) throw new TypeError(util.format('non-error thrown: %j', err));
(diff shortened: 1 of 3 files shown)


dead-horse:
hmmm, I'd like to move away from jest and keep the code base simple. see https://github.com/koajs/koa/pull/1363

PR #1355, opened 2019-07-11
fix: typo

Jeff-Tian:


docs/api/response.md
```diff
@@ -17,7 +17,7 @@
 
 ### response.socket
 
-  Request socket.
```
dead-horse: actually `res.socket` and `req.socket` are point to the same `net.Socket` instance.

Jeff-Tian: Made another PR to reflect this fact: https://github.com/koajs/koa/pull/1357.    :alien:

Issue #1367, opened 2019-08-10
ctx.originalUrl invalid when used as express route


matthieusieben:
When using a Koa app as an express middleware, with a path prefix, the `ctx.originalUrl` will not be valid and will not contain the path prefix:

```js
const expressApp = express()
const koaApp = new Koa()
const koaRouter = new Router()

koaRouter.get('/test', async (ctx, next) => {
  console.log(ctx.originalUrl) // will not contain "/api"
  await next()
})

koaApp.use(koaRouter.routes())
expressApp.use('/api', koaApp.callback())
```

This is due to the fact that, behind the scene, the express router will manipulate `req.url` when the request is being processed by router middleware.

I would suggest that Koa does something similar to what express does when it is itself used as a middleware, meaning preserve the original `originalUrl`:
[https://github.com/expressjs/express/blob/f3fa758af9664526ee18c58764590e48f559e6ae/lib/router/index.js#L172](https://github.com/expressjs/express/blob/f3fa758af9664526ee18c58764590e48f559e6ae/lib/router/index.js#L172)

My suggestion is to change the following line of `lib/application.js`:
```js
  context.originalUrl = request.originalUrl = req.url;
  // to
  context.originalUrl = request.originalUrl = req.originalUrl || req.url;
```

Now one can argue that this would be a breaking change, especially because any route defined in koa will need to reflect the prefix again. In my example above, the routes in the router would have to be defined as `koaRouter.get("/api/test", ...)`. That is true, but we also need to consider that the current implementation **already** breaks the request `URL` related getters.

PR #1368, opened 2019-08-10
Use req.originalUrl when used as express middleware

matthieusieben:
Attempt at fixing https://github.com/koajs/koa/issues/1367


matthieusieben:
I agree, this is not the right way of doing this.

Issue #1370, opened 2019-08-11
Options in the constructor


djake:
Would you accept a pull request to allow `env` (and perhaps other settings) to be passed as a configuration argument to the `Application` constructor which would take precedence over process.env.NODE_ENV?

dead-horse:
hmm,  sounds good to me. there are several configurations we can pass from the constructor, it's more clear that what options koa accepted.

- proxy
- subdomainOffset
- env
- keys

```js
// before
const app = new Koa();
app.proxy = true;
app.subdomainOffset = 3;
app.keys =  [ key1, key2 ];
app.env = 'production';

// after
const app = new Koa({
  env: 'production',
  keys: [ key1, key2 ],
  proxy: true,
  subdomainOffset: 3,
})
```

PR #1372, opened 2019-08-15
accept options in the Application constructor

djake:
Accept an options object on the application constructor to configure app.env, app.proxy, app.keys, and app.subdomainOffset.

https://github.com/koajs/koa/issues/1370

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -34,13 +34,22 @@ module.exports = class Application extends Emitter {
    * @api public
    */
 
-  constructor() {
+  /**
+    *
+    * @param {object} [options] Application options
+    * @param {string} [options.env='development'] Environment
+    * @param {string[]} [options.keys] Signed cookie keys
+    * @param {string} [options.proxy] Trust proxy headers
+    * @param {number} [options.subdomainOffset] Subdomain offset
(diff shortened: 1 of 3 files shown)


dead-horse:
Update the document https://github.com/koajs/koa/blob/master/docs/api/index.md#application  ?

lib/application.js
```diff
@@ -34,13 +34,22 @@ module.exports = class Application extends Emitter {
    * @api public
    */
 
-  constructor() {
+  /**
+    *
+    * @param {object} [options] Application options
+    * @param {string} [options.env='development'] Environment
+    * @param {string[]} [options.keys] Signed cookie keys
+    * @param {string} [options.proxy] Trust proxy headers
```
fl0w: options.proxy is not a string, but a boolean

lib/application.js
```diff
@@ -34,13 +34,22 @@ module.exports = class Application extends Emitter {
...
+    * @param {number} [options.subdomainOffset] Subdomain offset
+    *
+    */
```
fl0w: For consistency, add a newline between `constructor`-method and `*/`.

PR #1374, opened 2019-08-17
test: fix end after end

ronag:
This fixes a broken test which is cause failure in the Node test suite (https://github.com/nodejs/node/pull/28666).

### DIFF
--- a/test/application/respond.js
+++ b/test/application/respond.js
@@ -234,6 +234,7 @@ describe('app.respond', () => {
       app.use((ctx, next) => {
         const res = ctx.res;
         ctx.status = 200;
+        ctx.respond = false;
         res.setHeader('Content-Type', 'text/html');
         res.write('Hello');
         setTimeout(() => res.end('Goodbye'), 0);
--- a/test/application/respond.js
+++ b/test/application/respond.js
@@ -234,10 +234,9 @@ describe('app.respond', () => {
       app.use((ctx, next) => {
         const res = ctx.res;
         ctx.status = 200;
-        ctx.respond = false;
         res.setHeader('Content-Type', 'text/html');
         res.write('Hello');
-        setTimeout(() => res.end('Goodbye'), 0);
+        res.end('Goodbye');
       });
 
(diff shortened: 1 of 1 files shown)


test/application/respond.js
```diff
@@ -234,6 +234,7 @@ describe('app.respond', () => {
       app.use((ctx, next) => {
         const res = ctx.res;
         ctx.status = 200;
+        ctx.respond = false;
```
dead-horse: It's not necessary here because there is no more middleware this application used. Also it is testing the app.respond method to handle this situation, if `ctx.respond` was set to false, we can't test this situation.

ronag: Unless this is set to false the test will call ‘res.end()’ twice and with data the second time. That will cause a write after end error in future versions of node...

ronag: See, https://github.com/nodejs/node/pull/28666

dead-horse: > Unless this is set to false the test will call ‘res.end()’ twice and with data the second time. That will cause a write after end error in future versions of node...

Then this test case will be meaningless. I think you should remove this line `setTimeout(() => res.end('Goodbye'), 0);` to fix this.

ronag: sorted

test/application/respond.js
```diff
@@ -236,7 +236,7 @@ describe('app.respond', () => {
         ctx.status = 200;
         res.setHeader('Content-Type', 'text/html');
         res.write('Hello');
-        setTimeout(() => res.end('Goodbye'), 0);
+        res.end('Goodbye');
```
dead-horse: I think we should remove `res.end()` here, because this test case want `application.respond` to handle the situation that response has already been written to but not end yet.

ronag:
```suggestion
```

ronag: fixed

ronag:
@Trott: I think we've fixed this in koa. Can we get another CITGM?

PR #1375, opened 2019-08-17
test: fix body test

ronag:
Setting body will set the content-length header. Unless the
corresponding number of bytes are sent the response will
be aborted and not emit 'end'.

### DIFF
--- a/test/application/respond.js
+++ b/test/application/respond.js
@@ -20,6 +20,7 @@ describe('app.respond', () => {
         res.statusCode = 200;
         setImmediate(() => {
           res.setHeader('Content-Type', 'text/plain');
+          res.setHeader('Content-Length', '3');
           res.end('lol');
         });
       });
@@ -41,6 +42,7 @@ describe('app.respond', () => {
         const res = ctx.res;
         res.statusCode = 200;
         res.setHeader('Content-Type', 'text/plain');
+        res.setHeader('Content-Length', '3');
         res.end('lol');
         ctx.set('foo', 'bar');
       });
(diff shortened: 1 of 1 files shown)


dead-horse:
What‘s the real problem this pull request fixed?

ronag:
The test depends on ‘end’ being emitted. However, since the content length doesn’t match node will emit ‘aborted’. I’m future version of node, end will not be emitted after aborted (that’s a bug).

ronag:
See https://github.com/nodejs/node/pull/27984

test/application/respond.js
```diff
@@ -20,7 +20,7 @@ describe('app.respond', () => {
         res.statusCode = 200;
         setImmediate(() => {
           res.setHeader('Content-Type', 'text/plain');
-          res.end('lol');
+          res.end('Hello');
```
dead-horse: The better way to fix this is add `res.setHeader('Content-Length', '3')` here to make it more clearly

ronag: sure, but I find that more confusing since you set `body = 'Hello'`

ronag: not quite sure what the purpose of `body = ` is? Maybe thats' what should be removed?

dead-horse: If we let koa handle the respond, then `ctx.body=` will both set content-length and body.

I think the test case want to test the situation that set respond manually can override koa's default behavior.

ronag: Fixed

dead-horse:
thanks!

PR #1384, opened 2019-09-27
fix: encode redirect url if not already encoded

fengmk2:
Same bug from express https://github.com/expressjs/express/commit/76eaa326ee8c4dda05568c6452286a16adb84c0b


fengmk2:
cc @dougwilson

dougwilson:
LGTM

fengmk2:
2.8.2

PR #1390, opened 2019-10-12
feat: support sameSite=none cookies

panva:
This bumps the `cookies` package dependency to add support for `sameSite` cookie attribute value `none`.


dead-horse:
2.10.0 released

PR #1392, opened 2019-10-13
ctx.response.get - use .getHeader

tinovyatkin:
`response.header` fails back to standard method `response.getHeaders` on Node >= 7.7 (i.e. on all live Node versions today), so, it overkill to to get all headers first, then do `field.toLowercase` (specially without typechecking of `field`).
[response.getHeader](https://nodejs.org/api/http.html#http_response_getheader_name) is a standard method since Node 0.x and it's doing all lowercasing internally.

Additionally, coercing to empty string is invalid here, as headers may contain non-string values before being sent to network, for example, Content-Length may be a number value of 0.


lib/response.js
```diff
@@ -413,12 +413,12 @@ module.exports = {
    *     // => "text/plain"
    *
    * @param {String} field
-   * @return {String}
+   * @return {any}
    * @api public
    */
 
   get(field) {
-    return this.header[field.toLowerCase()] || '';
```
dead-horse: LGTM, please add a test case for this fix

dead-horse:
it is related with https://github.com/koajs/koa/pull/1394

Because we change all headers to string when set, so get header default to empty string is reasonable.

I'm -0 on both change even in the next major version, the current implementation is also reasonable, the new implementation does not have much benefit but introduces some incompatible changes.

tinovyatkin:
@dead-horse new implementation do have benefits, indeed:

1. Empty string is a valid value for HTTP header (see https://github.com/eclipse/jetty.project/issues/1116 for example), and return it in case where header wasn’t set at all is an error.

2. It’s rare and nowhere documented that ctx.response.set(‘foo’, 5) !== ctx.response.get(‘foo’). It’s even more confusing taking in account that ctx.res.setHeader(‘foo’, 5) === ctx.res.getHeader(‘foo’)

3. Node built in response.getHeader documentation directly says that return type is the same as response.setHeader. Koa documentation says nothing about conversion to string. So, any developer relying on a fact that ctx.response.get must always return a string is relying on a bug, that may be and should be fixed in patch/minor version.

tinovyatkin:
More with current implementation:

```js
ctx.response.header['content-length'] = 0;
ctx.response.get('Content-Length); // => ''
```

fl0w:
FYI, I think it's better to rebase the branch rather than to merge in changes (though if you rebase and squash the result will be equivalent). Just to omit some commit bloat I tend to stay clear from merge-commits if possible or squash. Others might disagree.

tinovyatkin:
@fl0w that's the GitHub suggested way to `Update branch`, but anyway you can squash at merging, right?

anlexN:
@tinovyatkin koa team seems don't like merge, can we together fork koa and fix many problem? i think you are right.

3imed-jaberi (approved):

jonathanong (approved):

PR #1394, opened 2019-10-13
don't coerce header value to string on `set`

tinovyatkin:
Official Koa documentation about `response.set` says  _[This delegates to setHeader which sets or updates headers](https://github.com/koajs/koa/blob/master/docs/api/response.md#responsesetfields)_, but `setHeader` [doesn't convert header value types](https://nodejs.org/api/http.html#http_response_getheader_name) until sending them to the network. This PR fixes behavior of `response.set` to be according to the documentation.
Also removes recursion on `.set` calling `.setHeader` directly, for performance reason.


dead-horse:
This change is reasonable but may contains breaking change, I'd like to merge it when we start next  major version's development.

tinovyatkin:
@dead-horse remember that in current implementation you have two more _documented_ ways to set a response header:

```js
ctx.res.setHeader('Content-Length', 256);
ctx.response.header['content-length'] = 256;
```
I've linked above that official documentation clearly states that `ctx.response.set` delegates to `ctx.res.setHeader` while in fact, it's _not exactly_ the same. So, I don't understand why do you think it's a breaking change while, according to official documentation of Koa, it should be a bug fix.

dougwilson:
The current major has been out and in use long enough that at this point the bug is in the documentation for it. Seems reasonable to adjust it in next major as @dead-horse said.

anlexN:
good idea.
@dead-horse can we make koa v3 branch to develop?

PR #1395, opened 2019-10-13
emit more events

tinovyatkin:
Koa extends `EventEmitter` but doesn't use that fact for anything apart of `error` event.
However, there are lifecycle hooks that have real world use and otherwise hard to implement. This PR adds 3 events: 'request', 'respond' and 'responded' as well as documenting the fact that `Application` is an EventEmitter in more clear way.

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -219,6 +219,64 @@ Note:
 - Many properties on `ctx` are defined using getters, setters, and `Object.defineProperty()`. You can only edit these properties (not recommended) by using `Object.defineProperty()` on `app.context`. See https://github.com/koajs/koa/issues/652.
 - Mounted apps currently use their parent's `ctx` and settings. Thus, mounted apps are really just groups of middleware.
 
+## Events
+
+  Koa application is an instance of `EventEmitter` and emits following events to allow hooks into lifecycle.
+  All these events (except `error`) have `ctx` as first parameter:
+
+### Event: 'request'
+
+  Emitted each time there is a request, with `ctx` as parameter, before passing to any middleware.
+  May be a good place for per-request context mutation, when needed, for example:
+
+  ```js
+  app.on('request', ctx => {
+    ctx.state.start = Date.now();
+    // or something more advanced
+    if(!ctx.get('DNT')) ctx.state = new Proxy({})
+  })
+  ```
+
+### Event: 'respond'
+
+  Emitted after passing all middleware, but before sending the response to network.
+  May be used when some action required to be latest after all middleware processing, for example:
+
+  ```js
+  app.on('respond', ctx => {
+    if (ctx.state.start) ctx.set('X-Response-Time', Date.now() - ctx.state.start)
+  })
+  ```
+
+### Event: 'responded'
+
+  Emitted when the response stream is finished. Good place to cleanup any resources attached to `ctx.state` for example:
+
+  ```js
+  app.on('responded', ctx => {
+    if (ctx.state.dataStream) ctx.state.dataStream.destroy();
+  })
+  ```
+
+  More advanced example, use events to detect that server is idling for some time:
+
+  ```js
+    const onIdle10000ms = () => { console.warn('Server is idle for 10 seconds!'); }
+    const IDLE_INTERVAL = 10000;
+    let idleInterval = setInterval(onIdle10000ms, IDLE_INTERVAL);
+    app
+      .on('request', () => {
+        clearInterval(idleInterval);
+      })
+      .on('responded', () => {
+        idleInterval = setInterval(onIdle10000ms, IDLE_INTERVAL);
+      })
+  ```
+
+### Event: 'error'
+
+  See **Error Handling** below.
+
 ## Error Handling
 
   By default outputs all errors to stderr unless `app.silent` is `true`.
--- a/lib/application.js
+++ b/lib/application.js
@@ -155,6 +155,7 @@ module.exports = class Application extends Emitter {
    */
 
   handleRequest(ctx, fnMiddleware) {
+    this.emit('request', ctx);
     const res = ctx.res;
     res.statusCode = 404;
     const onerror = err => ctx.onerror(err);
@@ -218,18 +219,21 @@ function respond(ctx) {
   let body = ctx.body;
   const code = ctx.status;
 
+  ctx.app.emit('respond', ctx);
+  const onResponded = () => ctx.app.emit('responded', ctx);
+
   // ignore body
   if (statuses.empty[code]) {
     // strip headers
     ctx.body = null;
-    return res.end();
+    return res.end(onResponded);
   }
 
   if ('HEAD' == ctx.method) {
     if (!res.headersSent && isJSON(body)) {
       ctx.length = Buffer.byteLength(JSON.stringify(body));
     }
-    return res.end();
+    return res.end(onResponded);
   }
 
   // status body
@@ -243,12 +247,12 @@ function respond(ctx) {
       ctx.type = 'text';
       ctx.length = Buffer.byteLength(body);
     }
-    return res.end(body);
+    return res.end(body, onResponded);
   }
 
   // responses
-  if (Buffer.isBuffer(body)) return res.end(body);
-  if ('string' == typeof body) return res.end(body);
+  if (Buffer.isBuffer(body)) return res.end(body, onResponded);
+  if ('string' == typeof body) return res.end(body, onResponded);
   if (body instanceof Stream) return body.pipe(res);
 
   // body: json
@@ -256,7 +260,7 @@ function respond(ctx) {
   if (!res.headersSent) {
     ctx.length = Buffer.byteLength(body);
   }
-  res.end(body);
+  res.end(body, onResponded);
 }
 
 /**


dead-horse:
all these events can implemented in `application.handleRequest`:

```js
handleRequest(ctx, fnMiddleware) {
    const res = ctx.res;
    res.statusCode = 404;
    const handleResponse = () => {
      this.emit('respond', ctx);
      respond(ctx);
    }
    onFinished(res, err => {
      onerror(err);
      this.emit('responded', ctx);
    });
    this.emit('request', ctx);
    return fnMiddleware(ctx).then(handleResponse).catch(onerror);
  }
```

And I'm not sure if we really need to add the `respond` event which will make people confuse, I'd like to keep `request` and `responded`(then rename `responded` to `respond`):

```js
handleRequest(ctx, fnMiddleware) {
    const res = ctx.res;
    res.statusCode = 404;
    const handleResponse = () => respond(ctx);
    onFinished(res, err => {
      onerror(err);
      this.emit('respond', ctx);
    });
    this.emit('request', ctx);
    return fnMiddleware(ctx).then(handleResponse).catch(onerror);
  }
```

tinovyatkin:
I would like to keep pre-send event ('respond' or if you suggest to rename it), as a good way to adjust context after all middleware but before sending to the network (there are [some](https://github.com/koajs/response-time) [real world use](https://github.com/koajs/statsd) for this event), don't see how it can be confusing while being documented, but if you have a rename suggestion - please do!

fl0w:
I'm personally against this. This feels like the source for allot of confusion.

> as a good way to adjust context after all middleware but before sending to the network

This can't be true? Koa doesn't wait for the events to execute, it'll be at the mercy of event scheduling and conflicting with promise micro tasks? As I read this comment this would be the first case to support my first statement - confusing.

tinovyatkin:
@fl0w so, why Koa is an EventEmitter in first place?
The whole Koa concept is about mixing Promises and events that you agains, simple because Koa mixing streams and events.
Node.JS is actively moving forward to make it easy to mix Promises and events - see recently added methods in `events` module - like `finished`, `once`, etc. 
How this hurts and confuse?

tinovyatkin:
Event handler and `.emit` are sync by the way, so, EventEmitter will wait.

tinovyatkin:
@fl0w what is correct and safe way to clean up files without using any events in this example: https://github.com/koajs/koa/issues/1114#issuecomment-542229407 ?

fl0w:
Yea, sorry, I read this wrong.

lib/application.js
```diff
@@ -218,6 +225,8 @@ function respond(ctx) {
   let body = ctx.body;
   const code = ctx.status;
 
+  ctx.app.emit('respond', ctx);
```
dead-horse: this line is repeated

dead-horse:
need test cases for this feature

docs/api/index.md
```diff
@@ -219,6 +219,64 @@ Note:
...
+
+  Emitted when the response stream is finished. Good place to cleanup any resources attached to `ctx.state` for example:
+
+  ```js
+  app.on('responded', ctx => {
+    if (ctx.state.dataStream) ctx.state.dataStream.destroy();
+  })
+  ```
+
+  More advanced example, use events to detect that server is idling for some time:
```
fengmk2: This example is not correct on the concurrent requests scene.

Two requests are processing in the same time. The first `responded` emitted, after 10 seconds `'Server is idle for 10 seconds!'` but the second request is still processing.

lib/application.js
```diff
@@ -156,8 +156,15 @@ module.exports = class Application extends Emitter {
   handleRequest(ctx, fnMiddleware) {
     const res = ctx.res;
     res.statusCode = 404;
-    const onerror = err => ctx.onerror(err);
-    const handleResponse = () => respond(ctx);
+    const onerror = err => {
+      if (null != err) ctx.onerror(err);
```
fengmk2: This is breaking change.

tinovyatkin: What does it breaks?!

docs/api/index.md
tinovyatkin: You are wrong, sorry, Node is not multithreading.

fengmk2: It's not a  multithreading, the concurrent requests scene will happen in Node.
The first request `responded` your code will start a timeout to wait for `onIdle` execute.
If the second request's `responded` event still not emit, then the `onIdle` will be executed and stdout will show log "Server is idle". But this log is not true, because the second request is still processing.

lib/application.js
```diff
@@ -156,9 +156,23 @@ module.exports = class Application extends Emitter {
...
     const res = ctx.res;
     res.statusCode = 404;
-    const onerror = err => ctx.onerror(err);
-    const handleResponse = () => respond(ctx);
+
+    const responding = Symbol('responding');
+    this.once(responding, () => this.emit('respond', ctx));
+
+    const onerror = err => {
+      if (null != err) {
```
fengmk2: the old code will execute `onerror` no matter the err is null or not. So I think this is a breaking change.

tinovyatkin: @fengmk2 please give yourself a pleasure to check first line of `onerror`. What kind of code review is that?!!

docs/api/index.md
```diff
@@ -221,6 +221,68 @@ Note:
 - Many properties on `ctx` are defined using getters, setters, and `Object.defineProperty()`. You can only edit these properties (not recommended) by using `Object.defineProperty()` on `app.context`. See https://github.com/koajs/koa/issues/652.
 - Mounted apps currently use their parent's `ctx` and settings. Thus, mounted apps are really just groups of middleware.
 
+## Events
```
jonathanong, line 224: thanks for this doc, very useful for a PR

docs/api/index.md
```diff
@@ -221,6 +221,68 @@ Note:
...
+
+  ```js
+  app.on('request', ctx => {
+    ctx.state.start = Date.now();
+    // or something more advanced
+    if(!ctx.get('DNT')) ctx.state = new Proxy({})
+  })
+  ```
+
+### Event: 'respond'
```
jonathanong, line 242: i don't see how this or the `request` event have any benefit over regular middleware

jonathanong: it actually has less benefit because you cannot use async functions

docs/api/index.md
```diff
@@ -221,6 +221,68 @@ Note:
...
+  Emitted after passing all middleware, but before sending the response to network.
+  May be used when some action required to be latest after all middleware processing, for example:
+
+  ```js
+  app.on('respond', ctx => {
+    if (ctx.state.start) ctx.set('X-Response-Time', Date.now() - ctx.state.start)
+  })
+  ```
+
+  Note: `respond` event may not be emitted in case of premature socket close (due to a middleware timeout, for example).
```
jonathanong, line 253: exceptions like this make it less useful and more difficult to maintain

docs/api/index.md
```diff
@@ -221,6 +221,68 @@ Note:
...
+
+### Event: 'responded'
+
+  Emitted when the response stream is finished. Good place to cleanup any resources attached to `ctx.state` for example:
+
+  ```js
+  app.on('responded', ctx => {
+    if (ctx.state.dataStream) ctx.state.dataStream.destroy();
```
jonathanong, line 261: i don't understand this use-case. clean up callbacks on streams should be handled when you create and pipe them.

docs/api/index.md
```diff
@@ -221,6 +221,68 @@ Note:
...
+  })
+  ```
+
+  More advanced example, use events to detect that server is idling for some time:
+
+  ```js
+    const server = app.listen();
```
jonathanong, line 268: i would actually recommend not doing this in koa. i would do this as a req/res decorator:

```js
const fn = app.callback()

const checkIdle = (fn) => (req, res) => {
  // do stuff
  return fn(req, res)
}

http.createServer(checkIdle(fn))
```

docs/api/index.md
tinovyatkin: How would you know from the stream that, for example, connection with a client was prematurely lost and not the whole stream was consumed?

docs/api/index.md
tinovyatkin: It's not an exception - it's a logic of this event because the response in this case never happened.

docs/api/index.md
tinovyatkin: There is no need to have advantage or disadvantage and not always sync call is inferior to async call from application point of view. If you follow Node.JS core development strategy then you certainly will notice “mix and match” innovation in handling async flow - `once` was added into `events` core module, `finished`, `Readable.from(AsyncIterator)`, and `Readable[AsyncIterator]` into the `stream` - so, modern Node.JS is inviting to pick events, streams or Promises whatever solves problem more efficiently OR USE THEM ALTOGETHER. If Koa is the EventEmitter why it doesn’t emit any lifecycle events apart of `error` and why event handlers should be considered inferior?

docs/api/index.md
tinovyatkin: I probably missing something here as I’m struggling to understand: if you created Koa as an extension of `EventEmitter` in the first place, why do you consider using events is antipattern and inferior to writing low-level decorator here?

PR #1397, opened 2019-10-14
implement response.has

tinovyatkin:
Implement (and use) `response.has` to match Node' `response.hasHeader`


lib/response.js
```diff
@@ -421,6 +421,29 @@ module.exports = {
...
+   *     // => true
+   *
+   *     this.get('content-type');
+   *     // => true
+   *
+   * @param {String} field
+   * @return {boolean}
+   * @api public
+   */
+  has(field) {
```
dead-horse: please add a test case for this feature

PR #1398, opened 2019-10-14
use response.get to get just one header

tinovyatkin:
It's more semantical to use `response.get` to get just one header and if #1392 will be merged it's also will be more performant.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -204,7 +204,7 @@ module.exports = {
     const len = this.get('Content-Length');
     const body = this.body;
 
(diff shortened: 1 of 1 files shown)


lib/response.js
```diff
@@ -201,10 +201,10 @@ module.exports = {
    */
 
   get length() {
-    const len = this.header['content-length'];
+    const len = this.get('Content-Length');
```
dead-horse: merge this change inot https://github.com/koajs/koa/pull/1399?

PR #1399, opened 2019-10-14
don't need isJSON here

tinovyatkin:
The logic at `response.length` is almost [completely repeats](https://github.com/koajs/is-json/blob/master/index.js#L9) whole `is-json` module content. Adding just one more comparison to `Stream` will eliminate need for external dependency that basically repeats the same checks - so, it's both more performant and safe.


lib/response.js
```diff
@@ -204,12 +204,11 @@ module.exports = {
     const len = this.header['content-length'];
     const body = this.body;
 
-    if (null == len) {
-      if (!body) return;
-      if ('string' == typeof body) return Buffer.byteLength(body);
+    if (!len) {
```
dead-horse: I'd like don't change this line, it may introduce unnecessary risks.

PR #1400, opened 2019-10-14
removes code duplication at handling HEAD method

tinovyatkin:
Handling of `HEAD` method sets content length for JSON body, repeating the measurement logic from `response.length`. This PR fixes that, removes code duplication and centralising response body length calculation at `response.length`, also removing external dependency here.
If #1399 will also be merge it will allow to remove `is-json` dependency completely.


lib/application.js
```diff
@@ -225,9 +224,10 @@ function respond(ctx) {
     return res.end();
   }
 
-  if ('HEAD' == ctx.method) {
-    if (!res.headersSent && isJSON(body)) {
-      ctx.length = Buffer.byteLength(JSON.stringify(body));
+  if ('HEAD' === ctx.method) {
+    if (!res.headersSent) {
```
dead-horse: please add a test case with plain text

dead-horse: actually we already set `ctx.response.length` for plain text in [set body()](https://github.com/koajs/koa/blob/5e22f9c8a02d23bf69d76a6f2c03ca69e1b11841/lib/response.js#L135), only expect for object body(for performance reason).

tinovyatkin:
Actually, I probably formulated bad the objective of this PR: I would like to remove body type checking and size calculation from `HEAD` handling at `Application` and have it just at one place - at `response.length`. No extra tests needed as HEAD already covered by existing tests.

lib/application.js
```diff
@@ -225,9 +224,10 @@ function respond(ctx) {
     return res.end();
   }
 
-  if ('HEAD' == ctx.method) {
-    if (!res.headersSent && isJSON(body)) {
-      ctx.length = Buffer.byteLength(JSON.stringify(body));
+  if ('HEAD' === ctx.method) {
+    if (!res.headersSent && !ctx.response.has('Content-Length')) {
+      const { length } = ctx.response;
+      if (Number.isInteger(length)) { ctx.length = length; }
```
dead-horse: I'd like to remove `{}` in single line, others LGTM

tinovyatkin: Fixed. Also calling `.hasHeader` directly on `res` here to make the logic here clearer and be in line with `res.headersSent` check in the same condition. 
Btw, did the team discussed the introduction of [Prettier](https://prettier.io/docs/en/why-prettier.html) on codebase to avoid code style reviews? I can make a PR for it 🙋🏻‍♂️

fl0w: I would prefer [Koa eslint config](https://github.com/koajs/eslint-config-koa) to be updated if anything.

PR #1405, opened 2019-10-15
handling Objects for streaming bodies

tinovyatkin:
Koa currently throws while attempting to pipe an object mode stream into HTTP response.
This PR fixes that behavior allowing to serve that streams as JSON body.

Node.js 12.3 introduced a [documented way to check that a stream in object mode](https://nodejs.org/api/stream.html#stream_readable_readableobjectmode) stream, so, I think Koa must support such streams as first-class citizens out of the box.

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -256,6 +256,7 @@ function respond(ctx) {
         (body._readableState &&
           body._readableState.objectMode)) {
       let first = true;
+      res.write('[');
       body = body
         .pipe(new Stream.Transform({
           writableObjectMode: true,
@@ -263,7 +264,7 @@ function respond(ctx) {
           transform(data, encoding, callback) {
             const str = JSON.stringify(data);
             if (first) {
-              this.push(`[${str}`);
+              this.push(`${str}`);
               first = false;
             } else {
               this.push(`,${str}`);
@@ -271,7 +272,7 @@ function respond(ctx) {
             return callback();
           },
           flush(callback) {
-            if (!first) this.push(`]`);
+            this.push(`]`);
             this.push(null);
             return callback();
           }
--- a/lib/application.js
+++ b/lib/application.js
@@ -262,13 +262,10 @@ function respond(ctx) {
           writableObjectMode: true,
           readableObjectMode: false,
           transform(data, encoding, callback) {
-            const str = JSON.stringify(data);
-            if (first) {
-              this.push(`${str}`);
-              first = false;
-            } else {
-              this.push(`,${str}`);
-            }
+            if (!first) this.push(',');
+            else first = false;
+
+            this.push(JSON.stringify(data));
             return callback();
           },
           flush(callback) {
--- a/lib/application.js
+++ b/lib/application.js
@@ -249,7 +249,36 @@ function respond(ctx) {
   // responses
   if (Buffer.isBuffer(body)) return res.end(body);
   if ('string' == typeof body) return res.end(body);
-  if (body instanceof Stream) return body.pipe(res);
+  if (body instanceof Stream) {
+    // check if it's an objectMode stream
+    // readableObjectMode is available since Node 12.3
+    if (body.readableObjectMode ||
+        (body._readableState &&
+          body._readableState.objectMode)) {
+      let first = true;
+      body = body
+        .pipe(new Stream.Transform({
+          writableObjectMode: true,
+          readableObjectMode: false,
+          transform(data, encoding, callback) {
+            const str = JSON.stringify(data);
+            if (first) {
+              this.push(`[${str}`);
+              first = false;
+            } else {
+              this.push(`,${str}`);
+            }
+            return callback();
+          },
+          flush(callback) {
+            if (!first) this.push(`]`);
+            this.push(null);
+            return callback();
+          }
+        }));
+    }
+    return body.pipe(res);
+  }
 
   // body: json
   body = JSON.stringify(body);
--- a/lib/application.js
+++ b/lib/application.js
@@ -273,7 +273,7 @@ function respond(ctx) {
             return callback();
           },
           flush(callback) {
-            this.push(`]`);
+            this.push(']');
             this.push(null);
             return callback();
           }
--- a/lib/response.js
+++ b/lib/response.js
@@ -166,14 +166,21 @@ module.exports = {
     }
 
     // stream
-    if ('function' == typeof val.pipe) {
+    if (val instanceof Stream) {
       onFinish(this.res, destroy.bind(null, val));
       ensureErrorHandler(val, err => this.ctx.onerror(err));
 
       // overwriting
       if (null != original && original != val) this.remove('Content-Length');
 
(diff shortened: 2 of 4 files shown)


dead-horse:
I'm not sure there is a real world example that need this feature, and people can transform object streams to normal streams and used with koa. I'd prefer not implement it in koa core and leave it to community.

tinovyatkin:
@dead-horse I think the wording _transform object streams to normal streams_ is the key to understand what's wrong with not having object streams in Koa core from the initial Koa architecture point of view.

Object streams are normal streams, if under _normal_ you mean that they are created by Node built-in `streams` module and have all events, functions, behavior of _normal_ stream. The only difference here that chunk of data for non-object mode stream can be string or Buffer, and objectMode adds one more type of chunk - Object. Does it sounds similar to something, right? Yes, it's like standard Node `response.end` type expectation - it must be a string or Buffer. 

The simplicity `ctx.body = stream` may confuse some developers, that Stream is just a third type of data that you can assign to body. While in fact, it's another _way_ to serve **the same types data** - it's sync `response.end` vs streaming `response.pipe`.

Let's see what Koa, as a framework, adding and promising to a developer here:

## Sync (ctx.body = string, ctx.body = buffers; ctx.body = object)
Sync way to pass data simply delegates to standard `response.end(data)`.
`response.end` can handle on it's own two types of data - string or Buffer. Passing anything other that that will result in underlying error that `string or Buffer value expected`.
What Koa added here? Right, handling third type of data - `Object`. Koa detects that `ctx.body` is an object not equaling to already supported string or Buffer, sets `Context-Type` to `json` and converts object into supported data type - string. Everything perfect.

## Streaming (ctx.body = stream)
Streaming way to serve response delegates to `response.pipe`.  What `response.pipe` is doing internally with data? Well, it adds listener for `.on(data`) and then delegates it to sync `response.write`. 
`response.write` handles the data the same way as `response.end`. That means if data chunk from `stream.on(data)` is not a string or Buffer you will see _exactly the same error message_ as passing something to `response.end` - `string or Buffer value expected`. So, why Koa on streaming way to deliver body doesn't do exactly the same magic as on sync way of delivering body? I mean, check the data type, handle Content-Type, stringify it.

So, what happened here, that Koa declares it's support for easy JSON handling, but while fully supporting _two ways_ of delivering data - sync and stream, actually supports object-to-JSON conversion for response, for some unknown reason, only for synchronous way of delivering response.

Object mode streams didn't feel like _first-class citizens_ in early Node versions, so, we've seen not so much real world use cases. But now, with documented `stream.readableObjectMode` / `stream.writableObjectMode`, convenience of async iterators `for await (const object of stream)`, new helper methods from `events` module like `finished` and `once`, they certainly will soon become way more popular as memory effective async way of handing big amounts of data. Real-word use today [also exists](https://stackoverflow.com/questions/20058614/stream-from-a-mongodb-cursor-to-express-response-in-node-js) as many streaming sources of data.

Supporting object for sync bodies is ever simpler:
```js
ctx.type = 'json'
ctx.body = JSON.stringify(object)
```

But Koa founders added that two lines into core, without delegating to community. They just forget to do the same on streaming way of delivering response, I assume due to lack of documentation or popularity of object mode streams.

So, this PR (20 lines of code) is not adding new feature, nor handling `non-normal` streams - it closes the gap of original Koa promise - you can assign objects to response and Koa will do the rest of magic.

Contrary to merging this PR will be adding to documentation some warning:
```md
Koa supports, like standard `http.ServerResponse`, two ways to set response body - sync and streaming. Also like standard response it supports two type of data - string or Buffer. It also adds support for any object via JSON.stringify, but only for sync way of setting body.
```

dead-horse:
Better to use module like https://github.com/stream-utils/streaming-json-stringify to do the stringify, also need to set `content-type` to json by default.

tinovyatkin:
@dead-horse the module you suggesting wasn't updated in last 3 years (so, not sure it will be maintained) and depends on two more external modules (`readable-stream/transform` and `json-stringify-safe`), so, it will be way more _non-normal_ stream and may be performance penalty. I don't want to bring a lot of third-party dependencies in just for 20 lines of code. 
Node.JS world is [kinda moving from using a lot of one-liner modules](https://paulmillr.com/posts/chokidar-3-save-32tb-of-traffic/) practice for some serious reasons.

Also, my next PR will be exposing a way to stringify body (ie, create `application.stringify = JSON.stringify.bind(JSON)` and use everything stringifing is happening.

tinovyatkin:
This PR sets content-type to `json` for object mode streams by default, see changes to `response.js` file

dead-horse:
Actually I'm not care about import some "one-liner module", many "one-liner module" have hundreds lines of test code. I don't want to re-implement these code in koa's code base.

`streaming-json-stringify` can be improved if the implement is outdated, I think @dougwilson and @jonathanong will be happy to accept pull requests with this module.

> Also, my next PR will be exposing a way to stringify body (ie, create application.stringify = JSON.stringify.bind(JSON) and use everywhere when stringifying is happening, so, again, it's better to have control over stringifiyng in core.

I'm not sure this is really helpful, change stringify method seems useless to 99.99% of developers. Even though `streaming-json-stringify` can specific JSON.stringify by options.

tinovyatkin:
@dead-horse There is [open and approved PR](https://github.com/stream-utils/streaming-json-stringify/pull/28) in `streaming-json-stringify` since 2017. I don't think it's still maintained.

The code of my implementation is completely different and easier (no extra options, etc), so, it will be rewrite of `streaming-json-stringify`, not an update.

dougwilson:
> There is open and approved PR in streaming-json-stringify since 2017. I don't think it's still maintained.

I mean, Koa itself has an "open and approved PR since 2017" (https://github.com/koajs/koa/pull/1011) so by that measurement Koa is not maintained, either :)

tinovyatkin:
@dead-horse found very similar to mine implementation - https://github.com/big-kahuna-burger/json-stringify-trough. Are you OK to introduce such dependency?
100% test coverage, no more dependencies.

tinovyatkin:
Nope, it's not exactly the same. Main difference between my implementation and any module I've found on the internet that in case of no data my implementation will output empty array (that's correct implementation of empty stream in my opinion and will be parsed well if content-type is `json`) while the rest outputs nothing.
Are you sure we need to move this functionality into an external module? I can publish one, but...

tinovyatkin:
As I wrote above, we soon will see more libraries adding support for object stream. Here [another example of recent update of very popular open source](https://github.com/websockets/ws/issues/1640) library adding `objectMode` streams support.

jonathanong:
this belongs in a separate module. i can give you access to https://github.com/koajs/koa/issues/281 if you want @tinovyatkin

tinovyatkin:
@jonathanong I would love to refactor `streaming-json-stringify` and reimplement this PR using that module. Could you give me access, please?

jonathanong:
done

lagden:
@tinovyatkin you can use http://oboejs.com/
it is a good lib!

PR #1406, opened 2019-10-15
prefer standard functions over external dependencies

tinovyatkin:
This PR fixes three issues, two happens to be related to generator functions middleware. 
While this functionality will be removed completely in Koa 3.0, the next major version is not on the radar yet (unfortunately), but these small changes will make Koa a little bit faster and safer _today_.

1. Use `util.types.isGeneratorFunction` when available (Node >= 10) instead of loading external dependency.

2. Remove `depd` dependency and use `process.emitWarning`. `depd` is huge for the purpose [install size](https://packagephobia.now.sh/result?p=depd) - it's 500+ lines of code (twice as big as Koa' `application.js`) and used just once - to deprecate generator middleware. `process.emitWarning` coupled with `--no-deprecation` and `--trace-deprecation` command-line flags resulting in the same functionality while having one less external dependency to load.

3. Use standard (since Node >= 10), `Stream.finished` instead of external `on-finished` module when supported. This required small change in context mocking to actually create stream (I think that was necessary anyway for correctness)

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -5,7 +5,6 @@
  * Module dependencies.
  */
 
-const isGeneratorFunction = require('is-generator-function');
 const debug = require('debug')('koa:application');
 const onFinished = require('on-finished');
 const response = require('./response');
@@ -15,6 +14,9 @@ const request = require('./request');
 const statuses = require('statuses');
 const Emitter = require('events');
 const util = require('util');
+const isGeneratorFunction = util.types
+  ? util.types.isGeneratorFunction
+  : require('is-generator-function');
 const Stream = require('stream');
 const http = require('http');
 const only = require('only');
--- a/lib/application.js
+++ b/lib/application.js
@@ -6,7 +6,6 @@
  */
 
 const debug = require('debug')('koa:application');
-const onFinished = require('on-finished');
 const response = require('./response');
 const compose = require('koa-compose');
 const context = require('./context');
@@ -18,6 +17,7 @@ const isGeneratorFunction = util.types
   ? util.types.isGeneratorFunction
   : require('is-generator-function');
 const Stream = require('stream');
+const onFinished = Stream.finished || require('on-finished');
 const http = require('http');
 const only = require('only');
 const convert = require('koa-convert');
--- a/lib/application.js
+++ b/lib/application.js
@@ -21,7 +21,6 @@ const Stream = require('stream');
 const http = require('http');
 const only = require('only');
 const convert = require('koa-convert');
-const deprecate = require('depd')('koa');
 const { HttpError } = require('http-errors');
 
 /**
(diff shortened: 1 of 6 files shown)


lib/application.js
```diff
@@ -116,9 +117,13 @@ module.exports = class Application extends Emitter {
   use(fn) {
     if (typeof fn !== 'function') throw new TypeError('middleware must be a function!');
     if (isGeneratorFunction(fn)) {
-      deprecate('Support for generators will be removed in v3. ' +
-                'See the documentation for examples of how to convert old middleware ' +
-                'https://github.com/koajs/koa/blob/master/docs/migration.md');
+      process.emitWarning([
```
dougwilson: https://nodejs.org/api/process.html#process_avoiding_duplicate_warnings

tinovyatkin: I've seen that. But this refers to _the same_ warning and this warning will be emitted once per middleware (and I've added middleware name of first line of code for convenience). So, it's completely in line with Node documentation suggestion - _the same_ warning emits just once per process.

dougwilson: Many times, you may have a graceful restart in the app where the Koa app is in a factory, and the same app may be constructed multiple times in the same process. This would emit a warning for each use for each time the same app is constructed. Perhaps then if the idea is to emit once per name, then track which names have already been emitted instead.

tinovyatkin: Isn't that too much functionality for deprecated feature? I can do it for sure, but don't want to have another 500+ lines of code :-)

dougwilson: Then why not just emit it only once? You said the idea is to emit once per generator, but I showed that that is not the case. What exactly is the case, then? This seems to emit unconditionally forever, against the Node.js guidelines for usage??

tinovyatkin: The original deprecations message will also appear several time in cases you suggesting (creating several Koa instances in the same process). So, we may be not at the _best practices_ of Node in that edge case, but definitely is no way more noise than with external dependency.

dougwilson: > The original deprecations message will also appear several times in cases you suggesting (creating several Koa instances in the same process).

No it doesn't. Can you provide code that demonstrates otherwise?

tinovyatkin: My bad, I've actually didn't set `DeprecationWarning` type, fixing now

lib/application.js
```diff
@@ -5,21 +5,22 @@
...
 const context = require('./context');
 const request = require('./request');
 const statuses = require('statuses');
 const Emitter = require('events');
 const util = require('util');
+const isGeneratorFunction = util.types
+  ? util.types.isGeneratorFunction
+  : require('is-generator-function');
 const Stream = require('stream');
+const onFinished = Stream.finished || require('on-finished');
```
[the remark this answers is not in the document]
tinovyatkin: All tests are passing. It's actually a little bit better regarding handling premature stream termination and Node docs directly gives that as correct way to handle streaming HTTP bodies.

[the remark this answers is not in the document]
dougwilson: I think the main difference is that `Stream.finished` will not execute the callback if the passed in stream was already finished. For example the client closes the TCP connection between when Koa picked it up and when this listener gets added. With `Stream.finished` you need to handle that yourself by checking if it's already finished.

I'm sure there are other differences, and they are not clones of each other (the `Stream.finished` is the `end-of-stream` module included into core Node.js). Koa tests likely assume the expected behavior and do not retest all the conditioned in the `on-finished` test suite, so Koa's own tests are unlikely to show some of the differences as a safety net for the change, unfortunately.

lib/application.js
```diff
@@ -5,21 +5,22 @@
...
 const debug = require('debug')('koa:application');
-const onFinished = require('on-finished');
 const response = require('./response');
 const compose = require('koa-compose');
 const context = require('./context');
 const request = require('./request');
 const statuses = require('statuses');
 const Emitter = require('events');
 const util = require('util');
+const isGeneratorFunction = util.types
```
jonathanong, line 16: we should just remove this check in the next major

anlexN:

fengmk2:
closed by https://github.com/koajs/koa/commit/b747e16d75554203a28e55d87296e81f68b992f5

PR #1408, opened 2019-10-16
feat: support sameSite=none cookies (bump cookies dependency) (#1390)

Elias-Serneels:
Updated cookie dependency on v1 branch to support sameSite=none cookies.

This change was already applied to the current (v2) but we needed this same functionality for v1.


dead-horse:
koa@1.7.0 released

PR #1409, opened 2019-10-17
remove `error-inject` and fix error handling

tinovyatkin:
The whole `error-inject` functionality (it's [3 lines actually](https://github.com/stream-utils/error-inject/blob/2070e23a55611a19f20e41bd88c3bcbb79b160d2/index.js#L4)) is to add `error` event listener as `on` listener, while **working around the fact that Node doesn't check if that handler already exists**, so, that check was added.

However, by obscuring this logic in 3 lines module even module author @dead-horse probably forgot what exactly his module is doing while introducing #800, which basically rendered this check useless by creating inline function.

To prevent this to happen again, and also taking into account that `ctx.onerror` calls `res.end` and so, `.once` is the more appropriate listener, I'm proposing removing `error-inject` dependency and use the fact that respond body assignment already has logic to make something once per stream.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -6,7 +6,6 @@
  */
 
 const contentDisposition = require('content-disposition');
-const ensureErrorHandler = require('error-inject');
 const getType = require('cache-content-type');
 const onFinish = require('on-finished');
 const escape = require('escape-html');
@@ -166,12 +165,14 @@ module.exports = {
     }
 
     // stream
-    if ('function' == typeof val.pipe) {
+    if (val instanceof Stream) {
       onFinish(this.res, destroy.bind(null, val));
-      ensureErrorHandler(val, err => this.ctx.onerror(err));
 
-      // overwriting
-      if (null != original && original != val) this.remove('Content-Length');
+      if (original != val) {
+        val.once('error', err => this.ctx.onerror(err));
+        // overwriting
+        if (null != original) this.remove('Content-Length');
+      }
 
       if (setType) this.type = 'bin';
       return;
--- a/test/response/body.js
+++ b/test/response/body.js
@@ -4,6 +4,7 @@
 const response = require('../helpers/context').response;
 const assert = require('assert');
 const fs = require('fs');
+const Stream = require('stream');
 
 describe('res.body=', () => {
   describe('when Content-Type is set', () => {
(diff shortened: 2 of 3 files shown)


lib/response.js
```diff
@@ -166,12 +165,14 @@ module.exports = {
...
     // stream
-    if ('function' == typeof val.pipe) {
+    if (val instanceof Stream) {
       onFinish(this.res, destroy.bind(null, val));
-      ensureErrorHandler(val, err => this.ctx.onerror(err));
 
-      // overwriting
-      if (null != original && original != val) this.remove('Content-Length');
+      if (original != val) {
+        val.once('error', err => this.ctx.onerror(err));
```
dead-horse: The two implementations seem to be not equal. We should only add `error` event listener when the stream has no other error event listener, which aim to avoid uncaught exception.

Also IMO koa's code philosophy is to keep code simple, so we don't treat 'one-liner module' as a bad thing.

tinovyatkin: In current implementation the check in the `error-inject` does nothing (as to inline created functions never the same) and just always add `.on('error')` handler. Do you mean that was desired behaviour?
I'm proposing to remove `error-inject` here not because it's one line, but because even you (module author) forgot that you breaking it functionality while refactored to inline function. It just does nothing in current version.

tinovyatkin:
> I don't think we should change this in any reason.

Ok, just left the code that was erroneously refactored and doing nothing now.

lib/response.js
jonathanong: i don't understand what you mean, can you explain in a test?

jonathanong: seems like a developer error

tinovyatkin: It is, but the most probably explanation here that by hiding the actual logic in a one-line module even the author of the module forgot the logic on next refactoring and broke the functionality. That's what this PR is all about - fix the error and bring the logic inside to prevent such a problem in the future.

test/response/body.js
```diff
@@ -108,6 +109,16 @@ describe('res.body=', () => {
       res.body = fs.createReadStream('LICENSE');
       assert.equal('application/octet-stream', res.header['content-type']);
     });
+
+    it('should add error handler to the stream, but only once', () => {
```
dead-horse, line 113: This test case clearly shows where the problem is, LGTM

PR #1430, opened 2020-01-07
fix: for-in loop is missing hasOwnProperty check

LFB:
Inspection info checks for unfiltered for-in loops, The use of this construct results in processing inherited or unexpected properties.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -463,7 +463,9 @@ module.exports = {
       this.res.setHeader(field, val);
     } else {
       for (const key in field) {
-        this.set(key, field[key]);
+        if (field.hasOwnProperty(key)) {
+          this.set(key, field[key]);
+        }
       }
     }
   },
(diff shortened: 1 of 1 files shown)


lib/response.js
```diff
@@ -463,7 +463,9 @@ module.exports = {
       this.res.setHeader(field, val);
     } else {
       for (const key in field) {
-        this.set(key, field[key]);
+        if (Object.prototype.hasOwnProperty.call(field, key)) {
```
dead-horse: It can be a breaking change, why you need to add this check?

LFB: Hello, the reason I want to modify is: checks for unfiltered for-in loops, The use of this construct results in processing inherited or unexpected properties. Make the code more robust.

dead-horse: Sorry, if there is no actual scene, we will not modify it here, because it runs the risk of breaking other people's code

LFB: Got it, thank you

PR #1445, opened 2020-04-18
Document response status with empty body

madarche:
Hello,

This is a very simple addition to the docs to explicitly document how to return a status code different than 400 when one doesn't set any body in the response.

As a new Koa user (coming from Express) this is a topic that has puzzled me at the beginning. Of course after having used Koa for a little while, the answer seems obvious. But I believe this addition will speed up even more the learning curve.


madarche:
Thanks @dead-horse. And thanks for Koa.js which is great.

PR #1447, opened 2020-04-20
feat: allow bodyless responses for non empty status codes

### LINKED ISSUE #1446: Preserve given status when response body is null
Koa should preserve status to a greater extent, as it's replacing what the user added if certain situations happens (null/undefined body)

Suppose a basic app with no middlewares

```ts
app.use((ctx, next) => {
  ctx.status = 404;
});
```
This sets the body to the plain text "Not Found", however there's no need to place any redundancy, as 404 already implies that, and when there's no useful info to return it's better to omit the body completely. However if you explicitly set the body to null, koa changes the status to 204 automatically

```ts
app.use((ctx, next) => {
  ctx.status = 404;
  ctx.body  = null; // same case when setting ctx.body = undefined
});
```

Of course 204 is no expected response for a "not found" response, so the only way to proceed is sending a redundant body. I'm aware of #1421 but this is not the same case, I don't want to return null or 'null', there shouldn't be any body at all (like 204 does).

ejose19:
Fixes #1446

Either "Content-Length" must be left (resulting in 0) or we need to add "Connection: close" header as well. The former is what it's in this PR, but opinions are welcomed if the later is better.

If approved will add test cases. Usage:

```js
app.use((ctx, next) => {
  ctx.body = null; // Must be set first, as this sets status to 204 automatically
  ctx.status = 404;
});
```

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -238,6 +238,11 @@ function respond(ctx) {
 
   // status body
   if (null == body) {
+    if (ctx.response._explicitNullBody) {
+      ctx.response.remove('Content-Type');
+      ctx.response.remove('Transfer-Encoding');
+      return res.end();
+    }
     if (ctx.req.httpVersionMajor >= 2) {
       body = String(code);
     } else {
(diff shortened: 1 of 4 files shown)


dead-horse:
I'm OK with this change, can you add a test case for this?

ejose19:
@dead-horse Done, should be good now.

PR #1450, opened 2020-04-24
add call method

shenchao591:
for this pr,[https://github.com/koajs/compose/pull/104](url)


dead-horse:
can you provide more informations for this change?

jonathanong:
@dead-horse this adds `this` support in middleware

@shenchao591 can you add a test?

dead-horse:
> this adds this support in middleware

because we prefer to use arrow functions which will lose `this` in middleware, I think a uniform use of `ctx` is a better choice.

dead-horse:
close for now.

PR #1455, opened 2020-05-01
feat: support asyncLocalStorage

### LINKED ISSUE #1453: AsyncLocalStorage support
https://nodejs.org/api/async_hooks.html#async_hooks_class_asynclocalstorage

we should run every request in it so that errors can be traced to the original req/res even if they were called in a different event loop. this will also allow us to not pass req/res/ctx to track sources of errors.

the `store` would be `ctx`

```js
const ctx = app.asyncLocalStorage.getStore()
if (ctx) // ...
```

dead-horse:
closes #1453 

- `options.asyncLocalStorage` to enable
- `app.ctxStorage` to get the asyncLocalStorage instance
- `app.currenctContext` to get currenct context
- node 13+ required

Unfortunately, there is a large performance loss after opening. The deeper the async call stack, the greater the performance loss.

- without asyncLocalStorage
```shell
  1 middleware
  21312.11

  5 middleware
  20286.01

  10 middleware
  20702.86

  15 middleware
  21740.76

  20 middleware
  21164.56

  30 middleware
  21155.04

  50 middleware
  20994.66

  100 middleware
  19356.09

  1 async middleware
  21410.63

  5 async middleware
  21537.31

  10 async middleware
  19848.90

  15 async middleware
  19031.05

  20 async middleware
  18554.31

  30 async middleware
  19903.52

  50 async middleware
  18689.10

  100 async middleware
  17426.28
```

- with asyncLocalStorage
```shell
  1 middleware
  17065.02

  5 middleware
  16251.25

  10 middleware
  15339.71

  15 middleware
  16713.63

  20 middleware
  14572.50

  30 middleware
  13395.76

  50 middleware
  14125.53

  100 middleware
  13956.02

  1 async middleware
  14470.50

  5 async middleware
  11909.50

  10 async middleware
  10390.65

  15 async middleware
  8125.27

  20 async middleware
  7202.19

  30 async middleware
  5950.12

  50 async middleware
  4170.79

  100 async middleware
  2383.32
```


damianobarbati:
@fengmk2 if it is optional, why blocking the merge? It won't impact those who do not opt-in. Right?

miwnwski:
Performance aside, is it not better to have it as a separate package? Much like composing is, it's out of domain for Koas IMO.

jmealo:
It's non-obvious how to use async local storage in koa without breaking error handling. I'm going to give this a try. If users have to monkeypatch createContext to get this to work then I think this should be in core.

jmealo:
> Need wait for asyncLocalStorage improve it's performance.

@fengmk2: The overhead is 8% for Node 14. I haven't tested this particular PR but async local storage is not slow. Waiting for it to change to support it in Koa doesn't seem warranted.

jmealo:
There's a couple of incorrect ways of doing this. I don't think it's a Koa issue though, the Node.js documentation (and my IDE's annotations) had `run` as a void. It apparently returns a promise and can/should be `await`ed. The documentation says that it returns the value of `run`.... I might need to read the Node.js source to verify the behavior.

This doesn't break my current app's error handling and allows me to access my Koa `ctx` throughout my codebase:
```javascript
  // Provides async local storage context throughout to koa allowing async operations executed
  // during a request to later retrieve the koa ctx without requiring a reference to be passed
  app.use(async function ContextMiddleware (ctx, next) {
    const store = new Map()
    store.set('ctx', ctx)
    await asyncLocalStorage.run(store, async () => {
      try {
        await next()
      } catch (error) {
        ctx.throw(error)
      }
    })
  })
```
If we don't want to merge async-local-storage into core, I think we should at least document some approaches, their pros/cons and/or a preferred method if there is one.

The issue I'm seeing is if you want to use open telemetry and structured logging, given the performance overhead, it would be ideal if there were a way to combine those responsibilities into an add-on that only mounts a single middleware. Perhaps if koa were to look into what the open telemetry instrumentation does, provide the requisite hooks to avoid shimming/wrapping.

Once you add tracing, the `koa-instrumentation` defaults to creating a span per middleware so the incentive to reduce the number of anonymous middleware becomes apparent.

From a performance perspective, if there were a way to accomplish this from a synchronous middleware that'd be ideal but I don't think that's possible.
@fengmk2: Any thoughts?

miwnwski:
@dead-horse mind rerunning this PR with 16.2 release that has updated async hooks that uses new v8::Context PromiseHook API?

miwnwski:
I'm sort of curious about the result. It looks like the performance gained using new PromiseHook API significantly reduces overhead, to the point that this could be finalised and merged IMO. Though others might disagree.

Running node 16.2 both master and async-hooks produced:

```sh
# master branch
$ npm run bench

> koa@2.13.1 bench
> make -C benchmarks


  1 middleware
  27583.00

  5 middleware
  27694.75

  10 middleware
  27476.26

  15 middleware
  27925.56

  20 middleware
  27453.76

  30 middleware
  26357.53

  50 middleware
  26231.70

  100 middleware
  25393.45

  1 async middleware
  28234.63

  5 async middleware
  27832.86

  10 async middleware
  27335.61

  15 async middleware
  27702.56

  20 async middleware
  27448.03

  30 async middleware
  26193.47

  50 async middleware
  25181.45

  100 async middleware
  22653.18
```

```sh
# async-hooks branch
 $ npm run bench

> koa@2.11.0 bench
> make -C benchmarks


  1 middleware
  28156.92

  5 middleware
  27567.77

  10 middleware
  26594.23

  15 middleware
  27381.46

  20 middleware
  27166.58

  30 middleware
  26120.80

  50 middleware
  25935.13

  100 middleware
  24950.71

  1 async middleware
  26585.90

  5 async middleware
  26596.86

  10 async middleware
  26399.98

  15 async middleware
  25465.95

  20 async middleware
  26197.93

  30 async middleware
  25552.58

  50 async middleware
  23979.93

  100 async middleware
  21888.75
```

dead-horse:
> This feels like I'm missing something though. @dead-horse can you confirm on your end?
Yep, you should change the benchmark test file to `new Koa({ asyncLocalStorage: true })` to enable this feature.

and I've tested in both node@16.1 and node@16.2, 16.2 performance is much better, I think we can consider supporting this feature in koa soon.

```
#node@16.2.0 without asyncLocalStorage
  1 async middleware
  21026.42

  5 async middleware
  20347.36

  10 async middleware
  20525.89

  15 async middleware
  20328.99

  20 async middleware
  20341.83

  30 async middleware
  19403.87

  50 async middleware
  18543.07

  100 async middleware
  16682.64

# node@16.1.0 with asyncLocalStorage
  1 async middleware
  13851.01

  5 async middleware
  13827.81

  10 async middleware
  11712.27

  15 async middleware
  10615.83

  20 async middleware
  9528.57

  30 async middleware
  7698.06

  50 async middleware
  5840.24

  100 async middleware
  3443.92

# node@16.2 with asyncLocalStorage
  1 async middleware
  18193.72

  5 async middleware
  17330.99

  10 async middleware
  16814.28

  15 async middleware
  16032.19

  20 async middleware
  15429.25

  30 async middleware
  13937.79

  50 async middleware
  12049.11

  100 async middleware
  9028.56
```

miwnwski:
> Yep, you should change the benchmark test file to new Koa({ asyncLocalStorage: true }) to enable this feature.

Yep, I misspelled it :)

I think naming could get a bit confusing. To follow Koa convention of delegating from Context, instead of `currentContex` how about `getStore`?

robahtou:
Any update on this?

miwnwski:
I'm inclined to merge this but I want an additional approve because this feels like kind of a feature creep (depending on perspective). I wonder if there's a different implementation approach to consider? Also, I wonder about the name `currentContext` vs `getContext` as previously mentioned.

Also, I think travis changes belongs in a different PR and we need conflicts resolved (and I would like to see a History addition here as well).

fengmk2:
I will rebase and add node 16, 18 test before merge and release minor version.

fengmk2:
https://nodejs.org/dist/latest-v18.x/docs/api/async_context.html#asynclocalstoragerunstore-callback-args

use asyncLocalStorage.run instead

fengmk2:
@jonathanong It's time to release koa v3.
Should we only support node >= 14?

fengmk2 (approved):

lib/application.js
```diff
@@ -139,12 +145,20 @@ module.exports = class Application extends Emitter {
...
       const ctx = this.createContext(req, res)
+      if (this.ctxStorage && this.ctxStorage.enterWith) this.ctxStorage.enterWith(ctx)
       return this.handleRequest(ctx, fn)
     }
 
     return handleRequest
   }
 
+  /**
+   * return currnect contenxt from async local storage
```
jmealo, line 156: Nit: Typo

jmealo:
Great work everyone getting this merged. Many thanks! 🥳

ZanMinKian:
I don't think this PR is a good feature. This PR finally will be refactored to use a middleware. See https://github.com/koajs/koa/pull/1718. But using a __built-in__ middleware seems not good. It breaks the principle that koa core does NOT bundle any middleware.
image
Actually, we can create a `@koa/async_local_storage` middleware to achieve our goal. There is no enough reason to bundle a built-in middleware into koa core.

siakc:
Is there an official documentation for this feature?

Issue #1451, opened 2020-03-24
Koajs converts thrown errors to status code 500


jhsware:
If you use the AWS S3 Node.js client to create a proxy, it throws an error with status code 304 when an object hasn't been modified. This propagates to the Koa.js error handling code. Because Node.js sets the http status code of the error on the prop `err.statusCode`

https://github.com/nodejs/node/blob/066bdec64352278ed324a1bb680f5422e00a3aa5/lib/_http_outgoing.js#L410

and Koa.js expects `err.status`, Koa.js will return a 500 instead of the actual status code set on the error object by Node.js.

https://github.com/koajs/koa/blob/eda27608f7d39ede86d7b402aae64b1867ce31c6/lib/context.js#L150

I suggest adding the line

```js
err.status = err.status || err.statusCode;
```

or changing to

```js
    // ENOENT support
    let statusCode = err.status || err.statusCode;

    if ('ENOENT' == err.code) statusCode = 404;

    // default to 500
    if ('number' != typeof statusCode || !statuses[statusCode]) statusCode = 500;

    // respond
    const code = statuses[statusCode];
    const msg = err.expose ? err.message : code;
    this.status = err.status = statusCode;
```
so the developer doesn't have to set it manually, which is non-trivial to figure out and adds unnecessary code complexity in situations where node already has set a status code.

vijaykrishnavanshi:
We can close this issue now. Fixed in #1460 PR.

jhsware:
@vijaykrishnavanshi please add a test so this doesn't break in the future. I would have contributed with a PR but I didn't have time to set up the testing environment and didn't feel that only contributing the code was a complete solution.

vijaykrishnavanshi:
Already added those in the PR 👍 
If you have any other specific case that you want the test for let me know.

jhsware:
Thanks @vijaykrishnavanshi now I found it!
https://github.com/koajs/koa/pull/1460/commits/ce9133a34027eee06667cde8afea2273d41a4a3f

PR #1460, opened 2020-05-15
fix: Koa.js handling of node.js set error codes

vijaykrishnavanshi:
Fixes issue:  #1451

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -143,16 +143,18 @@ const proto = module.exports = {
     // force text/plain
     this.type = 'text';
 
+    let statusCode = err.status || err.statusCode;
+
     // ENOENT support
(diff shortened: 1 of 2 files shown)


lib/context.js
```diff
@@ -143,16 +143,18 @@ const proto = module.exports = {
     // force text/plain
     this.type = 'text';
 
+    let statusCode = err.status || err.statusCode;
```
dead-horse, line 146: can we add a test case for this change?

vijaykrishnavanshi: Yeah sure

vijaykrishnavanshi: Added the test for statusCode change in the PR.

PR #1465, opened 2020-05-27
fix(context): use strict equality

JsonMa:
Using strict equality when comparing error to null. When passing an `undefined` variable  to `onerror` function, you will not get `non-error` result. `exp: null == undefined  -> true`

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -108,7 +108,7 @@ const proto = module.exports = {
     // don't do anything if there is no error.
     // this allows you to pass `this.onerror`
     // to node-style callbacks.
-    if (null == err) return;
+    if (err === null) return;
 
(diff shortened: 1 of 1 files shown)


lib/context.js
```diff
@@ -108,7 +108,7 @@ const proto = module.exports = {
     // don't do anything if there is no error.
     // this allows you to pass `this.onerror`
     // to node-style callbacks.
-    if (null == err) return;
```
dead-horse, line 111: `exp: null == undefined -> true`. This is what we expect here. :(

JsonMa: How  to capture the `non-error` error when passing `undefined`, cause you have already  returned.

Issue #1469, opened 2020-06-06
Feature Request: ESM module support


cookiengineer:
When using koa.js for a new project, I was wondering why there's no ESM module support?

I mean, conceptually speaking, koa.js embraces async/await syntax everywhere - why does it have no ESM `export` and `import` support then?

PR #1474, opened 2020-06-21
feat: support esm

ZYSzys:
Support ESM module.

Fixes: https://github.com/koajs/koa/issues/1469

Refs: https://github.com/addaleax/gen-esm-wrapper

### DIFF
--- a/test/load-with-esm.js
+++ b/test/load-with-esm.js
@@ -1,0 +1,32 @@
+const assert = require('assert');
+
+// ESM support is flagged on v12.x.
+const majorVersion = +process.version.split('.')[0].slice(1);
+if (majorVersion < 12) {
+  process.exit();
+}
+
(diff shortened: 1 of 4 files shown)


test/load-with-esm.js
```diff
@@ -0,0 +1,32 @@
+const assert = require('assert');
+
+// ESM support is flagged on v12.x.
+const majorVersion = +process.version.split('.')[0].slice(1);
+if (majorVersion < 12) {
+  process.exit();
```
dead-horse, line 6: this will exit the process and other test case can't be executed.

ZYSzys: Fixed, check `Node.js` version to determine whether skip in `before` hook.

panva:
Related - #1476

PR #1480, opened 2020-07-14
deps: update statuses to v2.0.1

3imed-jaberi:


dead-horse:
https://github.com/jshttp/statuses/blob/master/HISTORY.md#200--2020-04-19

looks like this upgrade contains incompatible changes

3imed-jaberi:
Kindly, can you tell me which change is incompatible and I will see what I can do.

test/application/respond.js
```diff
@@ -374,7 +374,7 @@ describe('app.respond', () => {
     describe('with custom status=700', () => {
       it('should respond with the associated status message', async() => {
         const app = new Koa();
-        statuses['700'] = 'custom status';
```
dead-horse, line 377: this is a breaking change for users who want to customize status code

dead-horse: IMO we need to provide a more controllable way to customize status code.

3imed-jaberi: > this is a breaking change for users who want to customize status code

According to what I understand form you, we should leave this PR to Koa@3  !!

3imed-jaberi: > IMO we need to provide a more controllable way to customize status code.

Perhaps we can provide a method to store, set, cache.. custom messages for custom codes.

I think this approach is simple and clean.

3imed-jaberi (commented):
Updated to latest version v2.0.1 🚀!

3imed-jaberi (commented):
@jonathanong @dougwilson @dead-horse @fengmk2 
Hello guys, take a look if this is considered as a breaking change you can create a PR for koa v3.x where I can compare it or make a label for v3.x ...!

fengmk2 (approved):

jonathanong:
merged here: https://github.com/koajs/koa/commit/0d9f062aa840c54b5c20911dab6ce62128d3b089 (sorry, you're not the author of this commit, not sure how to do that. I had to rebase)

PR #1481, opened 2020-07-14
deps: update type-is to v1.6.18

3imed-jaberi:
- https://github.com/jshttp/type-is/blob/master/HISTORY.md#1618--2019-04-26
- https://github.com/jshttp/type-is/blob/master/HISTORY.md#1617--2019-04-25


3imed-jaberi:
[Ref #comment](https://github.com/koajs/koa/pull/1486#issuecomment-675797312)

PR #1482, opened 2020-07-14
deps: update depd to v2.0.0

3imed-jaberi:


julienw:
I may be wrong, but I think we still depend on depd 1.1 because we depend on http-errors which depends on depd 1.1.

PR #1485, opened 2020-07-14
deps: update http-assert to v1.4.1

3imed-jaberi:
- https://github.com/jshttp/http-assert/blob/master/HISTORY.md#140--2018-09-09
- https://github.com/jshttp/http-assert/blob/master/HISTORY.md#141--2019-04-28


3imed-jaberi:
[Ref #comment](https://github.com/koajs/koa/pull/1486#issuecomment-675797312)

PR #1490, opened 2020-07-16
chore: fix grammatical and spelling errors in comments and tests

kubejm:
I noticed a handful of small things with regards to spelling and grammar as I've been digging through this code as of recent.  Even though small, I hope this helps.  If these changes seem reasonable, then let me know if you'd like me to take any further action or make adjustments.

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -146,7 +146,7 @@ module.exports = {
   },
 
   /**
-   * Set pathname, retaining the query-string when present.
+   * Set pathname, retaining the query string when present.
    *
    * @param {String} path
    * @api public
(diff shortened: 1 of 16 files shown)


lib/request.js
```diff
@@ -422,7 +422,7 @@ module.exports = {
    * When `app.proxy` is `true`, parse
    * the "X-Forwarded-For" ip address list.
    *
-   * For example if the value were "client, proxy1, proxy2"
+   * For example if the values were "client, proxy1, proxy2"
```
miwnwski, line 425: I disagree, a string is one value.

kubejm: Makes sense, I adjusted it from "the values were" to the "the value was", squashed the commits, and pushed.

Issue #1464, opened 2020-05-19
Response code auto set to 204 after being previously set to 200 with undefined or null body.


austin-rausch:
Koa version 2.12.0

Running this snippet and sending a request will return a 204 response code despite the status code being explicitly set to 200, it does return a 200 if you reverse the two lines:
```
const koa = require('koa');
const app = new koa();
app.use(ctx => {
  ctx.status = 200;
  ctx.body = undefined;
});
app.listen(3000);
```

This contradicts the [documentation](https://github.com/koajs/koa/blob/master/docs/api/response.md#responsebody-1) and I believe it is caused by not checking for `_explicitStatus` on [this](https://github.com/koajs/koa/blob/faeaff5c149a81a188ab8e5af0b994029e45acbb/lib/response.js#L140) line.

I'm happy to fix the issue by adding the check there, I just want to confirm this is un-intended behavior or if its a mistake in the documentation.  As far as I can tell the HTTP spec doesn't specifically mandate that a body be included on a 200 response, only that a 204 MUST NOT include a message-body.

ejose19:
Yes, there are some cases one wants to reply with "no body" for other status codes as well, that's why I submitted https://github.com/koajs/koa/pull/1447 which is already on latest version (2.12.0)

To obtain the result you want use as this

```ts
// This must be always set first before status, since null | undefined 
// body automatically sets the status to 204
  ctx.body = null;
// Now we override the 204 status with the desired one
  ctx.status = 200; 
```

notice the server won't respond with "null" value, it will be no response like it does for 204

austin-rausch:
I can't say I agree with the behavior, if a response code is set explicitly, then a body is set after it shouldn't change the response code.  Thats a strange side effect that I don't think many people will expect.

If I set a status code, I would expect that status code to be what is returned assuming it is valid for the body given.

If the response code and body are invalid together I would expect an error to be thrown.

At the very least this side effect needs documented correctly.  Again, offering to do it just looking for some confirmation or consensus.

miwnwski:
An update for response.body= seems sane 👍

PR #1493, opened 2020-07-26
docs: update specific for auto response status

AlbertAZ1992:
fix: https://github.com/koajs/koa/issues/1464

### DIFF
--- a/docs/api/response.md
+++ b/docs/api/response.md
@@ -130,9 +130,18 @@ ctx.response.status = 204;
   - `Buffer` written
   - `Stream` piped
   - `Object` || `Array` json-stringified
-  - `null` no content response
+  - `null` || `undefined` no content response
 
-If `response.status` has not been set, Koa will automatically set the status to `200` or `204`.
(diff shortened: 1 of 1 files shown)


docs/api/response.md
```diff
@@ -130,9 +130,18 @@ ctx.response.status = 204;
   - `Buffer` written
   - `Stream` piped
   - `Object` || `Array` json-stringified
-  - `null` no content response
+  - `null` || `undfined` no content response
+
+If `response.status` has not been set, Koa will automatically set the status to `200` or `204` depending on `response.body`. Specifically, if `response.body` has not been set or has been set as `null` || `undfined`, which means no content response, Koa will automatically set `response.status` to `204`. If you really want to send no content response with other status, you should override the `204` status as the following way:
```
miwnwski, line 135: `undfined` misspelled.

miwnwski: I would remove `which means no content response`, the reason to why null or undefined it set is ambiguous to Koa - but it interprets it as 204 No Content as you wrote.

miwnwski: NIT: I would not use `||`'s for documentation text, rather use explicit `or`

AlbertAZ1992: Thanks. That was updated. I also thought about `or` against `||` as definition listed above. Now i change to use `or` for documentation text.

PR #1495, opened 2020-08-09
Export respond() in application.js

LabMemNo003:
I want to use koa as a middleware into koa-compose, so I need to implement my own Application.callback() and Application.handleRequest() functions, which uses the 'function respond(ctx)' in application.js.
By now, I just copied respond()'s source code into my file, but it is helpful that this function can be exported by koa.

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -275,6 +275,12 @@ function respond(ctx) {
 }
 
 /**
+ * Export respond() for the users who want to call it elsewhere
+ */
+
+module.exports.respond = respond;
+
+/**
  * Make HttpError available to consumers of the library so that consumers don't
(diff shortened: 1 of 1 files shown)


miwnwski:
I'm not opposed to exporting respond, but I don't imagine this was ever intended to be public API.

However, if we do decide to export it I wonder if it's better to just stick it in prototype instead and document source as `@private`.

LabMemNo003:
> I'm not opposed to exporting respond, but I don't imagine this was ever intended to be public API.
> 
> However, if we do decide to export it I wonder if it's better to just stick it in prototype instead and document source as `@private`.

Thanks for your imformation!

Issue #1525, opened 2021-02-18
Explicit-Null-Body should contains Content-Length


nfer:
Koa version 2.13.1

Running below codes:
```js
const koa = require('koa');
const app = new koa();

app.use(ctx => {
  ctx.body = null;
  ctx.status = 404;
});

app.listen(3000, () => {} )
```

Both in chrome and curl test, it fails in 5s.

```shell
$> time curl http://localhost:3000/
curl http://localhost:3000/  0.01s user 0.01s system 0% cpu 5.013 total
```

I think the question is lost of Content-Length header:
> A user agent SHOULD send a Content-Length in a request message when
   no Transfer-Encoding is sent and the request method defines a meaning
   for an enclosed payload body. 

from [RFC7230](https://tools.ietf.org/html/rfc7230#section-3.3.2)

I add one line codes `ctx.length = 0;` in [application module](https://github.com/koajs/koa/blob/master/lib/application.js#L258), it pass all test cases and works for chrome and curl.

If it is really a bug, and the modification works, I will submit a PR.

ognjenjevremovic:
Hey @nfer 👋 .

Are you working on this? Do you perhaps need some assistance?

If you're not actively working on this (and the bug is still present), I'd be willing to provide PR myself is that is ok with you?
Let me know if I can help somehow with this issue.

Cheers!

nfer:
Thanks @ognjenjevremovic , I think PR is welcome.

ognjenjevremovic:
Any updates on the PR?

miwnwski:
This was merged by Jonathan into master so I'm closing this.

3N26:
this fix was not merged into master 
https://github.com/ognjenjevremovic/koa/blob/master/lib/application.js#L256

PR #1528, opened 2021-03-18
[Fix] Content-Length header on explicit null body response

ognjenjevremovic:
Fixes *#1525*

### Changes
* Set the context length value to 0 for *explicit null body* response(s),
* Provided the necessary set of tests for this change,
* Maintained the test coverage of 100%.

### Checklist
After the changes I ensured that:
 * no stylistic or unwanted errors are present, by running `npm run lint` from the root directory,
 * no tests are failing, by running `npm run test` from the root directory,
 * build passes with no issues, by running `npm run build` from the root directory.

### Done
- [x] Added/updated unit tests for this change
- [x] Included links to related issues/PRs

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -256,6 +256,7 @@ function respond(ctx) {
     if (ctx.response._explicitNullBody) {
       ctx.response.remove('Content-Type');
       ctx.response.remove('Transfer-Encoding');
+      ctx.length = 0;
       return res.end();
     }
     if (ctx.req.httpVersionMajor >= 2) {
--- a/test/application/respond.js
+++ b/test/application/respond.js
@@ -866,7 +866,46 @@ describe('app.respond', () => {
         .expect('')
         .expect({});
 
+      assert.equal(res.headers.hasOwnProperty('transfer-encoding'), false);
       assert.equal(res.headers.hasOwnProperty('content-type'), false);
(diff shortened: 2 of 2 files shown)


3imed-jaberi (approved):
lgtm +1

jkomyno (approved):
This commit seems legitimate.
But why is it necessary to test the absence of the 'transfer-encoding' header?

ognjenjevremovic:
@jkomyno I apologize for getting back late. I've been offline for a week, due to the health issues.

I've referenced the [RFC7230](https://tools.ietf.org/html/rfc7230#section-3.3.2).
It seems that the use case applies to _user agents_ instead (I guess I skimmed way too fast through the text and misunderstood the writing). It does seems redundant, especially in test cases but I guess I wanted to follow the recommended way of handling the response headers (however, now I see I got them confused for the request headers instead).
 
> A user agent SHOULD send a Content-Length in a request message when
no Transfer-Encoding is sent and the request method defines a meaning
for an enclosed payload body.

Do you think the assertion should be removed?
I can provide a commit for getting rid of the unnecessary header check.

ognjenjevremovic:
Any updates on the MR?

jonathanong (approved):

jonathanong:
@ognjenjevremovic I will merge it if I have time to do a release. hopefully someone else will get to it before I do

PR #1529, opened 2021-04-12
Feat: remove unnecessary code

ryouaki:
remove _explicitNullBody from response.js. 

We can do it at application.js。


ryouaki:
```js
    // response.js

    if (null == val) {
      if (!statuses.empty[this.status]) this.status = 204;
      // if (val === null) this._explicitNullBody = true; no need, we can do it at application.js
      this.remove('Content-Type');
      this.remove('Content-Length');
      this.remove('Transfer-Encoding');
      return;
    }
```

```js
    // application.js

    // ==== Start ====
    // The source code at response.js's line 146 to 148 has removed Content-Type & Transfer-Encoding when set null to body.
    // if (ctx.response._explicitNullBody) {
    //   ctx.response.remove('Content-Type');
    //   ctx.response.remove('Transfer-Encoding');
    //   return res.end();
    // }
    // ==== Replace ====
    if (body === null) {
      return res.end();
    }
    // ==== End ====
```

ryouaki:
@dead-horse 
@niftylettuce 

Please have a look at this PR.

ryouaki:
@miwnwski

lib/response.js
```diff
@@ -138,7 +138,7 @@ module.exports = {
     // no content
     if (null == val) {
       if (!statuses.empty[this.status]) this.status = 204;
-      if (val === null) this._explicitNullBody = true;
+      // if (val === null) this._explicitNullBody = true; no need, we can do it at application.js
```
miwnwski, line 141: You are removing this assignment here with the comment it can be done in Application. but I don't see you setting this in Application?

lib/application.js
```diff
@@ -253,9 +253,13 @@ function respond(ctx) {
...
-    if (ctx.response._explicitNullBody) {
-      ctx.response.remove('Content-Type');
-      ctx.response.remove('Transfer-Encoding');
+    // The source code at response.js's line 146 to 148 has removed Content-Type & Transfer-Encoding when set null to body.
+    // if (ctx.response._explicitNullBody) {
+    //   ctx.response.remove('Content-Type');
+    //   ctx.response.remove('Transfer-Encoding');
+    //   return res.end();
+    // }
+    if (body === null) {
```
miwnwski, line 262: I don't understand this comment at all. You're referencing line 146-148 and those are empty lines in response.js file - or specifically a closing bracket and a comment.

miwnwski (changes requested):
Why are we removing this? I don't understand it.
This was specifically added to allow body-less responses by #1447 - and I'm not exactly sure why you want to remove this? More over, if you're removing code why do you just comment it out in this PR?

I guess what I'm saying is that I don't understand this PR in its current form.

lib/response.js
ryouaki: "_explicitNullBody" is used for check if body  is null and used by application.js?
but at line 255 of application.js the code has do the check before. so "_explicitNullBody" is unnecessary.

lib/application.js
ryouaki: when body is null and "Content-Type" & 'Transfer-Encoding' had be removed from response.js before. and then at next code .still res.end() when body is null. So this code is unnecessary

Issue #1534, opened 2021-05-13
Upgrade outdated koa-convert dependency to ^2.0.0


atombrenner:
`koa` depends on `koa-convert@1.2.0` which depends on `koa-compose@3.2.1` which depends on `any-promise`.

`any-promise` is very old (last release 5 years ago) and uses dynamic requires, which break bundlers like esbuild

```
node_modules/any-promise/register.js:24:14: warning: This call to "require" will not be bundled because the argument is not a string literal (surround with a try/catch to silence this warning)
```
`koa-convert@2.0.0` depends on `koa-compose@4.1.0` which has the dependency to `any-promise` removed. Because `koa` has a already a direct dependency to `koa-compose@4.1.0` (see [here](https://github.com/koajs/koa/blob/master/package.json#L49)), I hope it is safe to upgrade the dependency to `koa-convert`

atombrenner:
see https://github.com/koajs/koa/pull/1535

PR #1535, opened 2021-05-13
chore: upgrade koa-convert dependency to ^2.0.0

atombrenner:
fix issue #1534

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "http-errors": "^1.6.3",
     "is-generator-function": "^1.0.7",
     "koa-compose": "^4.1.0",
-    "koa-convert": "^1.2.0",
+    "koa-convert": "^2.0.0",
     "on-finished": "^2.3.0",
     "only": "~0.0.2",
     "parseurl": "^1.3.2",


ryouaki (approved):

miwnwski:
I agree it should be upgraded (and removed in next) but this would result in a breaking change (I don't really remember why convert got bumped major, maybe you can enlighten me if you recently checked it out?).

This will have to wait to next version of Koa, though any possible next Koa dependency would actually remove this dependency.

atombrenner:
>  (I don't really remember why convert got bumped major, maybe you can enlighten me if you recently checked it out?).

As far as I can see, the change is the dependency upgrade of koa-compose@3.2.1 to koa-compose@4.1.0 plus some more cosmetic refactorings.

miwnwski:
Well in that case, it LGTM!

miwnwski (approved):

miwnwski:
Holding off merge a little

3imed-jaberi (approved):
lgtm +1 

> whisper: we must move quickly to work on next release of Koa.

PR #1545, opened 2021-06-12
fix: Fix deprecated folder mapping "./" in the "exports" field module

tkvern:
Run in Node.js 16.x ，I got warning message. then fixed for Node.js 16.x

image

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -504,7 +504,7 @@ module.exports = {
    * @api private
    */
   set accept(obj) {
-    return this._accept = obj;
+    this._accept = obj;
   },
 
   /**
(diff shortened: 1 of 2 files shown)


mnahkies:
Any update on this? Just upgraded to nodejs 16 on a service and started seeing this too.

tkvern:
> Any update on this? Just upgraded to nodejs 16 on a service and started seeing this too.

Author haven't see this problem

miwnwski:
If you update this (and thus trigger a new CI pass) I can take a look and merge this. I just want to make sure we're not breaking old behaviour with this change.

Issue #1547, opened 2021-06-17
bug report! If koa and http2server are used together, the response with method "head" request will never end!


masx200:
bug report! If koa and http2server are used together, the response with method "head" request will never end!
```js
import http2 from "http2";
import Koa from "koa";
import fs from "fs";
import { fileURLToPath } from "url";
import { join, dirname } from "path";
const __filename = fileURLToPath(import.meta.url);
const __dirname = dirname(__filename);
const app = new Koa();
const selfSignedKey = join(__dirname, "./certs/self-signed.key.pem");
const selfSignedCert = join(__dirname, "./certs/self-signed.cert.pem");
app.use((ctx) => {
    ctx.body = "Hello Koa";
});
const config = {
    sslKey: fs.readFileSync(selfSignedKey).toString(),
    sslCert: fs.readFileSync(selfSignedCert).toString(),
};
const serverHandler = app.callback();
const httpsopt = { cert: config.sslCert, key: config.sslKey };
const server = http2.createSecureServer(httpsopt, serverHandler);

server.on("listening", () => {
    console.log(`Server listening on ` + JSON.stringify(server.address()));
});
server.listen(3000);
```

```json
{
    "type": "module",
    "dependencies": {
        "koa": "^2.13.1"
    }
}
```
```shell
node index.js
```
Initiate a "head" request in the browser.
```
Request URL: https://localhost:3000/
Referrer Policy: strict-origin-when-cross-origin
```
```
:authority: localhost:3000
:method: HEAD
:path: /
:scheme: https
accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.9
accept-encoding: gzip, deflate, br
accept-language: zh-CN,zh;q=0.9
referer: https://localhost:3000/
sec-ch-ua: " Not;A Brand";v="99", "Google Chrome";v="91", "Chromium";v="91"
sec-ch-ua-mobile: ?0
sec-fetch-dest: empty
sec-fetch-mode: cors
sec-fetch-site: same-origin
upgrade-insecure-requests: 1
user-agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.106 Safari/537.36
```
the response with method "head" request will never end!

masx200:
My solution is modified as follows and everything works fine!
```js

import http2 from "http2";
[...]

PR #1548, opened 2021-06-24
fix bug https://github.com/koajs/koa/issues/1547

masx200:
fix bug https://github.com/koajs/koa/issues/1547

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -229,7 +229,9 @@ module.exports = class Application extends Emitter {
 function respond(ctx) {
   // allow bypassing koa
   if (false === ctx.respond) return;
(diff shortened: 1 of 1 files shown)


3imed-jaberi (approved):
LGTM 💯

jonathanong:
will need a test for this

masx200:
Thanks, I don't know how to write unit tests, and I'm just an amateur programmer.

jonathanong:
actually, something we need to do here is consume any bodies, otherwise if the body was a stream, this would create a leak. this would need some more complicated checks for this code as well as existing code

PR #1559, opened 2021-07-18
docs: document request.query automatic decodeURIComponent

madarche:
The elements to justify this addition are the following:
* https://github.com/koajs/koa/blob/master/lib/request.js#L175
* https://nodejs.org/api/querystring.html#querystring_querystring_parse_str_sep_eq_options
* https://nodejs.org/api/querystring.html#querystring_querystring_unescape_str

Thanks 🙇‍‍♂️


jonathanong:
the correct documentation here is that we rely on `querystring` and point the docs to it

PR #1562, opened 2021-08-10
fix: Do not response Content-Length if Transfer-Encoding is defined

charlyzeng:
## Background
Nodejs http module can not handle response that contains both Content-Length and Transfer-Encoding header, you can see it at https://github.com/nodejs/node/blob/v16.6.1/deps/llhttp/src/llhttp.c#L5811.

## Problem
Koa will auto set Content-Length response header when use `ctx.body = ...`, but if Transfer-Encoding response header is defined, Content-Length should not be set. Otherwise, the response can not be handled by nodejs http module.

## How To Appear
```javascript
// server.js
const Koa = require('koa');

const app = new Koa();
app.use((ctx) => {
  ctx.set('Transfer-Encoding', 'chunked');
  ctx.body = 'hello world';
});
app.listen(9000);

// client.js
const http = require('http');

http.get('http://127.0.0.1:9000');
```
As above, run `node server.js` to start server, then run `node client.js` to send a request. Finally, you'll get an error as follow:
image

## How To Resolve
Don't set Content-Length header if Transfer-Encoding is defined.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,11 +1,14 @@
-小菜 <xtx1130@gmail.com>
 Aaron Heckmann <aaron.heckmann+github@gmail.com>
 Adam L <skyros@gmail.com>
 Adam Lau <skyros@gmail.com>
 Aesop Wolf <aesopwolf@users.noreply.github.com>
+AlbertAZ1992 <ziyuximing@163.com>
+Alex Berk <berkalexanderc@gmail.com>
 AlexeyKhristov <AlexeyKhristov@users.noreply.github.com>
 Alexsey <agat00@gmail.com>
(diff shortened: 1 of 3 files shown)


3imed-jaberi (approved):
+1

xcatliu:
Hi @dead-horse , Why the latest version 2.13.3 does not contain these code?

siakc:
> Hi @dead-horse , Why the latest version 2.13.3 does not contain these code?

[koa](https://github.com/koajs/koa/tree/master)/[lib](https://github.com/koajs/koa/tree/master/lib)
/response.js:199

PR #1574, opened 2021-08-23
Feat: upgrade querystring to URLSearchParams

### LINKED ISSUE #1570: Platform: upgrade query string to URLSearchParams
https://nodejs.org/api/querystring.html

adaex:
close #1570 

I'm not sure if this is necessary.

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -11,7 +11,7 @@ const accepts = require('accepts')
 const contentType = require('content-type')
 const stringify = require('url').format
 const parse = require('parseurl')
-const qs = require('querystring')
+const URLSearchParams = require('url').URLSearchParams
 const typeis = require('type-is')
 const fresh = require('fresh')
 const only = require('only')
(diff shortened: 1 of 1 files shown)


jonathanong:
thanks! looking more into this, do you think it would be better if we just exposed the `URLSearchParams` instance as `ctx.request.searchParams` or something?

yeah, not necessary, but query string is considered legacy

adaex:
@jonathanong 

I think exposed the `URLSearchParams` instance as `ctx.request.searchParams` is a good idea. 

How do you see #1575 ? If there is no problem, I will continue to complete it.

miwnwski (changes requested):
LGTM, but requires some additions.

If you don't mind exposing searchParams as Request#searchParams, and it would be nice to add a line to History.md as well as this is a breaking change. Also a test or two for completeness!

Since this is available from node 10+ it's mergeable once the PR is massaged a bit.

I closed #1575 in favour of this PR.

jonathanong:
merged https://github.com/koajs/koa/pull/1828

PR #1593, opened 2021-09-28
fix missing cleanup, if response socket is no longer writeable (issue 1547)

Azlond:
Relates to issue #1547, see my [comment there](https://github.com/koajs/koa/issues/1547#issuecomment-908370552)

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,4 +1,10 @@
 
+3.0.0-alpha.3 / 2025-02-11
+==================
+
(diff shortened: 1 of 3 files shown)


miwnwski (commented):
LGTM, though would you mind adding a test for this?

Azlond:
I don't really know how to write tests for koa, so not sure if I can be of help with that.

PR #1608, opened 2021-11-09
test: fix ESM tests

### LINKED ISSUE #1571: Fix ESM tests
when upgrade to jest and testing on node v16, they break. gonna need some help fixing them

vbetsun:
Regarding to the [jest documentation](https://jestjs.io/docs/ecmascript-modules) we should use `--experimental-vm-modules` command-line option for enabling ES modules.
Closes #1571

### DIFF
--- a/__tests__/load-with-esm.js
+++ b/__tests__/load-with-esm.js
@@ -2,7 +2,7 @@ const assert = require('assert')
 
 let importESM = () => {}
 
-describe.skip('Load with esm', () => {
+describe('Load with esm', () => {
   beforeAll(function (){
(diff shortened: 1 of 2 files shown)


3imed-jaberi (approved):
+1

miwnwski (approved):
I would prefer a solution without NODE_OPTIONS, but maybe this just is the lesser of all evils. If that's the case, I'm fine with this!

miwnwski:
Actually, I'm going to close this without a merge because I don't think this is an issue anymore.
Please, reopen and correct me if I'm wrong here.

PR #1630, opened 2022-01-10
Added exports.default to package.json

daleyjem:
Per NodeJS documentation on the `package.json` `exports` object, there seems to be a need for a fallback `default` key:
https://nodejs.org/api/packages.html#conditional-exports

Additionally, I'm seeing things like eslint fail to resolve until a `default` key has been specified. I've not dug any further into why, but I think _some_ default should be defined until more options should later be added.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,8 @@
   "exports": {
     ".": {
       "require": "./lib/application.js",
-      "import": "./dist/koa.mjs"
+      "import": "./dist/koa.mjs",
+      "default": "./dist/koa.mjs"
     },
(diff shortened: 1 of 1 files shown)


jeremy-daley-kr (approved):

3imed-jaberi (approved):

jeremy-daley-kr:
@3imed-jaberi Who has the ability to merge this?

kevinpeno:
@jonathanong should be gtg now, if its something we want merged.

jonathanong (approved):

PR #1650, opened 2022-03-09
add cache for charset

buzuosheng:


miwnwski:
Is this a performance issue really worth caching? Do you have any benchmarks to provide?

PR #1653, opened 2022-03-25
ci: ci update

xtx1130:
update gitlab ci for koa repo

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -2,18 +2,23 @@ name: Node.js CI
 
 on:
   push:
-    branch: master
+    paths-ignore:
+      - 'docs/**'
+      - '*.md'
   pull_request:
-    branch: master
+    paths-ignore:
+      - 'docs/**'
+      - '*.md'
 
 jobs:
   build:
 
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
 
     strategy:
       matrix:
         node-version: [12.x, 14.x, 16.x]
+        os: [macos-latest, ubuntu-latest, windows-latest]
 
     steps:
     - uses: actions/checkout@v2
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ npm-debug.log
 .idea
 *.iml
 dist
+yarn.lock


xtx1130:
Is anyone here for code review?

.github/workflows/node.js.yml
```diff
@@ -2,18 +2,23 @@ name: Node.js CI
...
 
 jobs:
   build:
 
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
 
     strategy:
       matrix:
         node-version: [12.x, 14.x, 16.x]
```
3imed-jaberi, line 20: I think soon we will drop the 12.x version and we will support 14+ or maybe drop 14.x also and support 16+!!

3imed-jaberi (approved):
+1

.github/workflows/node.js.yml
xtx1130: Yes, I think when 12.x is outdated, we can add 17.x in ci

miwnwski: I don't think Koa 2 will drop any major version support. If we're following semver that would be a major break.

.github/workflows/node.js.yml
```diff
@@ -2,18 +2,23 @@ name: Node.js CI
...
+        os: [macos-latest, ubuntu-latest, windows-latest]
```
miwnwski, line 21: I ask myself, is this really necessary? :) We're not compiling anything, if node works on the platform, Koa will too. Just seems like more work for no real benefit? Do you have a different opinion here or am I missing something maybe?

miwnwski:
I'll close this unless you feel strongly about this or I'm missing something. Don't hesitate to reopen but right now I don't see the need to test Koa on multiple platforms.

.github/workflows/node.js.yml
xtx1130: Node v12.x is END-OF-LIFE now, and node v18.x will be active LTS in 2022/10/18. I think it is necessary to update ci version to consistent with nodejs releases

.github/workflows/node.js.yml
xtx1130: Nodejs has some APIs that return inconsistent results between win and other platforms(such as path). I think add these platforms can prevent some cross-platform bugs from happening.

PR #1654, opened 2022-03-25
Update koa-vs-express.md

jeffwillden:
As it was, this table made it look like Koa doesn't support routing, templating, sending files, jsonp, which is technically true, but there is Koa middleware built for each of those, so it's helpful to clarify that in the table, rather than leaving those cells empty. It paints Koa in a better light for those who are unfamiliar and are comparing their options, which is the target audience of this page.

### DIFF
--- a/docs/koa-vs-express.md
+++ b/docs/koa-vs-express.md
@@ -11,11 +11,11 @@
 
 | Feature           | Koa | Express | Connect |
 |------------------:|-----|---------|---------|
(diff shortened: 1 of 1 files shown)


3imed-jaberi (commented):
`Connect` also support all of these features via middleware...

miwnwski:
I can only guess but I imagine the entire point of that table is exactly that, to put significance on Koa's no bloat approach.
Instead of changing that table, I would have a separate table with some useful @koajs/plugins instead.

PR #1658, opened 2022-04-02
Fix all linter warnings

3imed-jaberi:
Fix all linter warnings.


miwnwski (approved):
If we're sticking to this linter rule, then I'm OK with this!

Issue #1672, opened 2022-06-10
[fix] Setter of `request.query` does not save array when length of the array is 1


3jins:
## Describe the bug

**Node.js version:** 16.10.0

**OS version:** Mac OS Monterey 12.3.1 (It doesn't matter...)

**Description:** 

## Actual behavior

Setter of `request.query` saves single element instead of the array.

[Setter of `request.query` uses `qs.stringify`](https://github.com/koajs/koa/blob/master/lib/request.js#L186) to save the input to `request.querystring`, and it would be converted into the form like `key=value1&key=value2`.
But when array size is 1, it becomes unable to know whether it's array or string since it would be saved as `key=value1`.
So preprocessing would be necessary I think.

## Expected behavior

An array should be saved as I passed.

## Code to reproduce

```typescript
  ctx.query = { tagIdList: ['abcd'] };
  console.log(ctx.query); // It prints out { tagIdList: 'abcd' }
```

miwnwski:
A solution was proposed in PR https://github.com/koajs/koa/pull/1673

However, I'm tentatively holding any approvals as I judge this to require a semver major bump, in which case Koa could benefit from other changes as well.

kravorkid:
@miwnwski Is there a v3 branch somewhere and a roadmap where we can help on some features ?

jonathanong:
koa is just using the official querystring library

PR #1673, opened 2022-06-11
fix: Add brackets to querystring to clarify it's array (#1672)

3jins:
If length of array is 1, it was being recognized as a string, not an array.

- AS-IS:
  { colors: ['black', 'mustard'] } -> colors=black&colors=mustard
  { colors: ['black'] } -> colors=black
- TO-BE:
  { colors: ['black', 'mustard'] } -> colors[]=black&colors[]=mustard
  { colors: ['black'] } -> colors[]=black

Added qs dependency for this fix.

## Checklist

- [ ] I have written a commit message that passes commitlint linting.


miwnwski (commented):
I'm for this change, however I fear this can break a lot of applications because users might rely on current behaviour today. I wouldn't be comfortable releasing this under Koa 2.x

miwnwski (approved):
LGTM

Seems like a decent change but I'm going to let other maintainers decide on this one.

3jins:
Is anyone willing to merge this PR?
I don't have write access so that I cannot merge this.

3imed-jaberi (commented):
@miwnwski From my side, in theory yes we can merge this PR now since we are on[ v3-alpha](https://github.com/koajs/koa/blob/master/package.json#L3). 

But we need to make it flexible because we have many strategies here and we don't push one over the other, let's prevent breaking changes

```js
qs.stringify({ a: ['b', 'c'] }, { arrayFormat: 'indices' })
// 'a[0]=b&a[1]=c'
qs.stringify({ a: ['b', 'c'] }, { arrayFormat: 'brackets' })
// 'a[]=b&a[]=c'
qs.stringify({ a: ['b', 'c'] }, { arrayFormat: 'repeat' })
// 'a=b&a=c'
qs.stringify({ a: ['b', 'c'] }, { arrayFormat: 'comma' })
// 'a=b,c'
```

cc: @fengmk2 🤔?!

andrew0:
Please can we not introduce `qs` to Koa... it's one of the things I wanted to escape from Express. The behavior is very unusual compared to my experience with other web frameworks and makes working with the query object a pain. Any query value could be either a string, an array of strings, or an object, and these arrays/objects can also be nested. So all query values accesses need to be overly defensive, or just call `.toString()` and accept that sometimes you might be getting `[object Object]` as a value.

The "bug" that this is fixing doesn't really seem like a problem to me. I think it's reasonable to assume that `ctx.request.query` will never have any values that are single-value arrays, and that you should always expect that it could be either a string or an array of strings. As far as I can tell, the only situation that this is "fixing" is if you are modifying the request's query object in middleware, setting its value to something that Node's built-in querystring parser would never produce, and you want the rest of your middleware to be able to access `ctx.request.query` while assuming certain values are an array without checking. Adding non-standard syntax to the query string to support this edge case feels like a hack.

jonathanong:
`qs` was intentionally not added to koa because it's very opinionated and has both performance and security issues. you can make your own plugin that defines `ctx.qs` or something. we should actually be moving to URLSearchParams

Issue #1683, opened 2022-08-16
[feat] request.params is not missing at the docs?


ArthurPedroti:
Hey everyone, I'm current use the params of a request at my backend, and it's a comom thing you do when you are creating the update route of your application.

And I don't thing anything about this property in the request object docs.

Shouldn`t that docs be there?

By the way I can even do this part of docs, I'm only asking myself why that is not over there.

## Checklist

- [x] I have searched through GitHub issues for similar issues.
- [x] I have completely read through the README and documentation.

kedarchandrayan:
PR https://github.com/koajs/koa/pull/1684 will solve this missing documentation issue. Please merge it.

miwnwski:
There is no request.params in Koa. I assume you are using [@koa/router](https://github.com/koajs/router), in which case this should be moved to that repo.

PR #1684, opened 2022-08-19
request.params missing in the docs. Issue#1683.

kedarchandrayan:
This PR will solve issue: https://github.com/koajs/koa/issues/1683

request.params documentation was missing. I have included it with suitable example.

### DIFF
--- a/docs/api/request.md
+++ b/docs/api/request.md
@@ -150,6 +150,18 @@ ctx.request.charset;
 ctx.query = { next: '/login' };
 ```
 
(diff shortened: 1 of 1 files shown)


miwnwski:
There is no request.params in Koa. I assume you are using [@koa/router](https://github.com/koajs/router), in which case this should be moved to that repo.

miwnwski (changes requested):
There is no request.params in Koa.

Issue #1607, opened 2021-11-05
Distribute TypeScript types with koa package


justinfagnani:
It'd be really convenient if the Koa types were distributed with the koa package rather than requiring a separate install of `@types/koa`.

I see that requests for a TypeScript rewrite have been closed, but could the types at least be pulled in? It might help keep them in sync as well.

jimmywarting:
sometimes i don't get it, vscode is able to follow js and understand jsdoc without having separate type definition files. I most often see this @types as unnecessary

miwnwski:
I'll leave this open for now but there's little engagement from maintainers to add/update TS support in any form to this repo.

justinfagnani:
That's unfortunate.

I know npm download numbers aren't perfectly representative, but I'd bet that _significant_ fraction of your users are using Koa via TypeScript and would benefit from typings being in sync.

<img width="1342" alt="Screen Shot 2022-07-01 at 8 14 00 AM" src="https://user-images.githubusercontent.com/522948/176922172-17d729ce-026d-429a-8ce4-00ef39d58978.png">

miwnwski:
Sure, maybe. I get it. I just have no personal stake in getting TS in so I will not push for it. If anything adding TS typings might just push me to leave entirely (not that anyone would care :), I'm just saying, it's not an investment I care to take on at least)

jaydenseric:
The best way to add type safety to Koa would be to use [TypeScript JSDoc comments](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html).

Add a `typescript` dev dependency:

https://github.com/jaydenseric/graphql-api-koa/blob/b8aac9a2927ec9aeb0954b93e170e4604aa676f5/package.json#L59

```
"typescript": "^4.7.4"
```

Because `koa` types probably reference Node.js types, the `@types/node` package should be a dependency with a `*` version:

https://github.com/jaydenseric/graphql-upload/blob/aa15ee0eb2b3a4e2421d098393bbbf9252f1a8c7/package.json#L65

```
"@types/node": "*",
```

If Node.js types are only used in non-published modules (i.e. tests) then it could be a dev dependency instead with the latest version:

https://github.com/jaydenseric/graphql-api-koa/blob/b8aac9a2927ec9aeb0954b93e170e4604aa676f5/package.json#L49

```
"@types/node": "^18.0.0",
```

[...]

PR #1726, opened 2022-12-07
Provide more typings for application's context and request/response

mihailik:
Few extra JSDoc decorations that convey types so end-users of the library get better experience:

Screenshot 2022-12-07 at 13 49 03

Similar to #1607 and [the suggestion](https://github.com/koajs/koa/issues/1607#issuecomment-1172759716) by @jaydenseric except no project changes, just tweaking JSDoc 👍

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -20,6 +20,20 @@ const http = require('http')
 const only = require('only')
 const { HttpError } = require('http-errors')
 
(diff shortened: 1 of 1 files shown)


mihailik (commented):
Sorry missed space

fengmk2 (approved):

PR #1732, opened 2022-12-24
Initialize context.state sooner

### LINKED ISSUE #1646: initialize `context.state` sooner
# Expected Behavior

If there's information I want to be available in every request, I'd like to be able to specify that information when I initialize the app, using `app.context.state`, like so:

```js
const app = new Koa()
app.context.state.prop = true // TypeError: Cannot set properties of undefined (setting 'prop')
app.use(ctx => console.log(ctx.state.prop))
```

Best practice for Koa is to use `context.state` to pass information through middleware, and to use `app.context` add properties used throughout the app, so this seems like the right approach. It's also parallel to the way that I can set `app.locals` in an Express app to make that information available to `res.locals`.

# Actual Behavior

The above code snippet throws a TypeError because `context.state` isn't initialized until the app receives a request, here:

https://github.com/koajs/koa/blob/aa816ca523e0f7f3ca7623163762a2e63a7b0ee3/lib/application.js#L168-L181

The following also doesn't work, because the `state` property gets overridden by the code linked above:

```js
const app = new Koa()
app.context.state = { prop: true }
app.use(ctx => console.log(ctx.state.prop)) // returns undefined
```

# Suggested Solution

Add the `state` property to the [context prototype](https://github.com/koajs/koa/blob/master/lib/context.js) instead of initializing it when a request comes in. Then, in `app.createContext()`, create a new `state` object that only lives for the lifetime of the request, using `app.context.state` as its prototype.

Replace:

https://github.com/koajs/koa/blob/aa816ca523e0f7f3ca7623163762a2e63a7b0ee3/lib/application.js#L179

with:

```js
// context.state = Object.assign({}, context.state) // <- old suggestion
context.state = Object.create(context.state)        // <- this is probably better
```

# Workaround

The workaround is to add the information directly to `app.context` rather than `app.context.state`. However, this is undesirable because it goes against the best practice of using `context.state` to encapsulate state.

```js
const app = new Koa()
app.context.prop = true
app.use(ctx => console.log(ctx.prop)) // returns true
```

I'd be happy to open a PR for this if it's a change you'd like implemented.

krisstern:
Fixes #1646

## Description

Allow context creation an overridable Koa option. 

## Checklist

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -208,7 +208,7 @@ module.exports = class Application extends Emitter {
     request.response = response
     response.request = request
     context.originalUrl = request.originalUrl = req.url
-    context.state = {}
+    context.state = Object.assign({}, context.state)
     return context
   }
 
--- a/__tests__/context/state.js
+++ b/__tests__/context/state.js
@@ -10,7 +10,7 @@ describe('ctx.state', () => {
     const app = new Koa()
 
     app.use(ctx => {
-      assert.deepStrictEqual(ctx.state, {})
+      assert.deepStrictEqual(ctx.state, this.state || {})
     })
 
     const server = app.listen()
--- a/lib/context.js
+++ b/lib/context.js
@@ -50,6 +50,7 @@ const proto = module.exports = {
       request: this.request.toJSON(),
       response: this.response.toJSON(),
       app: this.app.toJSON(),
+      state: this.state,
       originalUrl: this.originalUrl,
       req: '<original node req>',
       res: '<original node res>',


lib/application.js
```diff
@@ -208,7 +208,7 @@ module.exports = class Application extends Emitter {
     request.response = response
     response.request = request
     context.originalUrl = request.originalUrl = req.url
-    context.state = {}
+    context.state = Object.assign({}, context.state)
```
dwhieb, line 211: It might be better to use `Object.create()` here, to inherit the prototype from `context.state`.

```suggestion
    context.state = Object.create(context.state)
```

dwhieb: Sorry, just saw this PR fail because my suggestion broke it. This should fix it:

```suggestion
    context.state = Object.create(context.state || {})
```

This would be a great place to use [logical nullish assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment), but Koa has to support Node v12 which doesn't support that, and I don't think Koa transpiles its code (?), so better safe than sorry.

krisstern: Tried this one locally but some tests are faiiling, still does not work

dwhieb: Huh okay. Sorry for the detour. This still definitely covers the use case. Thanks for tackling this PR!

antaranyan (approved):

antaranyan (changes requested):
Change looks fine for what it does.

Consider asserting this in [context state test](https://github.com/koajs/koa/blob/bec13ecccdf7c734bccd5dd0ee9892621415af41/__tests__/context/state.js).

krisstern:
> Change looks fine for what it does.
> 
> Consider asserting this in [context state test](https://github.com/koajs/koa/blob/bec13ecccdf7c734bccd5dd0ee9892621415af41/__tests__/context/state.js).

Hi @antaranyan Sure, I have just modified the exisitng test for `context` > `state`.

lib/application.js
iwanofski, line 211: The whole point of `ctx.state` as I see it is to ensure there's no state bleed between requests.
As far as I see it, this change defeats this original purpose.

jonathanong:
what's the point of this? why not just make a middleware that does what you want?

dwhieb:
@jonathanong https://github.com/koajs/koa/issues/1646

kevinpeno:
Pending outcome of discussion in #1646

PR #1760, opened 2023-05-17
docs: fix node quickly install with nvm

Songkunda:
update docs/api/index.md:8  `nvm install  7 ` to `nvm install 12`


zacanger:
I know this is already closed, but rather than having a hardcoded version, why not use a flag like `--lts`?

PR #1796, opened 2024-01-11
Replacing eslint with standard cli

etroynov:
Hi,

While working on one of the plugins for koa which is used in our company I needed to study some points in the `Koa` source code, during the installation of dependencies I noticed `deprecating` warnings.

In the course of studying the project's dependencies I found:
- `eslint-plugin-node` - the plugin has not been updated for the last 4 years and for this reason a fork of `eslint-plugin-n` was created which is recommended for use;
- `eslint-plugin-standard` - marked as deprecated and the official `standard` website recommends using their cli utility;
- `.eslintrc.yml` - does not have any special settings and uses only:
```js
extends: standard
```

# why not just update eslint?

I can see several reasons why the official utility is better than the eslint custom config ( in my opinion ofc ):
- **fewer dependencies** - With eslint we have to monitor and update the list of dependencies:
```json
"eslint": "^7.32.0",
"eslint-config-standard": "^16.0.3",
"eslint-plugin-import": "^2.18.2",
"eslint-plugin-node": "^11.1.0",
"eslint-plugin-promise": "^5.1.0",
"eslint-plugin-standard": "^5.0.0",
```
With `standartd` we have only two dependecies and one of it is optional:
```json
"snazzy": "^9.0.0", // just for pretty output
"standard": "^17.1.0",
```
- **deprecated dependencies** - If, for example, one of the dependencies becomes unsupported, we will need to find a replacement for it, which will also need to be monitored, so it seems to me that if we can avoid this, we should do it.
- **official way** - If a tool has an official cli that is maintained and updated by their team, why not just use it?

## Checklist


etroynov:
@dead-horse, @fengmk2 

Can you take a look?

fengmk2 (approved):
+1

etroynov:
@fengmk2  thx a lot :)

PR #1797, opened 2024-01-11
Drop node.js 12.x, 14.x, 17.x support

etroynov:
# Why we should drop old node.js version support?

There are both pros and cons to supporting as many versions of a language as possible. However, in my subjective opinion, the pros are much less than the cons and I will try to explain it.

Pros:
- you don't need to update the infrastructure often - a complete infrastructure update is sometimes fraught with complex problems and sometimes impossible;
- allows to work with libraries that do not work with new language versions (unsupported libraries) but they have no analogs to replace them;

Cons:
- **security** - unsupported versions of the language do not receive patches to fix critical vulnerabilities found;
- **lack of new features** - supporting and allowing the use of obsolete language versions sometimes blocks the use of new functionality. A good example of this is the `express` library which supports node.js 

```json
"node": ">= 0.10.0"
``` 

That's one of the reasons why the express doesn't support `async/await` of the box ( and I'm not sure that's gonna happen in the foreseeable future );

- **human factor** - When tools support even very old versions of a language, people and companies (if there is no such practice) often forget that it needs to be updated. As an example, I witnessed a hack of company resources through a language vulnerability that was fixed in the next major version, but since the company did not have the practice of regular updates (not a large company), the team only at that moment found out that the language version used on the server has long been unsupported, and it was not supported for a long time.

# Summary

I think version 3 is a good reason to make it a practice to support only those versions that are active and regularly receive security patches.

## Checklist


etroynov:
@3imed-jaberi can you look on it?

jonathanong (approved):

3imed-jaberi (approved):
@etroynov could you please rebase!

etroynov:
@jonathanong @3imed-jaberi, done.

Thanks for the help!

My question is: is there an opportunity for me to join the Koa project as an active developer or maintainer? I have a lot of experience with JavaScript (over 15 years), I also work with Koa and Express every day and have a lot of free time.

If there is any documentation on the joining process or something similar, could you share the link?

fengmk2 (approved):
+1

Issue #1800, opened 2024-01-17
if "\@" in url ，the redirect result doesn't match the host of the new URL()


FDrag0n:
https://github.com/koajs/koa/blob/5f159415e58c42b54cb0703d7cfb10870d33d65f/lib/response.js#L269

<img width="910" alt="image" src="https://github.com/koajs/koa/assets/34733637/46c9d2dc-74a2-4e4a-bf94-d3e2e822004e">

Developer often use `new URL()` `host` to verify the redirected url, but here the `encodeurl` will cause the host verification to fail, thus creating a URL hopping vulnerability

PR #1805, opened 2024-03-21
fix: handle upper case protocol like HTTP or HTTPS

FDrag0n:
https://github.com/koajs/koa/issues/1800 's fix https://github.com/koajs/koa/pull/1804 forgot to consider the case where url is uppercase HTTP and HTTPS

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -266,7 +266,7 @@ module.exports = {
   redirect (url, alt) {
     // location
     if (url === 'back') url = this.ctx.get('Referrer') || alt || '/'
-    if (url.toLowerCase().startsWith('https://') || url.toLowerCase().startsWith('http://')) {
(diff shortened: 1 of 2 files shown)


lib/response.js
```diff
@@ -266,7 +266,7 @@ module.exports = {
   redirect (url, alt) {
     // location
     if (url === 'back') url = this.ctx.get('Referrer') || alt || '/'
-    if (url.startsWith('https://') || url.startsWith('http://')) {
+    if (/^https?:\/\//i.test(url)) {
```
fengmk2, line 269: @FDrag0n I change to use Regex to avoid lowerCase twice.

FDrag0n: It's a nice change👍

PR #1825, opened 2024-06-21
Add custom stream support

KristapsR:
## Description

There are situations when we need to send response stream that is not an instance of node `Readable` stream and current way of checking for stream `value instanceof Stream` fails.
e.g. when using alternative module like [readable-stream](https://github.com/nodejs/readable-stream) or some [other cases](https://stackoverflow.com/questions/23885095/nodejs-check-if-variable-is-readable-stream)

## My usecase

```js
import archiver from 'archiver'

const archive = archiver('zip')
const stream = ...
archive.append(stream, { name: 'archive.zip'})

res.body = archive
```

## Solution

- Added tests to reproduce issue with mocked custom stream object that represents possible `Readable` stream implementation.
- Addopted code from [is-stream](https://github.com/sindresorhus/is-stream)

## Tested also with this code

```js
const archiver = require('archiver')
const Stream = require('stream')
const CustomStream = require('./test-helpers/stream.js')
const isReadableStream = require('./lib/is-readable-stream.js')

const stream = new Stream()
const archive = archiver('zip')
const customstream = new CustomStream.Readable()

console.log(stream instanceof Stream) // true
console.log(customstream instanceof Stream) // false
console.log(archive instanceof Stream) // false

console.log(isReadableStream(stream)) // true
console.log(isReadableStream(customstream)) // true
console.log(isReadableStream(archive)) // true
```

### DIFF
--- a/__tests__/response/body.js
+++ b/__tests__/response/body.js
@@ -1,6 +1,7 @@
 'use strict'
 
 const response = require('../../test-helpers/context').response
+const CustomStream = require('../../test-helpers/stream')
 const assert = require('assert')
 const fs = require('fs')
 const Stream = require('stream')
@@ -106,6 +107,12 @@ describe('res.body=', () => {
     it('should default to an octet stream', () => {
       const res = response()
       res.body = fs.createReadStream('LICENSE')
+      assert.strictEqual('application/octet-stream', res.header['content-type'])
+    })
+
+    it('should support custom stream', () => {
+      const res = response()
+      res.body = new CustomStream.Readable()
       assert.strictEqual('application/octet-stream', res.header['content-type'])
     })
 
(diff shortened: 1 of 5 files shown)


jonathanong:
why not use is-stream directly? because it is in es6?

jonathanong:
an external dependency is preferred if maintained from a reliable author, I can try to look later

kevinpeno:
@jonathanong I reviewed the [library in question](https://www.npmjs.com/package/is-stream). Considering we're planning to move into the TS world, while I don't tend to prefer libraries that use `d.ts` instead of being natively written in TS, the way they have added a `d.ts` test file ensures that it isn't lying about its types. I'd move for inclusion if that's the preference.

jonathanong:
that's a good point, thanks! cc @kevinpeno

PR #1861, opened 2025-03-18
breaking: remove special ENOENT support

### LINKED ISSUE #1420: koa shouldn't special-case ENOENT errors
I noticed that koa special-cases ENOENT errors:
https://github.com/koajs/koa/blob/ed84ee50da8ae3cd08056f944d061e00d06ed87f/lib/context.js#L146-L147
I think this is a mistake: ENOENT errors could happen for a lot of reasons (this happens with google's storage library when the key file isn't found, for example) that shouldn't get a 404 status.
And BTW it's not even documented.

jonathanong:
This is a huge breaking change as it breaks `fs.createReadStream` support. Instead, you'll have to make sure you'll handle the stream's errors before it reaches the koa response logic. This is good as you'd want to do this for all streams anyways. You can have an upstream middleware like:

```js
app.use(async (ctx, next) {
  await next()

  // check if the object is a stream
  if (ctx.body?.readable && typeof ctx.body?.pipe === 'function') {
    await new Promise((resolve, reject) => { // wait for it to be either readable or throw an error
      ctx.body.on('readable', () => resolve())
      ctx.body.on('error', (err) => reject(err)) // TODO handle errors
    })
  }
})
```

closes #1420

### DIFF
--- a/__tests__/application/respond.test.js
+++ b/__tests__/application/respond.test.js
@@ -678,21 +678,6 @@ describe('app.respond', () => {
         assert.deepStrictEqual(res.body, pkg)
       })
 
-    it('should handle errors', (t, done) => {
-      const app = new Koa()
-
-      app.use(ctx => {
-        ctx.set('Content-Type', 'application/json; charset=utf-8')
-        ctx.body = fs.createReadStream('does not exist')
-      })
-
-      request(app.callback())
-        .get('/')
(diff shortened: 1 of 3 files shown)


fengmk2 (approved):

BryceMehring:
Does https://github.com/koajs/send need to be updated with the new logic?

Issue #1864, opened 2025-03-22
[feat] Improved stream handling


jonathanong:
## Describe the feature

- [ ] Improve how stream errors are handled
- [ ] Improve how stream pipelines are created
- [ ] Support various streams as responses (such as a `fetch` response stream and a body)

PR #1889, opened 2025-06-20
use pipeline() instead of pipe()

rosald:
## Checklist

pipe(rs, res) will not destroy res, when an error occurs. Thus causing client "hang" infinite when it has received some part of data.

Following code example, a readable stream has been implemented. When client request, the server will send some data, but after the error occurs, the client hang, the server doesn't close the socket. Chrome will hang. curl and ab will end in timeout.

After replacing with pipeline, when error occur, koa immediately close the socket, Chrome will say "Check internet connection" in download page. curl and ab will finish request as soon as possible.

stream.pipeline() will call stream.destroy(err) on all streams except they have already been closed.

the behavior may be better than hanging the request. Instead of client waitting indefinitely for the remaining bytes, closing the connection will break the expectation, but it's the only way to signal that the response is incomplete. The client will know that the response is truncated. Headers are already sent, an error occurs mid-stream, the client could handle this predictably.

Here's another to consider:

It's koa duty to close connection, it's app writer's duty to close the readable stream, and it's (better) stream implementers duty to release corresponding resources(request socket, fd, etc) of the readable stream when error occurs(but if it didn't work well, app writer should handle it, not koa).

Last benifit is pipeline() added support for webstreams(v18.16.0), making it easier.

```js
import { Readable } from 'node:stream';
import http from 'http';
import Koa from 'koa';

class MyStream extends Readable {
  #count = 0;
  _read(size) {
    this.#count++;
    this.push(':-)');
    if (this.#count === 4) {
      try {
        // https://nodejs.org/api/stream.html#errors-while-reading
        throw new Error('Error in stream');
      } catch (e) {
        this.destroy(e);
        return;
      }
    }
    if (this.#count === 5) {
      this.push(null);
    }
  }
}

const httpPort = 8081;
const app = new Koa();

app.use((ctx) => {
  if (ctx.request.path === '/favicon.ico') {
    ctx.status = 404;
    return;
  }

  const stream = new MyStream();

  ctx.body = stream;
});

http.createServer(app.callback()).listen(httpPort);

```

### DIFF
--- a/__tests__/application/respond.test.js
+++ b/__tests__/application/respond.test.js
@@ -678,12 +678,12 @@ describe('app.respond', () => {
         assert.deepStrictEqual(res.body, pkg)
       })
 
-    it('should handle errors when no content status', () => {
+    it('should not handle errors when no content status, but let user do it self', () => {
       const app = new Koa()
 
       app.use(ctx => {
         ctx.status = 204
-        ctx.body = fs.createReadStream('does not exist')
+        ctx.body = fs.createReadStream('does not exist').on('error', (err) => {})
       })
 
       return request(app.callback())
--- a/__tests__/response/body.test.js
+++ b/__tests__/response/body.test.js
@@ -117,14 +117,14 @@ describe('res.body=', () => {
       assert.strictEqual('application/octet-stream', res.header['content-type'])
     })
 
-    it('should add error handler to the stream, but only once', () => {
+    it('should not add error handler to the stream, but let the user do it self', () => {
       const res = response()
       const body = new Stream.PassThrough()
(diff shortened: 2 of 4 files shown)


jonathanong:
@rosald thanks for taking a stab! This is what I was thinking about for https://github.com/koajs/koa/issues/1864

However, the PR as-is wouldn't suffice as we are hiding errors. Most likely, the best approach is to have a new API, something like `ctx.body = stream` and `ctx.transforms = []`, which we then pipeline into `Stream.pipeline(ctx.body, ...ctx.transforms, ctx.res` with proper error handling. Then, middleware like `compress` can just be a transform function added to `ctx.transforms`

PR #1928, opened 2026-02-20
feat: improved stream handling

veeceey:
## Summary

Addresses #1864 - Improved stream handling in Koa.

- **Async iterable/generator support**: `ctx.body` now accepts async iterables (e.g. `async function*`) as response bodies, converting them to Node.js readable streams via `Stream.Readable.from()`. This is useful for streaming responses from async data sources without manually creating streams.
- **Better stream pipeline error handling**: `ERR_STREAM_PREMATURE_CLOSE` errors (caused by client disconnects) are now filtered out in the `stream.pipeline` callback, preventing spurious error events on the application when clients simply close their connections.
- **Body setter support for async iterables**: The response body setter properly handles async iterables by setting the default content type to octet-stream, stripping Content-Length (since the length is unknown ahead of time), and cleaning up any previous stream body.

### Changes

- `lib/is-stream.js`: Added `isAsyncIterable()` helper that detects async iterables while excluding types already handled (Node.js streams, ReadableStream, Blob, Response)
- `lib/application.js`: Added async iterable detection in `respond()`, filtered `ERR_STREAM_PREMATURE_CLOSE` in pipeline error callback
- `lib/response.js`: Added async iterable handling in the `body` setter
- `__tests__/application/respond.test.js`: Added 9 new tests covering async generator responses, async iterable responses, buffer yields, content type handling, error propagation, and premature close filtering
- `__tests__/response/body.test.js`: Added 4 new tests for body setter behavior with async iterables

## Test plan

- [x] All 432 tests pass (419 existing + 13 new)
- [x] Standard linter passes on all modified source files
- [ ] Verify async generator responses stream correctly with `ctx.body = asyncGenerator()`
- [ ] Verify client disconnects no longer emit `ERR_STREAM_PREMATURE_CLOSE` errors
- [ ] Verify existing stream, ReadableStream, Blob, and Response body types still work correctly

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -19,7 +19,6 @@ const request = require('./request')
 const response = require('./response')
 const context = require('./context')
 const isStream = require('./is-stream.js')
-const { isAsyncIterable } = require('./is-stream.js')
 const only = require('./only.js')
 
 /** @typedef {typeof import ('./context') & {
@@ -311,7 +310,7 @@ function respond (ctx) {
   else if (body instanceof ReadableStream) stream = Stream.Readable.from(body)
   else if (body instanceof Response) stream = Stream.Readable.from(body?.body || '')
   else if (isStream(body)) stream = body
-  else if (isAsyncIterable(body)) stream = Stream.Readable.from(body)
+  else if (typeof body[Symbol.asyncIterator] === 'function') stream = Stream.Readable.from(body)
 
   if (stream) {
     return Stream.pipeline(stream, res, err => {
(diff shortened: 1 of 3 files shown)


veeceey:
Good call, yeah that function was doing unnecessary work since all those types (Stream, ReadableStream, Blob, Response) are already handled by the preceding branches. Removed `isAsyncIterable` and inlined the `Symbol.asyncIterator` check directly. Thanks for the suggestion!

PR #1948, opened 2026-04-01
fix: handle invalid Referrer URLs in back() without crashing

guoyangzhen:
## Description

Both `redirect()` and `back()` called `new URL()` without error handling. Invalid URL input (malformed, malicious, or incomplete) would throw an uncaught `TypeError` and crash the request with a 500 error.

### redirect()

```js
// Before
url = new URL(url).toString() // throws on 'https://' or '://invalid'

// After
try {
  url = new URL(url).toString()
} catch {
  // invalid URL, encode as-is
}
```

### back()

```js
// Before
const url = new URL(referrer, this.ctx.href) // throws on invalid Referrer

// After
try {
  const url = new URL(referrer, this.ctx.href)
  if (url.host === this.ctx.host) { ... }
} catch {
  // invalid URL, fall through to alt
}
```

## Impact

Without this fix, a request with `Referrer: ://invalid` or `Location: https://` crashes the entire request handler.

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -96,7 +96,8 @@ module.exports = {
    */
 
   get origin () {
-    return this.req.headers.origin || null
+    if (!this.host) return null
+    return `${this.protocol}://${this.host}`
   },
 
   /**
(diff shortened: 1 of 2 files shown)


jonathanong:
this is intentional. you will need to handle errors yourself and decide what to do with them.

PR #1971, opened 2026-05-24
fix(request): return URL origin instead of Origin header

### LINKED ISSUE #1958: [fix] Missing Headers in Koa 3 vs 2 ( or update docs? )
## Describe the bug

**Node.js version:** v22.20.0

**OS version:** Mac OS 26.4

**Description:** Many of the ctx properties appear to be missing in Koa v3 vs v2

## Actual behavior

* Koa v3 output of `ctx.origin` = `null`

## Expected behavior

* Koa v2 (2.16.4) output `ctx.origin` = `http://127.0.0.1:3000`

## Code to reproduce

```
const Koa = require('koa');
const app = new Koa();

app.use(async ctx => {
  console.log(ctx.origin)
  ctx.body = 'Hello World';
});

app.listen(3000);
```

Not sure if this is a bug or I'm following the wrong documentations. Followed the guide here

https://github.com/koajs/koa/blob/master/docs/migration-v2-to-v3.md

## Checklist

- [x] I have searched through GitHub issues for similar issues.
- [x] I have completely read through the README and documentation.
- [x] I have tested my code with the latest version of Node.js and this package and confirmed it is still not working.

xxiaoxiong:
Fixes #1958

Why
ctx.origin in Koa 3 returns the value of the HTTP Origin request header (via req.headers.origin), which is null for direct browser requests. The expected behavior is to return the URL origin (e.g. http://127.0.0.1:3000), consistent with Koa 2 and the URL.origin specification.

Changes
Changed get origin() in lib/request.js to return this.protocol + :// + this.host instead of this.req.headers.origin || null.

Updated component: lib/request.js

Surface area
- Request
- Bug fix


jonathanong:
see original issue

additionally, when making PRs, please make sure you write tests.

PR #1984, opened 2026-06-26
docs: fix `has()` JSDoc example calling `get()` instead of `has()`

mahirhir:
## What

In `lib/response.js`, the JSDoc for `has()` contains a second example that mistakenly calls `get()` instead of `has()`:

```js
//     this.get('content-type');
//     // => true
```

`get(field)` returns `this.res.getHeader(field)` — a header value string, never `true`. The comment `// => true` is factually impossible for `get()`. The intent of the block is to show that `has()` is case-insensitive, so the call should be `this.has('content-type')`.

## Fix

One-line change: `this.get('content-type')` → `this.has('content-type')` in the `has()` JSDoc block (line 507).

No logic changes, no new tests needed.

## Summary by Sourcery

Documentation:
- Fix the `has()` method JSDoc example to call `has()` instead of `get()` when demonstrating case-insensitive header lookup.

## Summary by CodeRabbit

* **Documentation**
  * Corrected an example in the response header-checking docs to show the proper method for verifying whether a header is set.


Ziiyodullayevv (commented):
Good documentation fix. Correcting the has() JSDoc example to actually call has() instead of get() prevents developer confusion.

mahirhir:
Closing this, it's a cosmetic docs-only change and not worth your review time. Thanks.

Issue #1777, opened 2023-09-20
[feat] Support Web (WHATWG) stream, Blob and Response


hax:
## Describe the feature

Currently `response.body=` or `ctx.body=` support buffers, strings and Nodejs streams. I suggest we also add support to Web stream, Blob and Response (currently just fallback to json which break users expectation), maybe also all async iterables, Nodejs already have those APIs in global.

This is my workaround, it works but it's better to support them in the core.

```ts
app.use(async (ctx, next) => {
    await next();
    // Note: need to clear type before set body, or the type might always be json
    // See Koa impl: https://github.com/koajs/koa/blob/dbf4b8f41286befd53dfd802740f2021441435bf/lib/response.js#L134-L190
    if (ctx.body instanceof ReadableStream) {
        ctx.type = "";
        ctx.body = Readable.fromWeb(ctx.body);
    } else if (ctx.body instanceof Blob) {
        const blob = ctx.body;
        ctx.type = blob.type;
        ctx.length = blob.size;
        ctx.body = Readable.fromWeb(blob.stream());
    } else if (ctx.body instanceof Response) {
        const response = ctx.body;
        ctx.status = response.status;
        ctx.type = "";
        // Note: don't clear all headers so the headers added by middlewares still effective
        for (const [name, value] of response.headers) ctx.set(name, value);
        if (response.redirected) ctx.redirect(response.url);
        else ctx.body = Readable.fromWeb(response.body);
    }
    // maybe we could also support all async iterables
    // else if (ctx.body[Symbol.asyncIterator]) {
    //   ctx.type = ""; ctx.body = Readable.from(ctx.body);
    // }
});
```

## Checklist

- [x] I have searched through GitHub issues for similar issues.
- [x] I have completely read through the README and documentation.

fengmk2:
@hax Cool! Please send us a pull request, we love to merge it!

iwanofski:
I support this, especially if its additive only (i.e. no breaking)

image72:
@fengmk2  have any plan to support this feature?

fengmk2:
> @fengmk2 have any plan to support this feature?有计划支持这个功能吗？

@hax  has given the implementation code and you are welcome to submit a pull request.

TommyDew42:
Will have a PR for it soon

kravorkid:
@fengmk2 i've done a PR for this i'm open to suggestions for improvement 😉

PR #1987, opened 2026-07-01
fix(response): keep the Blob content type instead of forcing octet-stream

spokodev:
When a Blob is assigned to `ctx.body`, Koa forces `Content-Type: application/octet-stream` and ignores the Blob's own `type`. A Blob carries its MIME type in `blob.type`, so a typed Blob such as `new Blob(data, { type: 'image/png' })` (or a Blob returned by `fetch(...).then(r => r.blob())`) is sent with the wrong content type. Browsers then download the payload instead of rendering it, and any client that negotiates on `Content-Type` sees the wrong value.

The sibling `Response` body path already copies the body's own `Content-Type`, and the original feature request (#1777) specified `ctx.type = blob.type`, so the Blob path is the outlier.

## Cause

In `lib/response.js`, the Blob branch of the body setter is:

    if (setType) this.type = 'bin'

`'bin'` maps through `mime-types` to `application/octet-stream`, discarding `val.type`. The `setType` guard (an explicitly preset Content-Type wins) is correct; only the fallback value is wrong.

## Fix

    if (setType) this.type = val.type || 'bin'

Honor the Blob's own MIME type when present, and fall back to `application/octet-stream` for an untyped Blob (the existing default). An explicit `ctx.type` still wins.

Added two tests to `__tests__/response/body.test.js`: a typed Blob keeps its type, and an explicit content type is not overridden. The first fails before this change; the full suite (442) stays green.

## Summary by Sourcery

Preserve Blob MIME types in response headers instead of defaulting typed Blobs to application/octet-stream.

Bug Fixes:
- Preserve a Blob's declared MIME type when assigning it as a response body, while retaining the binary fallback for untyped Blobs and respecting explicitly configured content types.

Tests:
- Add coverage for preserving typed Blob content types and for not overriding explicit response content types.

## Summary by CodeRabbit

* **Bug Fixes**
  * Blob responses now preserve an existing media type when one is provided, instead of always defaulting to a generic type.
  * Explicit response content types are now respected when setting a Blob body, preventing them from being overwritten.
* **Tests**
  * Added coverage for Blob content-type handling to verify both default behavior and explicit type precedence.


Ziiyodullayevv (commented):
Good fix for Blob content type handling. Keeping the original content type instead of forcing octet-stream preserves the intended MIME type.

3imed-jaberi (approved):

spokodev:
Nudge on this one — approved on 2 August, still mergeable, no open threads. Happy to rebase if that helps.

yowainwright (approved):
LGTM! Thanks for the thoughtful work and the nudge @spokodev! 🥇

PR #1992, opened 2026-07-30
fix: throw Koa HttpError from ctx.assert instead of nested http-errors instance

### LINKED ISSUE #1925: [fix] ctx.assert() throws specific error classes
**Description:** The JSDoc for HttpError in Koa states:
> A re-export of HttpError from the http-error package. This is the error type that is thrown by ctx.assert() and ctx.throw().

However, ctx.assert() actually throws specific error class, not the HttpError class. This breaks instanceof HttpError checks in error handling.

## Code to reproduce

```import Router from '@koa/router'
import { HttpError } from 'koa'

const router = new Router()

router.get('/test', async (ctx) => {
  try {
    const data = null
    ctx.assert(data, 401, 'Unauthorized')
  } catch (e) {
    console.log(e instanceof HttpError) // ❌ false - should be true
  }
})

marceli1404:
Fixes #1925

\ctx.assert\ delegates to \http-assert\, which depends on \http-errors@~1.8.0\ (installed in its own \
ode_modules\). Since Koa exports \HttpError\ from its own \http-errors@^2.0.0\ dependency, errors thrown by \ctx.assert()\ did not satisfy \rr instanceof Koa.HttpError\.

This change wraps \http-assert\ to re-throw assertion failures using Koa's \createError\ (from \http-errors\ v2), so both \ctx.assert\ and all its helper methods (\.ok\, \.equal\, \.strictEqual\, etc.) throw proper \HttpError\ instances.

Test plan:
- \
ode --test\ — 446 tests, 0 failures
- \
pm run build\ — passes
- \
pm run lint\ — passes

## Summary by Sourcery

Ensure ctx.assert and its helper methods throw Koa.HttpError instances instead of errors from the nested http-assert/http-errors dependency.

Bug Fixes:
- Fix HttpError instanceof checks by rethrowing assertion failures using Koa's http-errors integration.

Enhancements:
- Wrap http-assert to normalize assertion errors into Koa HttpError, preserving relevant error properties.

Tests:
- Extend context assert tests to verify HttpError instances and behavior of all assert helper methods.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved context assertion errors so they consistently use standard HTTP error types.
  - Preserved assertion status, messages, exposure settings, and custom error details.
  - Ensured all built-in assertion helpers follow the same error-handling behavior.

- **Tests**
  - Expanded coverage for successful assertions and expected error properties across all assertion helpers.


marceli1404:
@fengmk2 @dead-horse — friendly ping. This PR fixes #1925: `ctx.assert()` now throws proper `Koa.HttpError` instances (matching the documented behavior and the JSDoc in `lib/context.js`), instead of errors from the nested `http-assert` copy of `http-errors`. Added a regression test covering custom error properties. All 447 tests pass. Thanks for considering!

fengmk2:
@marceli1404 I think we should change the document description is better than change the code.

PR #1994, opened 2026-08-17
docs: point request.hostname URL link at current Node.js docs

JiuYue0820:
## Problem

`docs/api/request.md` (`request.hostname`) links the WHATWG URL API at

```
https://nodejs.org/dist/latest-v8.x/docs/api/url.html#url_the_whatwg_url_api
```

That is the Node.js 8 documentation snapshot. Node 8 has been EOL since 2019, and Koa itself requires `node >= 18` (`package.json` `engines`). New WHATWG URL features referenced by current Koa behavior (e.g. `URL` available globally, `url.URL` parsing used for IPv6 hostnames) are documented in the current docs, not the Node 8 snapshot.

## Solution

Link to the current WHATWG URL API section:

```
https://nodejs.org/api/url.html#the-whatwg-url-api
```

## Verification

- `curl -L -o /dev/null -w "%{http_code}"` returns 200 for the new URL.
- The anchor `#the-whatwg-url-api` exists on the page (section "The WHATWG URL API").
- `grep -n "engines" -A2 package.json` shows `"node": ">= 18"`.

## Summary by Sourcery

Documentation:
- Update the WHATWG URL API link in request.hostname docs to point to the current Node.js documentation.

## Summary by CodeRabbit

* **Documentation**
  * Updated the `request.hostname` documentation link to reference the current Node.js URL API documentation.


yowainwright:
_Copied from human-written text in #1995_

The link update change and summation are correct! 👍 
> Tested: There is no change in webkit based browsers 
> by switching to the current linking mechanism vs what is there today. 
> so I'm unsure if the juice is worth the squeeze. 🤔

lbesecker195 (approved):
Verified: the `#the-whatwg-url-api` fragment exists on the current `nodejs.org/api/url.html`, and the old link pointed at the EOL Node 8 snapshot. LGTM.

PR #1995, opened 2026-08-17
docs: point response.set setHeader link at current Node.js docs

JiuYue0820:
## Problem

`docs/api/response.md` (`response.set`) links `setHeader` at

```
https://nodejs.org/dist/latest/docs/api/http.html#http_request_setheader_name_value
```

The page redirects to the current docs, but the fragment uses the legacy anchor id form. On the live page the section ids are the new style, so the browser cannot scroll to the `setHeader` section after the redirect (the old id only survives on the legacy snapshot pages).

## Solution

Link directly to the current docs page and its real anchor:

```
https://nodejs.org/api/http.html#requestsetheadername-value
```

## Verification

- `curl -sL https://nodejs.org/api/http.html | grep -c 'requestsetheadername-value'` returns 1.
- The legacy id `http_request_setheader_name_value` is absent from the current `/api/http.html` page (checked with grep).

## Summary by CodeRabbit

* **Documentation**
  * Updated the Node.js `setHeader` reference link to the current URL format in the API response documentation.


yowainwright:
The link update change and summation are correct! 👍 
> Tested: There is no change in webkit based browsers 
> by switching to the current linking mechanism vs what is there today. 
> so I'm unsure if the juice is worth the squeeze. 🤔 

Relates to #1994

PR #1999, opened 2026-09-01
fix: parse absolute-form request targets in request.URL

xia-chao:
## Problem

When the request line is HTTP/1.1 absolute-form (`GET http://example.com/foo HTTP/1.1`), `ctx.href` is correct but `ctx.URL` is not:

```js
// IncomingMessage.url === 'http://example.com/foo?q=1'
ctx.href      // 'http://example.com/foo?q=1'
ctx.URL.href  // 'http://127.0.0.1http//example.com/foo?q=1'
ctx.path      // '/foo'  (parseurl is fine)
```

The same garbage URL is produced even when the `Host` header already matches `example.com` (`http://example.comhttp//example.com/foo?q=1`). Proxies and some health checks send this request form.

## Fix

`href` already special-cases `/^https?:\/\//i` on `originalUrl` and returns it as-is. `request.URL` now uses that same branch and constructs `new URL(originalUrl)` instead of concatenating `protocol://host` onto an already-absolute URL. Invalid input still collapses to `Object.create(null)`.

`URL` continues to follow `originalUrl` after `ctx.path=` / `ctx.query=` rewrites, matching the existing `href` tests.

## Testing

```
node --test __tests__/request/whatwg-url.test.js
```

Covers http / https / `HTTP://` / port / matching Host / empty Host / path rewrite. New tests fail without the change and pass with it.

`npm test` — 447 pass, 0 fail.

## Summary by Sourcery

Correct `request.URL` parsing for absolute-form request targets.

Bug Fixes:
- Parse HTTP/1.1 absolute-form request targets correctly in `request.URL`, preserving their original scheme, host, port, path, and query instead of generating malformed URLs.

Enhancements:
- Keep `request.URL` aligned with `href` behavior and preserve the existing handling of invalid URLs and rewritten paths.

Tests:
- Add coverage for HTTP and HTTPS absolute-form targets, case-insensitive schemes, ports, Host header variations, and path rewrites.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved request URL parsing for absolute HTTP and HTTPS URLs.
  * Preserved URL details such as schemes, ports, and host headers when processing requests.
  * Ensured rewritten request paths do not overwrite the original absolute URL.


lib/request.js
```diff
@@ -298,7 +298,10 @@ module.exports = {
     if (!this.memoizedURL) {
       const originalUrl = this.originalUrl || '' // avoid undefined in template string
       try {
-        this.memoizedURL = new URL(`${this.protocol}://${this.host}${originalUrl}`)
+        // support: `GET http://example.com/foo` (same branch as href)
```
yowainwright:
```suggestion

```

yowainwright (commented):
LGTM!

javascript-unsafe:
recheck

yowainwright (approved):
Good one @bun-unsafe 🚀

PR #2010, opened 2026-09-18
fix: handle web response body lengths correctly

jakezwang:
`response.length` currently treats Web `ReadableStream` and `Response` bodies as JSON and returns `2`. This also makes HEAD responses send `Content-Length: 2` when the actual body length is unknown.

Leave those lengths undefined, and use `Blob.size` when a Blob has no Content-Length header. Explicit Content-Length values still take precedence. Tests cover the getter, HEAD responses, and preserving headers from a Response.

Validation: lint, build, commitlint, and all 464 tests with coverage on Node 22, 24, and 26.

## Checklist

## Summary by Sourcery

Handle Web response body lengths correctly without inventing Content-Length values for bodies whose size is unknown.

Bug Fixes:
- Correct response body length detection for Web ReadableStream and Response bodies by leaving unknown lengths undefined instead of guessing a serialized length.
- Use Blob size when no explicit Content-Length header is present while preserving explicitly supplied Content-Length values.

Tests:
- Add coverage for unknown Web body lengths, Blob size fallback, HEAD responses, and preservation of explicit Response headers.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved response length handling for `ReadableStream` and `Response` bodies when no `Content-Length` is available.
  * Blob-based response bodies now correctly report their size.
  * HEAD responses no longer infer or expose an incorrect content length.
  * Explicitly provided content lengths continue to take precedence.


yowainwright:
@jakezwang this is a good PR. Thank you!

I want to hold off on approving for a bit because even though this fix is correct, it reads like it could lead to a breaking change for products that have already configured workarounds for the content length being 2 vs. unknown—which you are right to address. If I'm incorrect in this concern, I'm all ears! @sourcery-ai @coderabbit-ai can you take a first-pass on my concern? 

If other teammates approve because I'm overthinking, I'm all for that as well!

I will merge this PR for sure, moving towards v4. 

Thank you! 🥇

yowainwright:
@jakezwang this is a good PR. Thank you!

I want to hold off on approving for a bit because even though this fix is correct, it reads like it could lead to a breaking change for products that have already configured workarounds for the content length being 2 vs. unknown—which you are right to address. If I'm incorrect in this concern, I'm all ears! @sourcery-ai @coderabbitai can you take a first-pass on my concern? 

If other teammates approve because I'm overthinking, I'm all for that as well!

I will merge this PR for sure, moving towards v4. 

Thank you! 🥇

